PR #28, opened 2014-10-01
Rewrote constraint application algorithm to be used consistently by

alvestrand:
getUserMedia and applyConstraints.

Changed wording to say that smaller distances should be preferred.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3029,7 +3029,6 @@
 		      its source.</p>
 		    <p>Call this set of tracks the <var>candidateSet</var>.
 		  </li>
-		  <li>
                   <li>If the value of the <var>T</var> entry of
                   <var>constraints</var> is "true", set CS to the empty
 		  constraint set (no constraint). Otherwise, continue with
@@ -3037,24 +3036,25 @@
                   set to the value of the <var>T</var> entry of
                   <var>constraints</var>.</li>
 
-		  <li><p>Run the <code><a>applyConstraint</a></code> algorithm
-		    on each track in <var>CandidateSet</var>. If the algorithm
-		    does not return an error, add the track to <var>finalSet</var>.
-		    This results in a single choice of configuration for each
-		    possible device, and eliminates devices unable to
+		  <li><p>Run
+		      the <code><a>applyConstraints()</a></code>
+		      algorithm on each track
+		      in <var>CandidateSet</var>. If the algorithm 
+		      does not return an error, add the track to
+		      <var>finalSet</var>.
+		      This results in a single choice of configuration for each
+		      possible device, and eliminates devices unable to
 		      satisfy the constraints.</p>
 		  </li>
+                  </ol>
 
-                   <p>If <var>finalSet</var> is the empty set,
-                   let <var>error</var> be a
-                   new <code><a>MediaStreamError</a> </code> object
-                   whose <code><a>name</a></code> attribute has the
-                   value <code>NotFoundError</code> and jump to the
-                   step labeled <em>Error Task</em> below.</p>
-                  </li>
-               
-                </ol>
-              </li>
+                  <p>If <var>finalSet</var> is the empty set,
+                    let <var>error</var> be a
+                    new <code><a>MediaStreamError</a> </code> object
+                    whose <code><a>name</a></code> attribute has the
+                    value <code>NotFoundError</code> and jump to the
+                    step labeled <em>Error Task</em> below.</p>
+                             </li>
 
               <li>
                 <p>Optionally, e.g., based on a previously-established user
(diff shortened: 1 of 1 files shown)


alvestrand:
Dan, can you take another look at this proposal?
I'm afraid the HTML formatting is not good; this should be fixed with "tidy", but tidy produces a 5000-line diff just now.

fluffy:
Waiting for Cullen to review

fluffy:
A bunch of the quotes and apostrophes seem to be replaced with windows style smart ones that mess up stuff a bit.

getusermedia.html
```diff
@@ -3161,8 +3071,12 @@
                     <a>MediaStreamTrack</a>
                   </code> MUST not change.</p>
 
-                <p class="issue">Define the event that should be raised when
-                the User Agent changes its choice of track.</p>
+		<p>The user agent MAY use the value of the computed "fitness
+		  distance" from the <a>SelectSettings()</a> algorithm, or
+		  any other internally-availble information about the
+		  devices, as an input
+		  to the selection algorithm.
```
fluffy: I don't understand what this means in this context. I think it needs more context to explain that in the case where the browser is not allowing the end user to pick, then it can choose however it wants. For the most part, I think UA should be encouraged to let the user pick.

getusermedia.html
```diff
@@ -3677,6 +3603,44 @@
...
+		  and the settings dictionary's value for the constraint does not
+		  satisfy the constraint, use positive infinity.</p>
+              </li>
+              <li>
+                For all numeric non-required constraints
+                (such as height, width, frameRate, aspectRatio,
+                sampleRate and sampleSize), use the
+                formula <pre>(actual == ideal) ? 0 : |actual -
+                  ideal|/max(|actual|,|ideal|)</pre>
+              </li>
```
fluffy: I think we should deal with what happens when actual and ideal are zero. One possibility would be to change the denominator to max( | actual | , | ideal | , 1.0 ). I would also be fine with just putting a note that this is an open issue and needs to be resolve and publishing. I don't think this should hold up getting a version out.

getusermedia.html
```diff
@@ -3755,39 +3696,97 @@
...
-              <li>If <var>requiredConstraints</var> is non-empty and cannot be
-              satisfied by <var>copy</var>, reject <var>p</var> with a new
-              <code>MediaStreamError</code> with name
-              <code>ConstraintNotSatisfied</code> and
-              <code>constraintName</code> set to any of the required
-              constraints that could not be satisfied, and abort these steps.
-              <var>existingConstraints</var> remain in effect in this
-              case.</li>
+	      <li><p>For each <code>settings dictionary</code> possible for <var>copy</var>,
+		compute its <code>fitness distance</code>.
```
fluffy: I think this should be phrased as "For every possible setting dictionary, compute its fitness distance." It would be nice to somehow make it clear in the case of resolutions and aspect ratio, this mean the set consists of all possible combinations of height, width, and aspect ratio.  Somewhere up at the front of this might be good to mention that we don't think people will actually implement this algorithm but will implement an more optimized algorithm that generate the same outcome. I would describe this algorithm as optimization by "Generate all possible solutions, filter out the ones that are not viable, then compute a fitness for each one and pick the best".

fluffy:
I like the new formulation based on optimizing the fitness function better than the old. I think a few teaks should get this in good shape to ship publish.

fluffy:
End of my comments for today :-)

getusermedia.html
alvestrand: This was in response to Dan's comment about wanting to say that separating the settings-picking algorithm from the device-picking doesn't prevent the UA from using information that's only mentioned in the description of the settings-picking algorithm as part of the device-picking process.
The paragraph actually doesn't mention the possibility of pre-stored permission, which I think is a bug (but one that is independent of this PR).

getusermedia.html
alvestrand: When actual and ideal are both zero, actual == ideal, and the result is zero - the ?: operator never executes its third part.
I don't see an issue here (except perhaps that we should explicitly outlaw negative values; actual=0 and ideal=-1 would give a division by zero).

getusermedia.html
alvestrand: The note you ask for is actually right after the algorithm. Do you think we need to move it to the front?
It seems odd to say this before describing the thing we're talk about.

I made the wording change you suggested.

alvestrand:
Applied some fixes based on Cullen's comments, and rebased to current master.
The usual confusion occured - I ended up pushing the result with git push --force.
PTAL - I think this is ready to merge.

getusermedia.html
fluffy: works for me

alvestrand:
This PR is obsoleted by #122 - same code but a cleaner history.

PR #61, opened 2014-11-05
Rewrite the definition of constraint

martinthomson:
I couldn't make a lot of sense of the original.  I suspect that it's rotted a little over time.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2550,7 +2550,7 @@
             </code> interface.</dd>
 
             <!-- I have no idea why the &#8288; are needed but respec is broken
-          and does not display stuff correctly if you remove them --> 
+          and does not display stuff correctly if you remove them -->
           <dt>Promise&lt;&#8288;sequence&lt;&#8288;MediaDeviceInfo&#8288;&gt;&#8288;&gt; enumerateDevices ()</dt>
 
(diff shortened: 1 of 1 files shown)


fluffy:
Seems reasonable

jan-ivar:
This language is definitely in the right direction, but still lacks an unequivocal statement that constraints are properties of the caller. They're demands, or asks. We need language guided by clear definitions here or readers wont stand a chance. For instance, "constraints" are *not* a feature, nor IMHO the (data)type for things like "width", "height", frameRate" etc. Those are capabilities that the UA implements and lets callers place constraints on. A caller's constraints are expressed as limits put on individual capabilities of the browser.

For instance, "Constraints are exposed on tracks", sounds like inherent track-information is revealed to callers, when this information comes *from* callers. Tracks merely remember the constraints put on them by callers, as a convenience, and what you already know can't be "exposed" to you.

See also my comment on #116 where I link to (admittedly scant) list-support for focusing on capabilities in the gUM narrative.

burnburn:
I have made changes which attempt to address the points raised by the original PR and in the discussion, in some cases using the provided text.  I am closing this PR as addressed.  Further suggestions can go into new pull requests.

Issue #69, opened 2014-11-08
constraint properties should not be scoped to "source" only


ShijunS:
[Section 14.1] In the “video” property table, the spec has the following for the "width" property, “The width or width range, in pixels, of the video source.”

Since constraints can be applied to track (not directly to the “source”), I’d suggest removing “of the video source”, so the definition becomes, "The width or width range in pixels."

There are two other places in the table, i.e. for "height" and "frameRate", which have the same issue.

PR #86, opened 2014-11-24
removed iana registration

### LINKED ISSUE #30: The iana reg draft points at wrong draft now

dontcallmedom:
instead describe members of MediaTrackConstraints dictionary
close #30 and #69 (supersedes #84)


alvestrand:
The chairs have concluded that we do not have consensus to change this approach.

For the Last Call version, we'll keep the IANA registration.

PR #98, opened 2014-11-27
Add variadic MediaStream constructor

martinthomson:
As a usability measure.  I'd remove the `sequence<MediaStreamTrack>` variation if I could, but I don't think that is possible right now.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -373,6 +373,14 @@
                 the <code><a>MediaStreamTrack</a></code> objects in the provided
                 sequence.</p>
               </li>
+
+              <li>
+                <p>One or more <code><a>MediaStreamTrack</a></code> objects as
+                variadic list:</p>
+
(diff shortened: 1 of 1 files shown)


alvestrand:
@dontcallmedom do you know if there is any architectural guidance for how one should do this - passing in an array or making a variadic constructor?

dontcallmedom:
I don't think there is any particular architectural guidance on this, it is mostly an API usability question. Given that this doesn't remove the constructor with an array of tracks, I would side with keeping it simple and not adding another variation to the constructor (which as @adam-be points out is easily replicated in code if needed)

adam-be:
An argument against an variadic argument list is that it not as straight forward to extend this constructor with additional arguments in the future (as pointed out by @burnburn offline).

I'm closing this one. We can always reopen if we change our minds.

PR #111, opened 2015-01-07
using a single type of quote and apostrophe

dontcallmedom:
see also https://github.com/jsoref/mediacapture-main/commit/3821459f4ad52e12ccb2d7fb4509062cc7d7697f#commitcomment-9165769


adam-be:
Thanks Dom

PR #112, opened 2015-01-08
constraint properties should not be scoped to 'source' only; close #69

### LINKED ISSUE #69: constraint properties should not be scoped to "source" only

dontcallmedom:


alvestrand:
Not sure this makes sense as proposed.
In every case, the term "video source" appears in the very next sentence.

For some properties (width and height especially), we have the idea that the UA can provide a different value to the track than the source has, so the property belongs to the track not the source. But to reflect that properly, we have to rewrite sentence 2 of the definition too.

dontcallmedom:
The second sentence of these definitions talks about **capabilities**, which *are* linked to a source; it’s only when talking about constraints in general that referring to sources doesn’t make sense as far as I understand.

jan-ivar:
@dontcallmedom I agree. Constraints are properties of the caller, not the browser. They're demands, or asks. I had someone just today who was confused about this, so it's still not clear enough I think. How about:

s/, of the video source/, requested by the caller/

?

burnburn:
I will merge this pull request as written and then adjust the text beginning this section to clarify that we are defining constrainable properties.  I will need to adjust some wording similarly in the constraints registry draft to clarify that we are defining properties which may have settings, capabilities, and constraints.

PR #117, opened 2015-01-09
group agreed to keep this doc to process 2005 until we reach last call

dontcallmedom:
see https://lists.w3.org/Archives/Public/public-media-capture/2014Dec/0033.html


fluffy:
LGTM

PR #119, opened 2015-01-28
reword event handler descriptions to make their circumstances clearer

### LINKED ISSUE #103: Improve description of EventHandler attributes
Statements of this form are almost completely devoid of information:

> This event handler, of type active, MUST be supported by all objects implementing the MediaStream interface.

A simple rewrite makes a big difference:

>  This event handler, of type active, is fired when the MediaStream becomes active.

And this is throughout.

Chasing down what the conditions are for firing an event is surprisingly difficult without something like this.

As for what I recommend removing, I have no idea why you would have strict normative requirements around support for this and not other things.

dontcallmedom:
close #103

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -687,9 +687,7 @@
           <dd>
             <p>This event handler, of type <code>
               <a href="#event-mediastream-active">active</a></code
-            >, MUST be supported by all objects implementing the <code>
-              <a>MediaStream</a>
-            </code> interface.</p>
+            >, is fired when the MediaStream becomes active.</p>
           </dd>
 
           <dt>attribute EventHandler oninactive</dt>
@@ -697,9 +695,7 @@
           <dd>
             <p>This event handler, of type <code>
               <a href="#event-mediastream-inactive">inactive</a></code
-            >, MUST be supported by all objects implementing the <code>
-              <a>MediaStream</a>
-            </code> interface.</p>
+            >, is fired when the MediaStream becomes inactive.</p>
           </dd>
 
           <dt>attribute EventHandler onaddtrack</dt>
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -3810,13 +3803,10 @@
...
-              <a>ConstrainablePattern</a>
-            </code> pattern. <p>The User Agent MUST raise a <code>MediaStreamErrorEvent</code>
-          named "overconstrained" if changing circumstances at runtime result
-          in it no longer being able to satisfy the
-          <var>requiredConstraints</var> from the currently valid Constraints.
-          This MediaStreamErrorEvent MUST contain a MediaStreamError whose
+          fires a <code>MediaStreamErrorEvent</code> when the User Agent is
+          no longer able to satisfy the <var>requiredConstraints</var> from
+          the currently valid Constraints.</p>
+          <p>This MediaStreamErrorEvent MUST contain a MediaStreamError whose
```
stefhak: It seems the rest of this PR works to get rid of unnecessary (uppercase) "MUST"s since the text is normative anyway. Get rid of this one too?

stefhak:
LGTM. One minor comment (above).

stefhak:
LGTM.

burnburn:
Two comments:
1. Some of these, particularly the ones beginning with 'mute', are actually providing definitions here that may already exist elsewhere.  Please confirm.  Alternatively, separate each of these so we can discuss individually.
2. The wording needs improvement, both here and in other places.  Events fire, and event handlers are executed when events fire.  The wording here and elsewhere needs to be fixed.

dontcallmedom:
re 1., the definitions are replicate of what is in the "events summary table; I feel that's a reasonable duplication
re 2; I'm happy to provide an updated PR that says executed instead of fired; will do once I get confirmation it's worthwile

burnburn:
Dom,

Yes, this is worthwhile.  Your updates from 'fired' to 'executed' are good.  Thank you.  Can you please update the onoverconstrained section as well?  This requires both the change from firing to executing (sounds violent either way, doesn't it?) and also a statement of what information is passed into the handler, not what is in the event.  They are of course related.

burnburn:
Thanks.  Merging.

PR #122, opened 2015-01-29
Constraint algorithm rewrite.

alvestrand:
This change uses a single constraint evaluation algorithm for both
device selection in getUserMedia and device reconfiguration in
applyConstraints.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2296,9 +2296,9 @@
         <tbody>
           <tr>
             <td>
-             
+
                 <code>active</code>
-            
+
             </td>
 
             <td>
@@ -2313,9 +2313,9 @@
 
           <tr>
             <td>
-             
+
                 <code>inactive</code>
-             
+
             </td>
 
             <td>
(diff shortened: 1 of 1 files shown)


fluffy:
LGTM

getusermedia.html
```diff
@@ -3169,8 +3079,12 @@
                     <a>MediaStreamTrack</a>
                   </code> MUST not change.</p>
 
-                <p class="issue">Define the event that should be raised when
-                the User Agent changes its choice of track.</p>
+		<p>The user agent MAY use the value of the computed "fitness
+		  distance" from the <a>SelectSettings()</a> algorithm, or
+		  any other internally-availble information about the
```
jan-ivar: typo (sorry just got around to reading this)

getusermedia.html
```diff
@@ -3707,52 +3671,29 @@
...
-              that 'advanced' is not a member of either of these sets since it
-              represents meta-information about constraints. In the following
-              we treat <var>requiredConstraints</var> and
-              <var>unorderedConstraints</var> as ConstraintSets. Note that
+	    <p>
+	      We define the function <dfn>SelectSettings()</dfn> as follows:
+             <ol>
+              <li>Each constraint MUST
+              specify one or more values (or a range of
+              values) for its property. A property MAY appear more than once
```
jan-ivar: (Not a change with this patch, but) I think the nomenclature is that dictionaries have members, not properties [1]. This seems to be a problem throughout this spec.

[1] https://heycam.github.io/webidl/#idl-dictionaries

getusermedia.html
```diff
@@ -3707,52 +3671,29 @@
...
-              <var>newConstraints</var>. Let <var>unorderedConstraints</var>
-              be all the other constraints in <var>newConstraints</var>. Note
-              that 'advanced' is not a member of either of these sets since it
-              represents meta-information about constraints. In the following
-              we treat <var>requiredConstraints</var> and
-              <var>unorderedConstraints</var> as ConstraintSets. Note that
+	    <p>
+	      We define the function <dfn>SelectSettings()</dfn> as follows:
+             <ol>
+              <li>Each constraint MUST
```
jan-ivar: Step 1 reads like notes that should come before or after, rather than a step, as there's nothing to do. It feels like an abrupt introduction to the whole thing and the SelectSettings() "function" which appears to be a unique construct of a set of sub-steps. This "function" isn't well defined, leaves inputs and output to be inferred from the text, and ends up referring to things that haven't been introduced yet, like "Each constraint".

Step 2 (not visible in patch) refers to "this method" which seems to be a reference to applyConstraints and/or getUserMedia (the two places that invoke it)? Ironically, applyConstraints, in its step 2, refers to itself as "this function".

I like the idea of reusing sub-steps, but might it give readers a softer introduction to define applyConstraints first and SelectSettings() second? Then the out-of-scope references might work, much like closures in JS gives a sub-function access to the scope outside it. getUserMedia and applyConstraints are similar enough methods on the same object, that it might just work.

jan-ivar: Hmm, I see my mistake was not reading all the definitions linearly in one gulp, and that "constraint" is indeed defined in an earlier definition. However, SelectSettings() is referenced multiple places in the document, so I think it'll be common to jump to it and start reading, like I did.

getusermedia.html
alvestrand: I disagree with the interpretation. A constraint specifies one or more values (or a range of values) for the property that it constrains.
A dictionary is the representation used for the constraint.
The constraint value may be (for instance) { min: 24, max: 48 }, which may be stored in the corresponding member of the dictionary (named with the property's name) used to represent the constraint set.
Using the term "member" for the concept of "one or more values for its property" would not improve the clarity of the spec.
I'm sure clarity can be improved, but don't agree with this proposal.

getusermedia.html
alvestrand: People who jump around will have to take their licks.
We should probably add some more hotlinks so that it's obvious how to get to the definition of "constraint" from the usage in this paragraph.

Please file a bug on this one.

getusermedia.html
jan-ivar: Ah, thanks for clearing up my misunderstanding. It's the thing we're trying to control. In that case I think I would prefer "capability" over "property", since that seems consistent with our existing language about this control surface.

jan-ivar:
Filed. Would you like issues for the "this method" and "this function" bits as well?

PR #124, opened 2015-02-02
Tweaks to new apply constraints algo

dontcallmedom:


getusermedia.html
```diff
@@ -3765,7 +3765,7 @@
               </li>
 	      <ol>
 		<li><p>Let <var>successfulSettings</var> be the result
-		    of running <a>SelectSettings(newConstraints)</a>.</p>
+		    of running the <a>SelectSettings</a> algorithm with <var>newConstraints</var> as the constraint set.</p>
 		<li><p>If <var>successfulSettings</var> is undefined,
```
alvestrand: Should there be <code> around this "undefined" too?

getusermedia.html
```diff
@@ -3043,10 +3043,10 @@
                   <var>constraints</var>.</li>
 
 		  <li><p>Run
-		      the <code><a>SelectSettings()</a></code>
+		      the <code><a>SelectSettings</a></code>
 		      algorithm on each track
 		      in <var>CandidateSet</var>. If the algorithm
```
alvestrand: Since we're being precise about this elsewhere, should we also add which constraints we're passing as the argument to SelectSettings?

alvestrand:
Looks good to me. Dom, you're a better markup expert than I am!

stefhak:
Looks good to me too.

PR #126, opened 2015-02-04
Fix language around advanced algorithm to be unambiguous.

jan-ivar:
Remove redundant steps in advanced algorithm, and clarify that fitness distance from this step is not used later.

Also, replace "SHOULD use the one with a smaller fitness distance" - because there's generally no stand-out small one - with "SHOULD use the one with the smallest fitness distance", the likely meaning. The alternative interpretation would be "SHOULD use one with a smaller fitness distance", both mild suggestions of varying precision.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3615,7 +3615,6 @@ if(!supports["width"] || !supports["height"]) {
               <a>settings dictionary</a> and a constraint set
               <var>CS</var> as the sum, for each constraint
               provided for a constraint name in <var>CS</var>,
-	      ignoring "advanced",
 	      of the following values:</p>
 
             <ol>
@@ -3715,7 +3714,7 @@ if(!supports["width"] || !supports["height"]) {
 		compute its <a>fitness distance</a>.
 	      Discard the <a title="settings dictionary">settings dictionaries</a> for which the result is
 	      infinity, and add the remaining <a title="settings dictionary">settings dictionaries</a> to the set
-	      of possible settings dictionaries <var>candidates</var></p></li>
+	      of possible settings dictionaries <var>candidates</var>.</p></li>
 
 	      <li><p>If the <a>fitness distance</a> of the best fitting
 		  settings dictionary is infinity, return <code>undefined</code> as the result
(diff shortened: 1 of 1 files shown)


alvestrand:
This PR removes the computation of "requiredConstraints". It seems that previous changes lost the setting of "requiredConstraints" at constraint application (instead we store the whole set of constraints that were passed in, we don't mark up which ones succeeded and which ones were ignored).

If we want to do this, we need to remove the remaining 2 occurences of RequiredConstraints too.

burnburn:
Actually, there are several problems with the combined 'fitness distance' calculation text and the SelectSettings algorithm.  In the former 'advanced' is to be ignored, but step 5 in the latter specifically calls for computation of the fitness distance using the 'advanced' ConstraintSets.  Also, in the latter we have
'3. For every possible settings dictionary of copy, compute its fitness distance. Discard the settings dictionaries for which the result is infinity, and add the remaining settings dictionaries to the set of possible settings dictionaries candidates

4. If the fitness distance of the best fitting settings dictionary is infinity, return undefined as the result of the function.'
which is clearly contradictory as written, since all infinite fitness distance settings dictionaries are removed in step 3.

I found it difficult to be sure that required constraints have been properly addressed without the current text.  I suspect that the best solution for now is to reintroduce the accidentally missing text on the initial setting of RequiredConstraints.

jan-ivar:
The fitness distance algorithm conflates two things (and the spec is better for it imho): 1) weeding out sources that don't meet required constraints, and 2) scoring finite sources against an ideal.

Therefore, advanced must be part of the algorithm. I don't see a flaw in the SelectSettings() algorithm other than that it could be a lot clearer about that finite distances in step 5 are not to be used for anything.

That's why this PR suggests adding the text it does ("The UA SHOULD use the one with the smallest <code>fitness distance</code>**, as calculated in step 3**.")

jan-ivar:
On Step3/4 I agree it is redundant. it combines two ways of reasoning about the same thing. I'd be fine in going either direction. (musings: Ironically, I implemented this part quite literally where I remove elements from an array (not very efficient). I'm reconsidering this to possibly leave candidates but since they start out with infinite distance, I'd need to track unset!=infinite.)

On RequiredConstraints, what clarity is lacking? I don't know what RequiredConstraints did, but since it is redundant in the algorithm, I hope we can solve it without needlessly complicating the algorithm.

jan-ivar:
Actually, re-reading Step 4 I see what it means. Saying that best-fitting == infinity is one way of saying none of them fit, i.e.empty set. I'm fine either way on that one.

jan-ivar:
Updated with changes from conversation with @burnburn just now, and rebased against master (hence why commits show in the [wrong order](https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/)).

PR #129, opened 2015-02-06
define fitness distance when ideal is not specified

### LINKED ISSUE #125: What's the fitness distance if no ideal is specified?
I assume 0 but I couldn't find any text on this.

dontcallmedom:
close #125

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3643,6 +3643,10 @@ if(!supports["width"] || !supports["height"]) {
 		  satisfy the constraint, use positive infinity.</p>
               </li>
               <li>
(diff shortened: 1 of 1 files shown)


dontcallmedom:
I think my latest commit makes it clear that each set is a bailing step, and that the value obtained is the fitness distance; I don't think making the formula use `!ideal` makes the algorithm clearer, so I'd rather keep the prose description.
(as for the line break / word-wrap, it's independent of my PR :)

stefhak:
LGTM.

burnburn:
LGTM

alvestrand:
LGTM.

Issue #114, opened 2015-01-09
 DeviceID, kind and label needs to be aligned with capabilities


dontcallmedom:
*Initially [reported in bugzilla](https://www.w3.org/Bugs/Public/show_bug.cgi?id=26954) by Harald Alvestrand 2014-10-02 14:36:10 UTC*

The document needs to be reviewed such that we align deviceId, kind and label with the new definition of capabilities.

dontcallmedom:
*Comment by Harald Alvestrand 2014-11-20 15:42:27 UTC*

This will be done as part of a general consistency review.

burnburn:
I had initially requested this review.

Upon reviewing, the usage of deviceId, kind, and label appear to be fine now.  However, in examining sourceType I found the following:

In section 4.3.2 'Tracks and Constraints', it says

"When applyConstraints() is called, a User Agent must queue a task to evaluate those changes when the task queue is next serviced. Similarly, if the sourceType changes, then the User Agent must perform the same actions to re-evaluate the constraints of each track affected by that source change."

The second sentence doesn't make sense to me.  The sourceType for a track already associated with a given source will never change.  If the sourceType constraint is set on a track (via applyConstraints) to a different value than than that of the track's source, then the track will immediately become overconstrained.  I would like clarification from anyone who understands what this sentence was intended to convey.

dontcallmedom:
I think this dates from a time where our model of constraints and sources was still fuzzy; 
in fact, https://github.com/w3c/mediacapture-main/commit/6198d1660936e0d931c609f1ee84309e626f83c6#diff-ea76d38900f79cfae8f60e5f7cf16dd1R666 says it comes unchanged from [the original setting proposal from Travis](https://dvcs.w3.org/hg/dap/raw-file/tip/media-stream-capture/proposals/SettingsAPI_proposal_v6.html#constraint-manipulation-api-extensions-to-mediastreamtrack) (via yourself :) which I have just confirmed, so I think it's fine to remove this sentence now — I'll make a PR to that effect.

I'll note that @adambe at expressed a similar concern in his (later reverted) commit  https://github.com/w3c/mediacapture-main/commit/71f3030cdd686f84bd6decb895d2196c13092bcd

burnburn:
Thanks Dom.  I'll let the other editors skim this as another sanity check.  Assuming no objections we'll be able to merge on Thursday.

alvestrand:
[...]

PR #131, opened 2015-02-09
remove sentence about sourceType change (doesn't happen)

dontcallmedom:
see https://github.com/w3c/mediacapture-main/issues/114#issuecomment-73332618


stefhak:
LGTM.

burnburn:
This is probably the right text change to make.  If someone calls applyConstraints() with a different sourceType, the existing constraint application algorithm will do the right thing.

Issue #128, opened 2015-02-06
 Fitness distance must interpret plain values in advanced array as exact.


jan-ivar:
Right now, the advanced array in [Example 3](http://w3c.github.io/mediacapture-main/getusermedia.html#constrainable-interface) does squat, unless it's re-written as:

    var constraints =
    {
      width: {min: 640, ideal: 1280},
      height: {min: 480, ideal: 720},
      advanced: [{width: {exact: 1920}, height: {exact: 1280 } },
                 {aspectRatio: {exact: 1.3333333333} }]
    };

dontcallmedom:
shouldn't `aspectRatio` also use `exact:`?

jan-ivar:
Yes, thanks, I've updated my comment above. The point though, so that there is no confusion, is that we should not alter the example but the algorithm. :-)

dontcallmedom:
oh, good you clarified, I was going to update the example :)

alvestrand:
Sent a note on this (together with #132) to the list. Awaiting comments.

stefhak:
The list input so far (in the thread starting at http://lists.w3.org/Archives/Public/public-media-capture/2015Feb/0018.html) indicate that bare values in the "advanced" section means "exact".

burnburn:
Addressed in commit af37807bb453969d50a0b0aee28851318931eacd

PR #132, opened 2015-02-09
clarify how a constraint is read

dontcallmedom:
* split up big block of text
* add "ReadConstraint" algorithm
* distinguish behavior in normal constraint vs advanced (cf https://github.com/w3c/mediacapture-main/issues/128)
* remove duplicate note about unknown properties

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3386,31 +3386,69 @@
 
       <p>The application can select the (range of) values it wants for an
       object's Capabilities by means of basic and/or advanced ConstraintSets
-      and the <code>applyConstraints()</code> method. A ConstraintSet consists
+      and the <code>applyConstraints()</code> method.</p>
+
+      <p>A ConstraintSet consists
       of the names of one or more properties of the object plus the desired
       value (or a range of desired values) for each property. Each of those
-      property/value pairs can be considered to be an individual constraint.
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -3386,31 +3386,69 @@
...
+      <p>The behavior is as follows:</p>
+      <ol>
+        <li>all <var>min</var>, <var>max</var>, and <var>exact</var>
+      values in a constraint are together treated as
+      the <em>required</em> set</li>
+        <li>if it is not possible to satisfy simultaneously
+          all of those individual constraints for the indicated property names,
+          the User Agent MUST reject the returned promise.</li>
+        <li>Otherwise, it must
+      apply the required constraints.
```
alvestrand: I think this bullet should be removed. Applying the constraints comes much later.

dontcallmedom: Note that this text comes from the original text as is (the bullets just make it more visible)

dontcallmedom:
I think @jan-ivar is right that this needs more work, and it's not clear this pull request is necessarily the best basis for that. I unfortunately won't have cycles to work on this myself for a while.

alvestrand:
Linking to issue #128 which this fixes.

alvestrand:
Closing, this will be addressed elsewhere.

PR #135, opened 2015-02-10
avoid MUST for non-UA requirement in SelectSettings definition.

jan-ivar:


burnburn:
Yes, I believe that enforcement of providing a value here is already handled by the syntax itself.  Harald, Adam, can you please confirm?

alvestrand:
I don't think this matters either way, but I think the requirement actually does something. I don't think the syntax enforces that you can't write 'height: {}' (empty dictionary).

fluffy:
don't care one way or the other

alvestrand:
No, we won't integrate this.

jan-ivar:
"Each constraint MUST specify one or more values," or else what? Throw? Please explain.

alvestrand:
That's a bug waiting to be filed - the action to be taken on constraints { height: {} } needs to be clear from the spec. Filed #139.

PR #136, opened 2015-02-11
Markup fixes

dontcallmedom:
* Various markup fixes needed to make the document publication ready
* Travis CI rule to keep us from breaking internal links


burnburn:
Is it okay these days to leave out the xml namespacing (for the doc) and xml white space preservation (for the examples)?  I'm fine with it -- just asking.

dontcallmedom:
yes, since we're using HTML5, leaving out the xml namespace and the xml:space attribute is the right thing to do.

stefhak:
I can't see any functional changes, and trusting that Dom knows what we need to be publication ready, I say: LGTM.

PR #144, opened 2015-03-10
getSupportedConstraints as list of values

### LINKED ISSUE #66: should getSupportedContraints be static?
If getSupportedContraints is "static" as in the current draft, all the examples need to be fixed.

For example, 
var supports = navigator.mediaDevices.getSupportedConstraints("video");

should be changed to 
var supports = MediaDevices.getSupportedConstraints("video");

### LINKED ISSUE #116: WebIDL types needed in Constrainable application

dontcallmedom:
alternative to #143
close #116
close #66

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3015,15 +3015,13 @@
         recognizes.</p>
 
         <dl class="idl" title="partial interface MediaDevices">
-          <dt>static Dictionary getSupportedConstraints(DOMString kind)</dt>
+          <dt>DOMString[] getSupportedConstraints(DOMString kind)</dt>
 
           <dd>
-            <p>Returns a dictionary whose members are the constrainable properties
-            known to the User Agent for the kind given as argument. A
-            supported constrainable property MUST be represented by a member whose name is
-            the constraint name and whose value is <code>true</code>. Any
+            <p>Returns a list of the names of constrainable properties
+            known to the User Agent for the kind given as argument. Any
             constrainable properties not supported by the User Agent MUST not be
-            present in the returned dictionary.</p>
+            present in the returned array.</p>
           </dd>
 
           <dt>Promise&lt;MediaStream&gt; getUserMedia(


adam-be:
Too bad the indexOf() != -1 test is so ugly. If we had an "in" keyword that could check if an element was in a list then I think this approach would be quite OK. Because all we want to know is if the browser knows a particular name (which happens to represent a constraint).

jan-ivar:
I think we should be more explicit and correct about types: We want to know whether it *implements* a particular *dictionary member*. Wherever the spec reverts to talking about "names" represents a breakdown of types to me. What exactly are the pros of using an array of strings over using the dictionary in question?

adam-be:
The benefit would be that we wouldn't need to define a new type with the sole purpose that if a member is present on one dictionary, you can use that member name in an other dictionary successfully. The fact that the name of a member is present in a list of strings gives you the same information as the fact that a dictionary member exist on a dictionary (if you disregard from a potential value).

jan-ivar:
But we'd need to define a new string with the sole purpose being that if a string is present in an array, then you can use that string as a dictionary member in other dictionaries successfully.

Does the "benefit" hinge on strings being easier to manage than dictionary members? If so then I don't see any benefit.

In any case I don't think we should prioritize specifiers/implementers over users/authors in the [order of constituencies](http://www.w3.org/TR/html-design-principles/#priority-of-constituencies).

fluffy:
took other option

PR #145, opened 2015-03-10
clarify enumerateDevices access model

### LINKED ISSUE #142: enumerateDevices' access model is unclear
From [the spec](http://w3c.github.io/mediacapture-main/getusermedia.html#methods-2):

> If none of the local devices are attached to an active MediaStreamTrack, let filteredList be a copy of resultList, and all its elements, where the label member is the empty string.

Taken literally, this exposes the label to all concurrently open tabs during active gUM use in *any* tab regardless of site, which is not what we mean, because [it further says](http://w3c.github.io/mediacapture-main/getusermedia.html#access-control-model):

> The algorithm described above means that the access to media device information depends on whether or not permission has been granted to the page's origin to use media devices.

This narrows access down to a page's origin, but is fuzzy about the duration of access, wrt whether access is afforded outside active use (MediaStreamTracks are stopped), or when persistent permission has been (previously) granted.

These statements seem quite contradictory in their description about what matters for access, and neither matches gUM permissions which are afforded, not even per page, but, *per instance* and not per origin (unless permanent access is granted). 

Questions:
1. Open same jsfiddle in two tabs, grant access in one and deny in the other. Who sees label?
2. Open different jsfiddles in two tabs, grant access in one and deny in the other. Who sees label?
3. Stop the MediaStream. Can you still see label?
4. Never start a MediaStream, but persistent permissions were granted on a previous visit. See label?
5. Never start a MediaStream, but persistent permissions were granted on a previous visit to another page on the same host. See label?

dontcallmedom:
close #142

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2743,7 +2743,11 @@
 
               <li>
                 <p>If none of the local devices are attached to an active
-                <code><a>MediaStreamTrack</a></code>, let
+                <code><a>MediaStreamTrack</a></code> in a browsing
+                context from the page's origin, and if
+                no <a href="stored-permissions">persistent
+                permission</a> to access these local devices have been
+                granted, let
                 <var>filteredList</var> be a copy of <var>resultList</var>, and
                 all its elements, where the <code><a href=
                 "#widl-MediaDeviceInfo-label">label</a></code> member is the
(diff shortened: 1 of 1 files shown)


alvestrand:
This (which only concerns access to device information, not to the devices themselves) looks good to me with the change suggested by Jan-Ivar.
It was clearly not clear enough before what was expected - probably because the cases hadn't been worked through.

PR #146, opened 2015-03-17
WebIDL for supported constraints

dontcallmedom:


getusermedia.html
```diff
@@ -1319,6 +1319,62 @@
...
+           list of constraints recognized by a User Agent for controlling the
+           <a>Capabilities</a> of a <code><a>MediaStreamTrack</a></code> object.
+          </p>
+
+          <dl class="idl" title="dictionary MediaTrackSupportedConstraints">
+            <dt>boolean width = true</dt>
+
+            <dd />
+
+            <dt>boolean height = true</dt>
```
alvestrand: what do the "=true" add?

If "true" imeans "it's supported", and "=true" means that it is assumed to be supported if omitted, I think this is the wrong value - the default should be "false".

If the UA doesn't tell you it support frobnicator, one should assume it doesn't support frobnicator.

getusermedia.html
```diff
@@ -3019,15 +3075,15 @@
         recognizes.</p>
 
         <dl class="idl" title="partial interface MediaDevices">
-          <dt>static Dictionary getSupportedConstraints(DOMString kind)</dt>
+          <dt>MediaTrackSupportedConstraints getSupportedConstraints(DOMString kind)</dt>
 
           <dd>
-            <p>Returns a dictionary whose members are the constrainable properties
+            <p>Returns an object whose properties are the constrainable properties
```
alvestrand: The IDL says "dictionary with members", the text is changed to "object with properties". I'm confused.

getusermedia.html
dontcallmedom: if you support a particular constraint, the only logical value (and thus its default) for the matching member in the dictionary is true; since per the spec you're expected to support all the constraints, I think using the default value of true make sense.
But we can leave with no default if this is likely to create more confusion.

getusermedia.html
dontcallmedom: dictionary and member describes WebIDL entities, object and properties describe JavaScript entities; since we're describing the JavaScript behavior, I think using JavaScript entity names make more sense. But again, I'm happy to revert if needed.

jan-ivar: I have to agree with Harald. WebIDL defines dictionaries for both inputs and outputs, and the spec is described in WebIDL not JavaScript. We use dictionary/member elsewhere, and these are dictionaries.

getusermedia.html
jan-ivar: I agree with no default. These dictionaries are only ever filled out by UAs and only ever read by users, so what the default is matters little. The only benefit of having a default is that it gives a guarantee to readers that the member will be there. However, this guarantee is absent where it matters for this API, which is dealing with older browsers that don't have the member defined in WebIDL.

adam-be:
This adds a fourth object that needs to be updated when a new constraint is added. I know there are a lot of opinions about that, but this PR is in line with the approach we've taken so far with "tailored" objects for each type of usage.

PR #150, opened 2015-03-19
Update text for how constraint dictionaries get extended 

fluffy:


getusermedia.html
```diff
@@ -3083,7 +3104,9 @@
             supported constrainable property MUST be represented by a member whose name is
             the constraint name and whose value is <code>true</code>. Any
             constrainable properties not supported by the User Agent MUST not be
-            present in the returned dictionary.</p>
+            present in the returned dictionary. The values returned represent
+            what the browser implements and will not change during a single
```
stefhak: I think we could remove the word "single".
(BTW, we currently use "session" 5 times in the spec without really defining it, but it is probably well enough understood for this to work)

getusermedia.html
```diff
@@ -1327,6 +1327,11 @@
            <a>Capabilities</a> of a <code><a>MediaStreamTrack</a></code> object.
           </p>
 
+          <p> Future specification extend the MediaTrackSupportedConstraints 
+          dictionary by defining a partial dictionary with DictionaryMembers of Type
```
stefhak: s/Future specification extend/Future specifications can extend/ ?

stefhak:
I made a couple of minor comments, but overall this LGTM.

getusermedia.html
alvestrand: Arguing from context for the term "session history" in http://www.w3.org/TR/2011/WD-html5-20110113/history.html#history it seems that "session" should be considered equivalent to "browsing context":

But I suspect that the term is used more synonymously with "top-level browsing context".

http://www.w3.org/TR/2011/WD-html5-20110113/browsers.html#browsing-context

Defining it by reference would be a Nice Thing (but not necessary for this pull request).

alvestrand:
LGTM. It does not address the question of browser conformance, so I regard this as recording consensus.

adam-be:
s/DictionaryMembers/dictionary members/

otherwise LGTM

getusermedia.html
burnburn: I agree with this suggestion (applied to all four text sections).

fluffy:
OK - I think I made all the sugested  changes

stefhak:
Thanks Cullen. LGTM.
@adam-be : will you merge so we can have the editing team review before doing a dated draft?

adam-be:
Merged and added change log entry

fluffy:
thank you

PR #151, opened 2015-03-31
Remove outdated issues in the doc

dontcallmedom:
(3 other issues are still in the doc after this; separate pull requests to come)


alvestrand:
@burnburn please check if those are the same issues you found.

burnburn:
Agree that these three can go.

PR #152, opened 2015-03-31
using "browsing session" instead of "session"

dontcallmedom:
("browsing session" is already used e.g. in geo, powerful-features)

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -159,12 +159,7 @@
 
         <p>Source capabilities are effectively constant. Applications should be
         able to depend on a specific source having the same capabilities for
-        any session.</p>
-
-        <div class="note">
-          <p>Open Issue: Is "session" the correct term? Should it be "top-level
-          browsing context" as defined in HTML spec?</p>
-        </div>
+        any browsing session.</p>
 
         <p>This API is intentionally simplified. Capabilities are not capable
         of describing interactions between different values. For instance, it
@@ -2425,7 +2420,7 @@
 
                 <li>
                   <p>If this method has been called previously within this
-                  application session, let <var>oldList</var> be the list of
+                  browsing session, let <var>oldList</var> be the list of
                   <code><a>MediaDeviceInfo</a></code> objects that was produced
                   at that call (<var>resultList</var>); otherwise, let
                   <var>oldList</var> be an empty list.</p>
@@ -2527,7 +2522,7 @@
           <p>A unique identifier for the represented device.</p>
 
           <p>All enumerable devices have an identifier that MUST be unique to
-          the application and persistent across sessions. Unique and stable
+          the application and persistent across browsing sessions. Unique and stable
           identifiers let the application save, identify the availability of,
           and directly request specific sources.</p>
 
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -159,12 +159,7 @@
 
         <p>Source capabilities are effectively constant. Applications should be
         able to depend on a specific source having the same capabilities for
-        any session.</p>
-
-        <div class="note">
-          <p>Open Issue: Is "session" the correct term? Should it be "top-level
-          browsing context" as defined in HTML spec?</p>
-        </div>
+        any browsing session.</p>
```
alvestrand: Is the term "browsing session" well defined, and if so, is there a reference we can point at?
"Top level browsing context" is a term defined in HTML5, but I couldn't find the term "session" there.

jan-ivar: Does it matter what kind of sessions persistence is unaffected by?

alvestrand: Yes; we can't implement the algorithm described here without knowing whether we're in the same session or not.

jan-ivar: See my comments on algorithm in https://github.com/w3c/mediacapture-main/issues/155

dontcallmedom: @alvestrand I haven't found the term "browsing session" formally defined (and it doesn't appear in the HTML5 spec), but it is used in the [geolocation API](http://www.w3.org/TR/geolocation-API/#privacy_for_uas) (with an inline definition of "the time [after which] the browsing context is navigated to another URL"), the ["powerful features" doc](http://www.w3.org/TR/powerful-features/#is-feature-powerful), the [Push API](http://www.w3.org/TR/push-api/#security-and-privacy-considerations).

Given the time-related aspect of what we describe, I don't think top-level browsing context is right (although it is the term I initially thought I would use here).

getusermedia.html
```diff
@@ -4305,10 +4300,10 @@ <h2 id="sec-track-properties">Track Constrainable Property
 
             <td>DOMString</td>
 
-            <td>The application-unique identifier for the source of the
+            <td>The origin-unique identifier for the source of the
             <a>MediaStreamTrack</a>. The same identifier MUST be valid between
-            sessions of this application, but MUST also be different for other
-            applications. Some sort of GUID is recommended for the identifier.
+            browsing sessions of this origin, but MUST also be different for other
+            origins. Some sort of GUID is recommended for the identifier.
```
burnburn: It appears you changed more than "session" -> "browsing session" since you also changed ids from being application-unique to origin-unique.  I *believe* your change is a more accurate description of what we all originally (no pun intended) intended, so I'm okay with it.

dontcallmedom: Oh right, when re-reading the text to use browsing session, using "application" felt jarring there, so I also made that change (but I probably should have made it a separate PR, or at least a separate commit)

jan-ivar: Does the scope of groupIds need to be limited to the browser session?

alvestrand: Let's make that a new issue, or a new pull request with the change if change is needed.

PR #153, opened 2015-03-31
short descriptions of arguments to callbacks in callback-based gUM

dontcallmedom:


alvestrand:
Looks good to me.

burnburn:
These should refer to navigator.getUserMedia() rather than navigator.mediaDevices.getUserMedia() since these are only applicable to the callback version of the API.  Otherwise, thanks for filling these in :)

dontcallmedom:
@burnburn I've used mediaDevices references on purpose — that's where the algorithms are defined nowadays.

alvestrand:
After talking to Dan, I think he's right. We need to point to navigator.getUserMedia for the procedure, because that's where we define what the argument for the callback is - the mediaDevices.getUserMedia definition talks about resolving a promise.

dontcallmedom:
ok, will update

Issue #161, opened 2015-04-15
Remove "Direct Assignment to Media Elements"


annevk:
It conflicts with the definition given in the HTML Standard, which also allows for setting a `Blob` object and such. Given that it's integrated there, providing a pointer seems better.

adam-be:
Yes. That's been our intention since hixie offered to spec it up.

alvestrand:
The text in http://www.w3.org/html/wg/drafts/html/master/single-page.html#dom-media-srcobject seems like an appropriate reference. Assigning to Dom as he is the expert on cross-document references.

PR #165, opened 2015-04-22
defer to HTML5.1 for direct assignment

dontcallmedom:
per #161


alvestrand:
The reference as given is to a nightly build.
There's a WD link possible:

http://www.w3.org/TR/2015/WD-html51-20150417/semantics.html#location-of-the-media-resource

Should we use that (or the undated equivalent) instead?

Also, the pointed-to text is shorter than our text - someone needs to verify that all the text is actually present. @JimBarnett tagged?

dontcallmedom:
I think the nightly draft is OK, but if there is a preference for a more stable draft, I'm happy to update the PR — let me know.

The text is shorter, but the algorithm to load and play media resources includes the stuff about MediaProvider; what I haven't found the time to check is whether the mapping between attributes values and MediaStream specificities has been reproduced in the prose.

JimBarnett:
Dom,
   I looked at the PR, and it looks to me like some of the details about 
the relationship between the MediaStream and the corresponding HTML 
elements are missing.For example, the removed text includes statements 
like:

"For each MediaStreamTrack in the MediaStream , including those that are 
added after the User Agent enters the media element load algorithm, the 
User Agent MUST create a corresponding AudioTrack or VideoTrack as 
defined in [HTML5]. Since the order in the MediaStream 's track set is 
undefined, no requirements are put how the AudioTrackList and 
VideoTrackList are ordered.

The properties of the AudioTrack and VideoTrack objects MUST be 
initialized as follows. Let

AudioTrack.id and VideoTrack.id have the value of the corresponding 
MediaStreamTrack.id attribute
AudioTrack.kind and VideoTrack.kind be "main"
AudioTrack.label and VideoTrack.label have the value of the 
corresponding MediaStreamTrack.label attribute
AudioTrack.language and VideoTrack.language be the empty string"

I don't find anything like that in the HTML text.Perhaps the authors 
have decided that these details fall out from more general statements 
elsewhere in the spec, or they want to leave them up to the 
implementation.It's fine with me if that's the result of a conscious 
decision, but as it stands the text does not specify things to the level 
of detail that the Media Capture group thought was necessary.

- Jim

dontcallmedom:
@JimBarnett the requirement to update the audio/video track lists when a MediaStream is updated is covered by the following text:
> Whenever new data for the current media resource becomes available, queue a task to run the first appropriate steps from the [media data processing steps](http://www.w3.org/TR/2015/WD-html51-20150417/semantics.html#media-data-processing-steps-list) list below.

in the mode "local" of the [resource fetch algorithm](http://www.w3.org/TR/2015/WD-html51-20150417/semantics.html#concept-media-load-resource)

I haven't found any text that deals with setting of id, kind and label attributes to the MediaStreamTrack values — that should probably be filed as a bug on the HTML spec.

stefhak:
I assume the setting if id, kind and label is implicit in the step "Create an Audio/VideoTrack object to represent the audio/video track.", but I think it should be explicit that the id, label and kind is inherited for the MediaStreamTrack, so +1 for filing a bug on HTML spec

alvestrand:
Closing - replaced by #171

Issue #158, opened 2015-04-15
Language does not seem tamper-free


annevk:
E.g. if I overwrite `MediaStream.prototype.addTrack` does that affect `MediaStream`'s constructor? I think both `addTrack()` and the constructor are meant to invoke a "private" algorithm. This happens throughout the specification.

jan-ivar:
Agree. Would you mind pointing to an example that writes it that way?

annevk:
https://dom.spec.whatwg.org/#dom-node-appendchild refers to an internal "append" algorithm which defers to an internal "pre-insert" algorithm. If you click on the definitions of **append** and **pre-insert** you'll find that there's many places throughout the specification that use these algorithms.

adam-be:
Yes. MediaStream.clone() suffers from the same thing (calling MediaStreamTrack.clone()). Let's do a private algorithm.

adam-be:
Proposed fix: PR #167

alvestrand:
Addressed by #167

PR #167, opened 2015-04-24
Refer to interal algorithms instead of API layer methods (Issue #158)

adam-be:
Fixes internal references to MediaStream.addTrack() and MediaStreamTrack.clone().

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -568,16 +568,9 @@
             </li>
 
             <li>
-              <p>Let <var>clonedTracks</var> be a list that contains the result
-              of running <code><a href=
-              "#dom-mediastreamtrack-clone">MediaStreamTrack.clone()</a></code>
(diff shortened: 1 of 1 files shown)


fluffy:
LGTM

getusermedia.html
```diff
@@ -505,26 +526,9 @@
...
-              "#track-set">track set</a>, then abort these steps.</p>
-            </li>
-
-            <li>
-              <p>Add <var>track</var> to <var>stream</var>'s <a href=
-              "#track-set">track set</a>.</p>
-            </li>
-          </ol>
+          invoked, the User Agent MUST <a href="#stream-add-track">add the
+          track</a>, specified by the methods first argument, to this
```
alvestrand: nit: don't think there should be commas here.

Issue #159, opened 2015-04-15
MediaStream's active attribute


annevk:
The link from the constructor to "active" indicates a `MediaStream` state of active/inactive. However, the prose doesn't reference to this as a state but rather as something that is a boolean. This is rather confusing.

jan-ivar:
I gather the prose meant to link to the [#widl-MediaStream-onactive](http://w3c.github.io/mediacapture-main/getusermedia.html#widl-MediaStream-onactive) attribute rather than the state of the same name.

annevk:
I see. That's the same problem as #158. It's best to refer to the internal slots as public members can start throwing and such.

adam-be:
OK. Recap for my own understanding.
Depending on if we refer to the active attribute or the active/inactive concept, we should link consistently.
active attribute: #widl-MediaStream-active
active/inactive concept: #stream-inactive.

adam-be:
Proposed fix: #168

alvestrand:
Closed after merging #168.

PR #168, opened 2015-04-24
Distinguish between links to the active attribute and the active/inactive concept (Issue #159)

adam-be:


fluffy:
LGTM

adam-be:
Modifying the public API attribute here is might not be what we want.

Actually, I don't think we need to say anything about the "active" state when we construct the MediaStream since the state is indirectly defined by the stream's tracks.

"
A MediaStream object is said to be active when it has at least one MediaStreamTrack that has not ended. A MediaStream that does not have any tracks or only has tracks that are ended is inactive.
"

alvestrand:
What's the proper idiom for describing "this looks like an attribute, but is actually the result of a computation over the underlying state of the object's constituent parts"?

annevk:
If active/inactive are the result of something else, you don't need to modify the attribute. IDL attributes are getter/setter pairs when you get to JavaScript. I've started defining them as such, e.g.:

"The active attribute's getter must return true if the context object's is active, and false otherwise."

Given that active/inactive are the result of a computation, you don't need to explicitly set either those or the active attribute in the constructor as @adam-be observes.

adam-be:
Updated to reflect recent discussion.

Issue #162, opened 2015-04-15
MediaStreamError should not be an interface


annevk:
Errors in the platform are represented by JavaScript `Error` object subclasses.

annevk:
@domenic

domenic:
There's a note:

> We use MediaStreamError rather than deriving from Error until the situation with adding additional information into an error has been clarified. 

I do think this needs to be resolved before the spec advances though. If nothing else, they can define their own Error subclass using full-on ES spec terminology, similarly to what Web IDL does (but probably less crazily since Web IDL is caught in a mix of ES5 and ES6 worlds).

jan-ivar:
@domenic Thanks, I'm not an editor but I'm sure any pointers to examples of how that should or might look like would be appreciated.

domenic:
@jan-ivar https://people.mozilla.org/~jorendorff/es6-draft.html#sec-nativeerror-object-structure is basically what you want to do. It's templatized on _NativeError_ so just use `MediaStreamError` instead and add a few steps to the constructor specifying how to set the extra properties. (Probably don't add to the .prototype like .message and .name are, those seem like largely legacy...)

You'll probably also want something at the top of such a section saying something like "assume that the rules for [ECMAScript standard built-in objects](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-ecmascript-standard-built-in-objects) are in effect in this section" so that it's clear that e.g. the property descriptor defaults are the same as in ES, `.length` and `.name` are inferred the same way, etc.

adam-be:
@domenic, are you saying that we shouldn't use WebIDL to describe MediaStreamError, but rather do it in prose?

domenic:
Yes, WebIDL does not have the capability to create new Error subclasses.

alvestrand:
Is it kosher to declare a WebIDL object and use prose to say "we are declaring an Error subclass that has the same fields as this WebIDL object"?

Or is it best to avoid WebIDL altogether?

domenic:
Yeah, that doesn't really work, as WebIDL interfaces have entirely different semantics than ES error types do. You might as well use a bulleted list and say "we are declaring an Error subclass that has the same properties as this bulleted list". (But you would of course have to add details on e.g. what the constructor signature is, how/when the properties are defined, etc. like the ES spec does, so that would just be an informative note.)

jan-ivar:
Is it too late to to reconsider putting the failing constraint name in the message...?

[...]

PR #170, opened 2015-05-06
rough draft of new PermissionDeniedError object

burnburn:
This is not a complete pull request intended to go into the spec.  It is a rough draft of how the new error definitions might look, based on the recommendations given by @domenic in issue #162 .
It is lacking many links and proper style information, and it only shows one error definition, one that does not add additional information fields.  However, if this general approach looks correct, without any major missing content, then I will do the same for the rest of the errors currently lumped together as "MediaStreamError".

Please comment.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2028,45 +2028,142 @@
   <section>
     <h2>Error Handling</h2>
 
-    <p>All promises in this specification, when they are rejected, are rejected
-    with an object that implements the <code><a>MediaStreamError</a></code>
-    interface.</p>
+    <p>This section and its subsections extend the list of Error
+    subclasses defined in [ES6] following the pattern for NativeError
+    in [ES6, sec 19.5.6].  Assume the following:</p>
+
+    <ul>
+      <li>that use of syntax such as [[something]] and %something% is
+      as used in [ES6].</li>
+      <li>that the rules for ECMAScript standard built-in objects
+      [ES6, sec 17] are in effect in this section.</li>
+      <li>that the new intrinsic objects %PermissionDeniedError% and
+      %PermissionDeniedErrorPrototype% are available as if they had
+      been included in [ES6, Table 7] and all referencing sections,
+      e.g. [ES6, sec 8.2.2], thus behave appropriately.</li>
+    </ul>
 
     <section>
-      <h3>MediaStreamError</h3>
+      <h2>PermissionDeniedError Object</h2>
 
-      <p>All errors defined in this specification implement the following
-      interface:</p>
+      <section>
+        <h2>PermissionDeniedError Constructor</h2>
 
-      <dl class="idl" title="[NoInterfaceObject] interface MediaStreamError">
-        <dt>readonly attribute DOMString name</dt>
+        <p>The PermissionDeniedError Constructor is the
+        %PermissionDeniedError% intrinsic object.  When the
+        PermissionDeniedError constructor is called as a function
+        rather than as a constructor, it creates and initializes a new
+        PermissionDeniedError object. A call of the object as a
+        function is equivalent to calling it as a constructor with the
+        same arguments. Thus the function call
+        PermissionDeniedError(...) is equivalent to the object
+        creation expression new PermissionDeniedError(...) with the
+        same arguments.</p>
 
-        <dd>
-          <p>The name of the error</p>
-        </dd>
+        <p>The PermissionDeniedError constructor is designed to be
+        subclassable. It may be used as the value of an extends clause
+        of a class definition. Subclass constructors that intend to
+        inherit the specified PermissionDeniedError behaviour must
+        include a super call to the PermissionDeniedError constructor
+        to create and initialize subclass instances with an
+        [[ErrorData]] internal slot.</p>
 
-        <dt>readonly attribute DOMString? message</dt>
+        <section>
+          <h2>PermissionDeniedError ( message )</h2>
 
-        <dd>
-          <p>A User Agent-dependent string offering extra human-readable
-          information about the error.</p>
-        </dd>
+          <p>When a PermissionDeniedError function is called with
+          argument message the following steps are taken:</p>
+          <ol>
+            <li>If NewTarget is undefined, let newTarget be the active
+            function object, else let newTarget be NewTarget.</li>
+            <li>Let O be OrdinaryCreateFromConstructor(newTarget,
+            "%PermissionDeniedErrorPrototype%", «[[ErrorData]]»
+            ).</li>
+            <li>ReturnIfAbrupt(O).</li>
+            <li>If message is not undefined, then</li>
+            <ol>
+              <li>Let msg be ToString(message).</li>
+              <li>Let msgDesc be the PropertyDescriptor{[[Value]]:
+              msg, [[Writable]]: true, [[Enumerable]]: false,
+              [[Configurable]]: true}.</li>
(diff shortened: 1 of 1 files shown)


domenic:
This is looking good! Although, I am not sure you need to introduce new error types for all your different cases; what is the full plan exactly? A single `MediaStreamError` with a few extra identifying fields should probably suffice, right?

burnburn:
That was actually my first thought, but based on how EvalError,
ReferenceError, etc. were defined it seems that the name attribute is
supposed to only have the value of the error type, e.g. "ReferenceError"
for a ReferenceError, and so on, with a completely different error defined
for each semantically unique failure.  This suggests that there really is
no such thing as a MediaStreamError with sub fields to distinguish among
PermissionDeniedError, ConstraintNotSatisfiedError, OverconstrainedError,
and SourceUnavailableError, each of which are semantically different.  For
example, ConstraintNotSatisfiedError needs to be able to return a string
containing the name of the failing constraint, while PermissionDeniedError
does not.
  Thus, it appears that the best way to follow the ES6 pattern is to define
a separate error for each.  This would be easier if in ES6 the error name,
type, and class were not essentially the same thing.

domenic:
I don't think that's quite accurate. ES errors are quite broad in general. E.g., look at the number of places in the spec TypeError, RangeError, and ReferenceError are used. You shouldn't pollute the global namespace with 4 types of media error. Rather, you should have the minimum necessary given your extra constraints. That's why I ask for an idea of the full plan.

burnburn:
Okay, then there are two approaches that I think we could take:  the right way, and the simple/quick way.

The right way is to create two new errors:  MediaStreamError (which would cover both PermissionDeniedError and SourceUnavailableError, the two that are truly errors that are specific to acquiring access to a MediaStream) and ConstraintError (which would cover both ConstraintNotSatisfiedError and OverconstrainedError, the two that are really about constraint satisfaction rather than anything to do specifically with MediaStreams).

The simple way is to just call all of them a MediaStreamError since no one else is currently using constraints and figure that the constraint error pieces can be pulled out later if/when anyone else starts using constraints.
One potentially complicating issue is that some of these may occur as events and not just Promise rejections.  For example, OverconstrainedError (as an event) can definitely happen multiple times, and be fixed, without causing a Promise rejection.

I have not vetted this on the list yet -- I'm just trying to get a rough sense from you of what might be reasonable and to make sure that I didn't miss any major specification pieces.

alvestrand:
@domenic , we have a need to give back errors that allow Javascript to take different decisions based on something predictable (that is - *not message*).
So far, we've used the name field for distinguishing the error. If Ecmascript says you have to pollute the namespace when having different values in the name field .... I don't see how we can avoid polluting the namespace.

We're already pretty close to the minimum set we need - either we have to have those names, or we'll have to find some other way to make distinguishable errors that Javascript can use.

What's your advice?

domenic:
> @domenic , we have a need to give back errors that allow Javascript to take different decisions based on something predictable

Profuse apologies since I am probably retreading old ground, but can you explain why this is? (I.e. the user-facing use case---different UI that will be shown, for example.) In contrast, JS does not include a way to distinguish between the TypeError caused by trying to call a non-function and the TypeError caused by trying to access a property of `null`. DOM does not allow distinction between the InvalidStateError DOMException caused by dispatching an event twice and that caused by using surroundContents() incorrectly. Etc.

If there are such use cases, then perhaps separate errors are the way to go; the alternative would be introducing a property like `type` or `cause` that takes a string enum, but that might be a bit too ad-hoc. (At one time I outlined [vague hopes](https://esdiscuss.org/topic/error-objects-in-w3c-apis#content-4) for something more comprehensive, but I wouldn't want to make you guys the guinea pigs for such a scheme.)

But it's unclear to me why MSE is special such that unlike other web specs there's a need to distinguish.

alvestrand:
On 05/21/2015 05:17 PM, Domenic Denicola wrote:
>
>     @domenic <https://github.com/domenic> , we have a need to give
>     back errors that allow Javascript to take different decisions
>     based on something predictable
>
> Profuse apologies since I am probably retreading old ground, but can
> you explain why this is? (I.e. the user-facing use case---different UI
> that will be shown, for example.)
>

The typical example (and one that consumes one of our errors) is the
difference between "no such device" and "permission denied".
In this case, the JS need is to say "please plug in a camera" versus
"OK, if you don't want to make a call, let's not do it".

> In contrast, JS does not include a way to distinguish between the
> TypeError caused by trying to call a non-function and the TypeError
> caused by trying to access a property of |null|. DOM does not allow
> distinction between the InvalidStateError DOMException caused by
> dispatching an event twice and that caused by using surroundContents()
> incorrectly. Etc.
>
In the case of a TypeError, one can (theoretically) inspect the object
and figure out what caused the error. In the case of a getUserMedia
call, all that is returned in an error case is the error; the state that
caused the error is not made available for inspection.

> If there are such use cases, then perhaps separate errors are the way
> to go; the alternative would be introducing a property like |type| or
> |cause| that takes a string enum, but that might be a bit too ad-hoc.
> (At one time I outlined vague hopes
> <https://esdiscuss.org/topic/error-objects-in-w3c-apis#content-4> for
> something more comprehensive, but I wouldn't want to make you guys the
> guinea pigs for such a scheme.)
>
> But it's unclear to me why MSE is special such that unlike other web
> specs there's a need to distinguish.
>

I think the cases in the document where new error codes are used, the
state is relatively clear.

We *also* have the issue of returning additional information (such as
the constraint that was unsatisfiable in the constraint error case),
which the WG has idnicated that it strongly wants.
Assuring that the extra information is available has traditionally been
the role of the type system, which means that extra types seems like the
way to go, no matter how the other issue resolves.

domenic:
Thanks for explaining. At first I thought these custom errors were just to satisfy the

> We *also* have the issue of returning additional information (such as the constraint that was unsatisfiable in the constraint error case),

use case, and did not know about the other case.

With that in mind, I guess new types is probably the best way to go. The alternative would be to follow the path of DOMException and have a `.name` that does not align with the constructor name, but it is probably better to follow ES here.

With that said, I would still urge caution and consolidation where possible. For example:

- OverconstrainedError and ConstraintNotSatisfiedError seem like they could be merged. (I at least can't imagine when distinguishing is going to result in different behavior.)
- AbortError, NotSupportedError, and NotFoundError could just use DOMException with the appropriate name; 
-  PermissionDeniedError could probably be a DOMException with name SecurityError
- SourceUnavailableError could probably be a DOMException with name NotReadableError

I might be wrong on specifics here, where you actually do need to e.g. distinguish a SourceUnavailableError from a NotReadableError, but hopefully you can see the general direction I am driving at. You clearly need at least one new type, to support the additional info about the violated constraint. But after that, being conservative would make more sense.

jan-ivar:
> OverconstrainedError and ConstraintNotSatisfiedError seem like they could be merged.

+1 https://lists.w3.org/Archives/Public/public-media-capture/2014Oct/0212.html

> AbortError, NotSupportedError, and NotFoundError could just use DOMException with the appropriate name; 

+1

> PermissionDeniedError could probably be a DOMException with name SecurityError
> SourceUnavailableError could probably be a DOMException with name NotReadableError

sgtm, but that would probably require list discussion.

jan-ivar:
In hindsight I think like the OverconstrainedError name better fwiw.

alvestrand:
Closing this PR - a new set of PRs has been created.

PR #171, opened 2015-05-20
Referring to html5.1 for describing srcObject in media elements.

stefhak:
per issue #161, intended to replace PR #165

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1655,249 +1655,213 @@
     <code>MediaStream</code> is playing. The timeline does not increment when
     the playout of the <code>MediaStream</code> is paused.</p>
 
+    
+
     <p>User Agents that support this specification MUST support
-    the <code><a href="http://www.w3.org/html/wg/drafts/html/master/single-page.html#dom-media-srcobject">srcObject</a></code>
+    the <code><a href="http://www.w3.org/TR/2015/WD-html51-20150506/semantics.html#dom-media-srcobject">srcObject</a></code>
     attribute of the <code>HTMLMediaElement</code> interface defined
     in [[!HTML51]], which includes support for playing <code>MediaStream</code>
-    objects. Any <code><a href="http://www.w3.org/html/wg/drafts/html/master/single-page.html#audiotrack">AudioTrack</a></code>
-    or  <code><a href="http://www.w3.org/html/wg/drafts/html/master/single-page.html#videotrack">VideoTrack</a></code> created in the <code><a href="http://www.w3.org/html/wg/drafts/html/master/single-page.html#media-data-processing-steps-list">media-data-processing-steps-list</a></code>
-    defined in [[!HTML51]] must be initialized as follows:</p>
+    objects.</p>
+
+    <p>The [[!HTML51]] document outlines how the <code>HTMLMediaElement</code>
+    works with <code><a>MediaStream</a></code>'s. For clarity: </p>
 
     <ul>
-      <li>
-        <p><code>AudioTrack.id</code> and <code>VideoTrack.id</code> to
-        the value of the corresponding <code><a href=
-        "#dom-mediastreamtrack-id">MediaStreamTrack.id</a></code>
-        attribute</p>
-      </li>
+      <li><p>Whenever an [[!HTML51]] <code><a href="http://www.w3.org/TR/2015/WD-html51-20150506/semantics.html#audiotrack">AudioTrack</a></code> or a <code><a href="http://www.w3.org/TR/2015/WD-html51-20150506/semantics.html#videotrack">VideoTrack</a></code> is created, the <code>id</code> and <code>label</code>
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -1626,188 +1626,55 @@
...
-
-      <ul>
-        <li>
-          <p>Whenever the User Agent runs the <a href=
-          "http://www.w3.org/TR/html5/embedded-content-0.html#media-element-load-algorithm">
-          media element load algorithm</a>, if <code>srcObject</code> is
-          specified, the User Agent must immediately go to the <a href=
-          "http://www.w3.org/TR/html5/embedded-content-0.html#concept-media-load-resource">
-          resource fetch phase</a> of the algorithm.</p>
-        </li>
```
burnburn: Is the media element load algorithm modification above addressed in the new reference?  Should it be?  In HTML5.1 there seem to be several steps between the load algorithm and the fetch phase.

burnburn:
I made a comment at one place in the text that really applies more generally.  There are a number of exceptions to the load algorithm that the original text proposed, as well as a number of error scenarios and notifications, that are not specifically called out in the replacement text.  Is that because they are all fully addressed in the HTML 5.1 text (not clear to me from a quick review), because they are no longer relevant, or for some other reason?

getusermedia.html
stefhak: When I read HTML5.1 I come to the conclusion that this is addressed: "media element load algorithm" + "resource selection algorithm" (with mode==object) + "resource fetch algorithm" (object==local).

But there are many words in there :)

stefhak:
I read the html5.1 as this was already addressed. The case when the source is a media provider object (includes MediaStream) is covered there: http://www.w3.org/TR/2015/WD-html51-20150417/semantics.html#media-provider-object

burnburn:
I agree that there is a media provider object definition (and in fact I read the 2-3 sentences for it).
The question comes down to the behavior in the media load algorithm http://www.w3.org/TR/html5/embedded-content-0.html#media-element-load-algorithm

There are many steps there that officially apply now that did not in the prior text, so I want to make sure that for HTML 5.1 implementers they either see no changes or see appropriate changes that do not, for example, introduce unnecessary delays or surprising error conditions into the current behavior.

alvestrand:
Adam will review the spec to figure out if we caught all the pieces.

stefhak:
I'm working on this still. I will update the PR.

stefhak:
I've now updated the PR to the best of my knowledge. Some comments:
* readyState and networkState is pretty well defined in html5.1, so I've removed those from the table. However, this will mean a change in behavior since during the load/resFetch/resFetch process networkState will toggle (idle, noSource, netwLoading, idle) and not constantly stay at idle. Also, for networkState the state "HaveMetadate" will be gone through. List discussion on this part?
* there is no text about live track lists and so on, I find that pretty clear in html51
* also no text about what happens if the MediaStream is inactive when assigned as srcObject; I find that too being clear in html51
* I have added text saying that if a MediaStream goes from active to inactive the ended event should fire from the MediaElement and so on. Questionable if this really adds anything of value though - the app interested could just as well listen to MediaStream events.

An open question is if we should ask the html people to add clarifications related to buffered, playoutRate etc. to cover "live" media (so we could remove those parts).

Assigning to Adam for a second round of reviews.

alvestrand:
On 06/23/2015 02:45 PM, stefan hakansson wrote:
>
> I've now updated the PR to the best of my knowledge. Some comments:
>
>   * readyState and networkState is pretty well defined in html5.1, so
>     I've removed those from the table. However, this will mean a
>     change in behavior since during the load/resFetch/resFetch process
>     networkState will toggle (idle, noSource, netwLoading, idle) and
>     not constantly stay at idle. Also, for networkState the state
>     "HaveMetadate" will be gone through. List discussion on this part?
>
I think this is not a problem. The difference will be apparent only to 
people who watch these transitions.

>   * there is no text about live track lists and so on, I find that
>     pretty clear in html51
>   * also no text about what happens if the MediaStream is inactive
>     when assigned as srcObject; I find that too being clear in html51
>   * I have added text saying that if a MediaStream goes from active to
>     inactive the ended event should fire from the MediaElement and so
>     on. Questionable if this really adds anything of value though -
>     the app interested could just as well listen to MediaStream events.
>
I think this has value if (for example) one is using a "player" library 
that creates controls and so on for the media element, and which doesn't 
know anything about mediastreams. Not sure how important this is.
>
> An open question is if we should ask the html people to add 
> clarifications related to buffered, playoutRate etc. to cover "live" 
> media (so we could remove those parts).
>

I am fine with not bothering them :-)

adam-be:
I don't think it's super important that we keep the exact behaviour we had before this patch for two reasons. 1) In may not have been the correct behaviour. 2) It's more important to align with other types of srcObjects, so that implementers don't have to add more exceptions for MediaStream than necessary.

Issue #163, opened 2015-04-15
Please use [Exposed]


annevk:
That way it is much clearer what is exposed to Window and/or Worker.

adam-be:
I don't see reason not to.

alvestrand:
What is the default exposure if nothing's specified?
It seems we need a decision on what's exposed to Worker; I've been assuming that everything's exposed to Window.

alvestrand:
Note - [Exposed] is not defined in http://www.w3.org/TR/WebIDL/ but is defined in http://heycam.github.io/webidl/#Exposed - so we need to make sure our references point to the right place. (It's not the only new WebIDL feature we depend on, so we should reference the new spec anyway).

annevk:
I would prefer that even the default is explicitly specified. That way it is more clear that you meant the default rather than forgot to mention this.

alvestrand:
Conclusion of list (non)discussion was that we do [Exposed=Window].

alvestrand:
Resolved in #172

PR #172, opened 2015-05-21
Issue #163: Use [Exposed]

adam-be:


adam-be:
Can you take a quick look at this resolution @annevk?

annevk:
This looks good.

alvestrand:
The Travis error looks like the IETF cert problem. This should be mergeable.

adam-be:
Yes, I read it as "unable to merge", but that was not the case.

PR #175, opened 2015-05-28
LC-3025: Add note about addtrack and removetrack not being used by th…

adam-be:
…is spec


adam-be:
@burnburn, it worked. PR on a PR :)

PR #177, opened 2015-05-28
Add latency constraint.

pthatcherg:


getusermedia.html
```diff
@@ -4491,6 +4507,14 @@ <h2 id="sec-track-properties">Track Constrainable Property
             that no audio artifacts are introduced. This allows applications to
             control this behavior.</td>
           </tr>
+
+          <tr id="def-constraint-latency">
+            <td>latency</td>
+
+            <td><code><a>ConstrainDouble</a></code></td>
+
+            <td>The latency or latency range, in seconds.</td>
```
alvestrand: A proposed "more text":

The latency is the time between start of processing (for instance, when sound occurs in the real world) to the data being available to the next step in the process. Low latency is critical for some applications; high latency may be acceptable for other applications because it helps with power constraints. The number is expected to be the target latency of the configuration; the actual latency may show some variation from that.

stefhak:
I think your "more text" makes sense Harald.
One question: should the latency be in whole seconds (not fractions), with 0 meaning shortest latency possible? It seems the use case is a long latency (30 seconds or so).

alvestrand:
@pthatcherg, I assume you'll update.

pthatcherg:
I have updated.

pthatcherg:
I think it's important to be able to provide values less than one second,
but more than 0.  Plus, the consensus on the list as floating point
seconds.

PR #181, opened 2015-06-08
Remove text about firing event handlers (from event handler attribute descriptions)

adam-be:
We should not have any text in the event handler attribute descriptions saying that this event handler is fired when this or that happens. The event handler attributes are fired as part of the event flow procedure and that's redundant to describe for every single attribute handler.

This PR simplifies the event handler attribute descriptions to only mention the corresponding event type (that links to the event summary table).

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3354,28 +3354,6 @@ if(!supports["width"] || !supports["height"]) {
       "#media-stream-track-interface-definition">MediaStreamTrack Interface
       Definition</a> for an example of this.</p>
 
-      <p>When the User Agent is no longer able to satisfy the
-      required constraints</var> from the currently valid Constraints, the User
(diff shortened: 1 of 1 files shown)


adam-be:
@burnburn, can you review the change regarding the "overconstrained" event. That edit required moving quite a lot of text out of the event handler description.

burnburn:
I agree that an event *handler* description should not explain the *event*, but we still need to explain the event itself.  I dislike where you've moved the overconstrained text, but I think it may be fixable after the next PR I send, since I need to completely revamp the event section listing anyway because we will be removing MediaStreamError and MediaStreamErrorEvent.

Let's revisit this one after I send in the MediaStreamError(Event) rewrite.

alvestrand:
Waiting for #194 and the followup PR - easier to place this right when the error restructuring is complete.

stefhak:
Adam will remove the overconstrained part and merge.

Issue #174, opened 2015-05-28
getUserMedia algorithm should be made more synchronous 


youennf:
As per http://w3c.github.io/mediacapture-main/archives/20150324/getusermedia.html#dom-mediadevices-getusermedia, step3 of the getUserMedia implementation is done asynchronously.

But steps 3.1 to 3.5 could be run synchronously.
In particular, the implementation should be able to return already rejected promises, since JS callbacks will be called asynchronously anyway.

Step 3.6 should also be started synchronouly, although actual prompting and user decision will usually happen asynchronously.

Would it be possible to rewrite the description to take that into account?

jan-ivar:
To what end?

In practice there are inherent reasons to get a list of media devices in parallel as to not take up time on the main JS thread (potentially querying devices). So this matches at least one implementation.

3.1. is just a declaration for 3.2. so this seems to follow general programming best practice of defining variables as close to their use as possible. 

3.6. avoids asking users for devices they don't have (3.2.) Some implementations also let the user select devices in their permission prompt (firefox).

3.5. Seems related to 3.6. and as both optionally jump to 3.11. Moving 3.5. to the synchronous part would entail duplicating 3.11.

youennf:
As I read it, the spec rules out implementation that could reject/resolve the promise synchronously.
That is unfortunate.

In WebKit, step 3.2 is about to be implemented synchronously.
From the JS end-user prospective, this does not matter, the promise will call the reject callback asynchronously anyway.

If we were to follow closely the spec (or my interpretation of it), we would:
1. schedule an asynchronous call in which we would:
1.a check that the dictionary is empty.
1.b. reject the promise.
2. schedule another asynchronous call to actually call the promise reject callback.
This seems like unnecessary overhead.

The same may be applied to step 3.5. If the browser knows already that the getUserMedia call will fail, why schedule an asynchronous call to reject the promise?

The spec should allow that behavior.

jan-ivar:
Is it important to optimize failure?

As for ruling out implementations, [the spec says](http://w3c.github.io/mediacapture-main/getusermedia.html#conformance):
> Conformance requirements phrased as algorithms or specific steps may be implemented in any manner, so long as the end result is equivalent.

youennf:
I am not sure what exactly means "so long as the end result is equivalent".

The order in which resolved/rejected promise callbacks are called is guaranteed to be unambiguous.

[...]

PR #182, opened 2015-06-09
[Issue  #174] Move sanity checks to sync section of getUserMedia algorithm and tidy up

adam-be:
See individual commit messages for details.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2959,11 +2959,8 @@
                 </li>
 
                 <li>
-                  <p>Resolve <var>p</var> with <var>stream</var>.</p>
-                </li>
-
-                <li>
-                  <p>Abort these steps.</p>
+                  <p>Resolve <var>p</var> with <var>stream</var> and abort
+                  these steps.</p>
                 </li>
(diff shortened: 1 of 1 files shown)


adam-be:
Only first commit changes behaviour; the three others are editorial.

getusermedia.html
```diff
@@ -2807,24 +2807,23 @@
...
-                  <code>NotSupportedError</code> and jump to the step labeled
-                  <em>Error Task</em> below.</p>
-                </li>
+            <li>
+              <p>Run the following steps asynchronously:</p>
 
+              <ol>
                 <li>
                   <p>Let <var>finalSet</var> be an (initially) empty set.</p>
                 </li><!-- we could throw an exception instead (that's
```
alvestrand: can you remove the comment about throwing exception here? Promise-returning functions aren't allowed to throw exceptions, so this is OBE.

adam-be:
@alvestrand, I removed the entire comment including the old spec text. It's easy to dig up if we ever need it.

PR #185, opened 2015-06-10
Change NotSupportedError from a MediaStreamError to a DOMException

burnburn:
Partially fulfills #162.  See also #170.


stefhak:
LGTM.

PR #188, opened 2015-06-10
Change SourceAvailableError to a NotReadableError DOMException

burnburn:
Partially fulfills #162. See also #170.


stefhak:
LGTM.

PR #194, opened 2015-06-18
Overconstrainederror

burnburn:
This is the new OverconstrainedError in partial fulfillment of Issue #162 .  See closed PR #170 for discussion.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2041,47 +2041,47 @@
       as used in [ES6].</li>
       <li>that the rules for ECMAScript standard built-in objects
       [ES6, sec 17] are in effect in this section.</li>
-      <li>that the new intrinsic objects %PermissionDeniedError% and
-      %PermissionDeniedErrorPrototype% are available as if they had
+      <li>that the new intrinsic objects %OverconstrainedError% and
+      %OverconstrainedErrorPrototype% are available as if they had
       been included in [ES6, Table 7] and all referencing sections,
       e.g. [ES6, sec 8.2.2], thus behave appropriately.</li>
     </ul>
 
     <section>
-      <h2>PermissionDeniedError Object</h2>
+      <h2>OverconstrainedError Object</h2>
 
       <section>
-        <h2>PermissionDeniedError Constructor</h2>
+        <h2>OverconstrainedError Constructor</h2>
 
-        <p>The PermissionDeniedError Constructor is the
-        %PermissionDeniedError% intrinsic object.  When the
-        PermissionDeniedError constructor is called as a function
+        <p>The OverconstrainedError Constructor is the
+        %OverconstrainedError% intrinsic object.  When the
+        OverconstrainedError constructor is called as a function
         rather than as a constructor, it creates and initializes a new
-        PermissionDeniedError object. A call of the object as a
+        OverconstrainedError object. A call of the object as a
         function is equivalent to calling it as a constructor with the
         same arguments. Thus the function call
-        PermissionDeniedError(...) is equivalent to the object
-        creation expression new PermissionDeniedError(...) with the
+        OverconstrainedError(...) is equivalent to the object
+        creation expression new OverconstrainedError(...) with the
         same arguments.</p>
 
-        <p>The PermissionDeniedError constructor is designed to be
+        <p>The OverconstrainedError constructor is designed to be
         subclassable. It may be used as the value of an extends clause
         of a class definition. Subclass constructors that intend to
-        inherit the specified PermissionDeniedError behaviour must
-        include a super call to the PermissionDeniedError constructor
+        inherit the specified OverconstrainedError behaviour must
+        include a super call to the OverconstrainedError constructor
         to create and initialize subclass instances with an
         [[ErrorData]] internal slot.</p>
 
         <section>
-          <h2>PermissionDeniedError ( message )</h2>
+          <h2>OverconstrainedError ( message )</h2>
 
-          <p>When a PermissionDeniedError function is called with
+          <p>When a OverconstrainedError function is called with
           argument message the following steps are taken:</p>
           <ol>
             <li>If NewTarget is undefined, let newTarget be the active
             function object, else let newTarget be NewTarget.</li>
             <li>Let O be OrdinaryCreateFromConstructor(newTarget,
-            "%PermissionDeniedErrorPrototype%", «[[ErrorData]]»
+            "%OverconstrainedErrorPrototype%", «[[ErrorData]]»
             ).</li>
             <li>ReturnIfAbrupt(O).</li>
             <li>If message is not undefined, then</li>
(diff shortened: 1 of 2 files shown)


alvestrand:
@domenic can you take a look at this?

getusermedia.html
```diff
@@ -2032,45 +2032,272 @@
...
+          <ol>
+            <li>If NewTarget is <b>undefined</b>, let <i>newTarget</i>
+            be the active function object, else let <i>newTarget</i>
+            be NewTarget.</li>
+            <li>Let <i>O</i> be
+            OrdinaryCreateFromConstructor(<i>newTarget</i>,
+            <code>"%OverconstrainedErrorPrototype%"</code>, «[[\ErrorData]]»
+            ).</li>
+            <li>ReturnIfAbrupt(<i>O</i>).</li>
+            <li>Assert: Type(<i>constraint</i>) is String.</li>
```
domenic: This is not accurate. constraint is a parameter passed in by the user: it could be anything.

If you want to convert it into a string, or do any validation, you need steps for that manually, similarly to what is done for `message`.

getusermedia.html
```diff
@@ -2032,45 +2032,272 @@
...
+        <section>
+          <h2>OverconstrainedError.prototype.constructor</h2>
+
+          <p>The initial value of the constructor property of the
+          prototype for the OverconstrainedError constructor is the
+          intrinsic
+          object <a href="#error-handling">%OverconstrainedError%</a>.</p>
+        </section>
+
+        <section>
```
domenic: @allenwb, in your opinion should new Error subclasses define prototype properties like this? They seem kind of pointless. But on the other hand, consistency is nice.

getusermedia.html
```diff
@@ -2965,21 +3192,18 @@
...
-                  "#widl-MediaStreamError-constraintName">constraintName</a></code>
-                  attribute is set to the name of the constraint that caused
-                  the error.</p>
+                  <p><em>Constraint Failure</em>:
+                  Let <var>constraint</var> be the name of the
+                  constraint that caused the error, and
+                  let <var>error</var> be the result of creating a
+                  new <code><a>OverconstrainedError</a></code> by
+                  calling
+                  OverconstrainedError(<var>constraint</var>).</p>
```
domenic: Probably want to allow implementations to include a message here (and below).

getusermedia.js
```diff
@@ -87,6 +87,15 @@ var respecConfig = {
         }
       ]
     }
-  ]
+  ],
+   localBiblio: {
+     ES6: {
+       title:  "ECMAScript 2015",
+       href: "http://wiki.ecmascript.org/lib/exe/fetch.php?id=harmony%3Aspecification_drafts&cache=cache&media=harmony:ecma-262_6th_edition_final_draft_-04-14-15.pdf",
```
domenic: You can use http://www.ecma-international.org/ecma-262/6.0/

getusermedia.js
```diff
@@ -87,6 +87,15 @@ var respecConfig = {
...
+       authors: ["Allen Wirfs-Brock"],
+       status: "Draft Standard",
```
domenic: No longer draft

getusermedia.js
```diff
@@ -87,6 +87,15 @@ var respecConfig = {
...
+       publisher: "ECMA"
```
domenic: Capitalization: it's ECMAScript, but Ecma International.

domenic:
Overall looking good.

getusermedia.html
allenwb: On Jun 18, 2015, at 4:50 PM, Domenic Denicola wrote:

> In getusermedia.html:
> 
> > +        object. It is not an Error instance and does not have an
> > +        [[\ErrorData]] internal slot.</p>
> > +        <p>The value of the [[\Prototype]] internal slot of the
> > +        OverconstrainedError prototype object is the intrinsic object
> > +        %ErrorPrototype%.</p>
> > +
> > +        <section>
> > +          <h2>OverconstrainedError.prototype.constructor</h2>
> > +
> > +          <p>The initial value of the constructor property of the
> > +          prototype for the OverconstrainedError constructor is the
> > +          intrinsic
> > +          object <a href="#error-handling">%OverconstrainedError%</a>.</p>
> > +        </section>
> > +
> > +        <section>
> @allenwb, in your opinion should new Error subclasses define prototype properties like this? They seem kind of pointless. But on the other hand, consistency is nice.
> 
> —
> Reply to this email directly or view it on GitHub.
> 
I'm not a big fan of defining a lot of distinct error classes (although wanting to add instance properties is a good reason for doing so).

But if you are going to define error classes then you should follow the full class pattern including having a unique prototype object and  a constructor property on that prototype.

Does OverconstrainedError.prototype really directly inherit from %ErrorPrototype%.  Isn't there something like a DOMError class that WebIDL introduces that  platform API errors should inherit from?

And speaking of WebIDL, if platform APIs are going to be routinely defining next exception classes like this wouldn't it be better for their structure to be defined in the WebIDL spec so that each feature area spec, like this one, doesn't need to replicate the boiler plate.  That's essentially what the ES6 spec.  does by specify each built-in exception class to as conforming to the *NativeError* object structure pattern.

domenic: > But if you are going to define error classes then you should follow the full class pattern including having a unique prototype object and a constructor property on that prototype. 

Yes, that's what this is doing. The question is whether the prototype should have properties for all the other instance properties (constraint and message), even though that's kind of pointless.

> Does OverconstrainedError.prototype really directly inherit from %ErrorPrototype%. Isn't there something like a DOMError class that WebIDL introduces that platform API errors should inherit from?

It could inherit from DOMException, I suppose. No real win one way or another IMO.

> And speaking of WebIDL, if platform APIs are going to be routinely defining next exception classes like this wouldn't it be better

Yes, *if* that were true :). This should be a one-off (it's the first time an instance property has been needed).

getusermedia.html
burnburn: Actually, what I wanted was a value set in our browser algorithm.  I would be thrilled if users never touched this.  Is there a better way to accomplish that, or do I just need to make it match how message works?

getusermedia.html
burnburn: @allenwb, my only comment to this thread is that we would much rather just use WebIDL and extend to provide the constraint info we need, but there doesn't seem to be a way to do that.  Domenic suggested we follow the NativeError pattern, so that is what I did.  I copied everything as done for the various NativeError objects, including prototypes, prototype properties, and all.  We think it's beating a nail with a sledgehammer but are open to precise instructions on how to do it differently if what I did was incorrect.

getusermedia.html
burnburn: Good point.  I'll add that.

getusermedia.js
burnburn: I'd love to use that, but on that page it says "The PDF version is the definitive specification", so that encourages me to keep using a pointer to the PDF version.  @dontcallmedom , any thoughts here?

getusermedia.js
burnburn: Excellent.  I will update.

getusermedia.js
burnburn: Thanks.

burnburn:
Thanks for your review comments, @domenic .  I have addressed some of your points and asked followup questions to others above.

getusermedia.html
domenic: Yeah, just matching message is probably best, although you could throw for `undefined` instead of ignoring it.

getusermedia.html
allenwb: On Jun 18, 2015, at 5:21 PM, Domenic Denicola wrote:

> In getusermedia.html:
> 
> > +        object. It is not an Error instance and does not have an
> > +        [[\ErrorData]] internal slot.</p>
> > +        <p>The value of the [[\Prototype]] internal slot of the
> > +        OverconstrainedError prototype object is the intrinsic object
> > +        %ErrorPrototype%.</p>
> > +
> > +        <section>
> > +          <h2>OverconstrainedError.prototype.constructor</h2>
> > +
> > +          <p>The initial value of the constructor property of the
> > +          prototype for the OverconstrainedError constructor is the
> > +          intrinsic
> > +          object <a href="#error-handling">%OverconstrainedError%</a>.</p>
> > +        </section>
> > +
> > +        <section>
> But if you are going to define error classes then you should follow the full class pattern including having a unique prototype object and a constructor property on that prototype.
> 
> Yes, that's what this is doing. The question is whether the prototype should have properties for all the other instance properties (constraint and message), even though that's kind of pointless.
> 
Normally, data properties wouldn't be placed on a prototype.  However, Error does this with 'message' and 'name'.  There is some sense for doing it for 'name', while 'message' is kind of borderline. Fundamentally, it is providing default non-undefined values for those properties in case a subclass constructor doesn't add them to its instances (or prototype).  What do other DOM exceptions do?  That's what I'd try to be consistent with.


> Does OverconstrainedError.prototype really directly inherit from %ErrorPrototype%. Isn't there something like a DOMError class that WebIDL introduces that platform API errors should inherit from?
> 
> It could inherit from DOMException, I suppose. No real win one way or another IMO.
> 
Platform consistancy


> And speaking of WebIDL, if platform APIs are going to be routinely defining next exception classes like this wouldn't it be better
> 
> Yes, if that were true :). This should be a one-off (it's the first time an instance property has been needed).
> 
> 

There may also be an issue where WRT how WebIDL is used to describe object interfaces. Does 'constraint' show up in the OverConstrainedError interface description.  If so do the WebIDL rules require that all such properties exist on a prototype?

Allen

getusermedia.js
allenwb: The content of the the HTML and PDF are the same, although there are very minor formatting differences.  That note about the PDF being definitive is only there in case some one thinks  the two documents are saying different things. But, really they are the same and Ecma's intent is that the HTML version can be normatively referenced.

getusermedia.html
domenic: > What do other DOM exceptions do? That's what I'd try to be consistent with.

No other DOM exceptions have instance properties (except message, which they copy NativeError on).

> There may also be an issue where WRT how WebIDL is used to describe object interfaces.

WebIDL is not used at all to describe OverconstrainedError, since it does not have e.g. support for data properties or for subclassing ES built-ins.

alvestrand: Just a piece of history - we started off by subclassing DOMError back in 2013 or so. We've been completely consistent in what we need to have in the subclass ever since, but have been trying to navigate the labyrinth of actually getting a form of definition that people wouldn't object to on procedural grounds ever since.

We couldn't agree more that it would be simpler if WebIDL provided the mechanism we need - but since it doesn't, we try to make do.

stefhak:
@domenic and @allenwb: we plan to merge this next week (after releasing a new dated Editor's draft), and will then ask you to review.

PR #199, opened 2015-06-24
Change ended event to be a MediaStreamTrackEvent with new error property

burnburn:
The ended event on an MST is in many cases not an error, just an event.  To simplify the upcoming change of MediaStreamErrorEvent to be ConstrainedErrorEvent, and to better match the semantics of the ended event, this PR proposes to change the ended event to be a MediaStreamTrackEvent with a new, optional, error property that can be used when the track has ended due to an error.

This is in related to the fulfillment of Issue #162 . See closed PR #170 for discussion.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1415,9 +1415,11 @@
     <section>
       <h3>MediaStreamTrackEvent</h3>
 
-      <p>The <code><a href="#event-mediastream-addtrack">addtrack</a></code>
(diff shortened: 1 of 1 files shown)


adam-be:
I'm a bit confused by this change. Two events that are fired at different targets (MediaStream, MediaStreamTrack) gets merged into one. We never need the track for the 'ended' event since we have evt.target and the 'addtrack' and 'removetrack' events never use the error.

burnburn:
Replaced by PR #206.

PR #206, opened 2015-07-02
Create new EndedEvent

burnburn:
The ended event on an MST is in many cases not an error, just an event.  This PR proposes to change the ended event to be a new EndedEvent with an optional error property that can be used when the track has ended due to an error.

This is in related to the fulfillment of Issue #162 . See closed PR #170 for discussion.
This replaces PR #199 .

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1968,13 +1968,6 @@
           <p>Constructs a new <code><a>EndedEvent</a></code>.</p>
         </dd>
 
-        <dt>readonly attribute MediaStreamTrack track</dt>
-
-        <dd>
-          <p>The <code><a>MediaStreamTrack</a></code> that has
-          <a href="#track-ended">ended</a>.</p>
-        </dd>
-
         <dt>readonly attribute Error? error</dt>
 
         <dd>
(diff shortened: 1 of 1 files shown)


burnburn:
@adam-be , let me know if this addresses your concerns with PR #199.

adam-be:
Made a PR against your PR-branch to add the section about firing the event and how the bubbling and cancelable properties are set.

adam-be:
We could make this event interface more reusable by calling it something like ErrorEvent instead of the rather specific EndedEvent.

burnburn:
We could, but then it would be wrong.  Keep in mind that the goal was to have an event for Ended, which only in some cases is due to an error.  It looks weird that the only attribute is for an error, but the name of the event already tells us that it is for ended.  If we change the name we lose that.

getusermedia.html
```diff
@@ -1955,6 +1955,39 @@
...
+        eventInitDict)</dt>
+
+        <dd>
+          <p>Constructs a new <code><a>EndedEvent</a></code>.</p>
+        </dd>
+
+        <dt>readonly attribute Error? error</dt>
+
+        <dd>
+          <p>If the track <a href="#tarck-ended">ended</a> as a result of
```
stpeter: s/tarck-ended/track-ended/

getusermedia.html
```diff
@@ -1955,6 +1955,39 @@
...
+          an error, this attribute may be set to that error object.</p>
+        </dd>
+      </dl>
+
+      <dl class="idl"
+          title="dictionary EndedEventInit : EventInit">
+        <dt>Error? error = null</dt>
+
+        <dd>
+          <p>If the track <a href="#tarck-ended">ended</a> as a result of
```
stpeter: s/tarck-ended/track-ended/

burnburn:
Adam convinced me that a generic ErrorEvent would be better.  That change is proposed in a new PR #214, which replaces this one, so I'm closing this one.

Issue #201, opened 2015-06-25
Order of events ('addtrack' and 'active')


adam-be:
(The below text is sent to the list)

Consider a case with an inactive MediaStream. For some reason, a new 
'live' track is added to it by the user agent. This means that both an 
'addtrack' and 'active' event must be fired at the stream. I don't 
believe we have specified the order of these.

My proposal is that the 'addtrack and 'removestrack' events fire before 
'active' and 'inactive' in all such cases.

This means that the active attribute will be false when the 'addtrack' 
handler is fired, but it's better than the other way around. And in this 
particular case, it's pretty easy to foresee the upcoming activity 
change by looking at the readyState of the track.

stefhak:
Adam has sent mail to list: https://lists.w3.org/Archives/Public/public-media-capture/2015Jun/0040.html

adam-be:
PR for proposed solution (no discussion yet)
#207

stefhak:
Fixed via #207

PR #207, opened 2015-07-02
Specify order of add/removetrack and active/inactive events (issue #201)

adam-be:
PR for proposed solution


stefhak:
Awaiting list discussion.

stefhak:
List pinged https://lists.w3.org/Archives/Public/public-media-capture/2015Jul/0040.html

stefhak:
If there are no further comments on the list today I think we should merge this one.

PR #208, opened 2015-07-02
Add WebIDL definitions to Constrainable Pattern section to pass WebIDL validator

adam-be:


stefhak:
Let's wait until it actually passes the validator before merging.

adam-be:
The bug in the validator has been fixed, but I've not had a chance to re-run the check (since the machine, that I've temporarily hosted the HTML file on, is down)

stefhak:
I think we should merge this PR now.

PR #209, opened 2015-07-02
Once you implement a MediaTrackSupportedConstraints member, you support it

jan-ivar:
The [spec says](http://w3c.github.io/mediacapture-main/getusermedia.html#widl-MediaDevices-getSupportedConstraints-MediaTrackSupportedConstraints):

"A supported constrainable property MUST be represented by a member whose name is the constraint name and *whose value is true*. Any constrainable properties not supported by the User Agent MUST *not be present* in the returned dictionary."

In other words, members are always true. This optimizes the WebIDL to match.

I was originally against the breakdown of discrete dictionaries here, because of the code maintenance burden (more places to forget to update for each new constraint implemented), but I like that this at least can be managed entirely in the WebIDL definition file (and WebIDL binding code) without needing to define members *and* remembering to add manual code somewhere as well to set values to true.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1158,51 +1158,51 @@
         the [[!RTCWEB-CONSTRAINTS]] registry.</p>
 
         <dl class="idl" title="dictionary MediaTrackSupportedConstraints">
-          <dt>boolean width</dt>
+          <dt>boolean width = true</dt>
 
           <dd />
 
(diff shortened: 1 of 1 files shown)


stefhak:
As far as I can understand this makes sense.

stpeter:
@adam-be Perhaps you could take a look at this one?

PR #210, opened 2015-07-02
Add channelCount to audio track settings, constraints, capabilities

joeberkovitz:
LC proposal documented in https://lists.w3.org/Archives/Public/public-media-capture/2015Jun/0031.html to address lack of channel count information in audio track capabilities.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1198,6 +1198,10 @@
 
           <dd />
 
+          <dt>boolean channelCount</dt>
+
+          <dd />
+
           <dt>boolean deviceId</dt>
 
           <dd />
@@ -1258,6 +1262,10 @@
           <dd />
 
           <dt>(double or DoubleRange) latency</dt>
+
+          <dd />
+
+          <dt>(long or LongRange) channelCount</dt>
 
           <dd />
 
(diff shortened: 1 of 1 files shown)


stefhak:
This LGTM.

burnburn:
I haven't seen any list support for this yet.  Is there anyone who can give some insight as to who might implement this?

alvestrand:
Mozilla has said on the list that they will implement this.
Google has people arguing for this who are part of the WebAudio team.

getusermedia.html
```diff
@@ -1261,6 +1265,10 @@
 
           <dd />
 
+          <dt>(long or LongRange) channelCount</dt>
```
jan-ivar: Are we sure we want a range and not an array of numbers?

alvestrand: that's a list-discussion item. The Audio WG had earlier abandoned the idea of a list of configurations (2, 5.1, 7.2), and gone for a number instead. If it's a number, not an identifier, I think ranges are the correct treatment.

joeberkovitz: I would say that this should be treated the same as `sampleRate`, which is currently a range (even though it is likely that only discrete values within that range are acceptable values, much as for `channelCount`).

jan-ivar: OK just checking, no need for discussion on my account.

burnburn:
This LGTM.

PR #211, opened 2015-07-02
Allow device capabilities to be discoverable via enumerateDevices

joeberkovitz:
LC proposal documented in https://lists.w3.org/Archives/Public/public-media-capture/2015Jun/0034.html to allow applications to discover device capabilities.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2415,8 +2415,7 @@
           <code><a>MediaStream</a></code> returned by <code>getUserMedia({deviceId: <i>id</i>})</code>
           where <i>id</i> is the value of the <code>deviceId</code> attribute of this <code>MediaDeviceInfo</code>.</p>
 
-          <p>For output devices, this attribute will always be <code>null</code>. Future changes to this API will address
-          the acquisition of output device capabilities.</p>
+          <p>For output devices, this attribute has the value <code>null</code>.</p>
         </dd>
 
         <dt>readonly attribute DOMString groupId</dt>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2248,9 +2248,8 @@
           User Agent's available media input and output devices if enumeration
           is successful.</p>
 
-          <p>Elements of this sequence which represent input devices will be of type
-          <code><a>InputDeviceInfo</a></code> which extends <code><a>MediaDeviceInfo</a></code>
-          with input-specific elements.</p>
+          <p>Elements of this sequence that represent input devices will be of type
+          <code><a>InputDeviceInfo</a></code> which extends <code><a>MediaDeviceInfo</a></code>.</p>
 
           <p>Camera and microphone sources <em class="rfc2119">should</em> be
           enumerable. Specifications that add additional types of source will
@@ -2441,19 +2440,22 @@
       </dl>
     </section>
     <section>
-      <h2>Input-Specific Device Info</h2>
+      <h2>Input-specific Device Info</h2>
       <dl class="idl" title="interface InputDeviceInfo : MediaDeviceInfo">
         <dt>MediaTrackCapabilities getCapabilities()</dt>
 
         <dd>
-          <p>Returns a <code><a>MediaTrackCapabilities</a></code> object describing the device's audio or video track
+          <p>Returns a <code><a>MediaTrackCapabilities</a></code> object describing the primary
+          audio or video track of a device's <code><a>MediaStream</a></code>
           (according to its <code>kind</code> value), in the absence of any user-supplied constraints.
-          These capabilities will be identical to those that would have been obtained by calling
+          These capabilities MUST be identical to those that would have been obtained by calling
           <code>getCapabilities()</code> on the first <code><a>MediaStreamTrack</a></code> of this type in a
           <code><a>MediaStream</a></code> returned by <code>getUserMedia({deviceId: <i>id</i>})</code>
-          where <i>id</i> is the value of the <code>deviceId</code> attribute of this <code>MediaDeviceInfo</code>.</p>
-          <p>If no access has been granted to any local devices and this <code>InputDeviceInfo</code> has been filtered
-          with respect to unique identifying information (see above description of <code>enumerateDevices()</code> result), then this method returns <code>null</code>.
+          where <i>id</i> is the value of the <code>deviceId</code> attribute
+          of this <code>MediaDeviceInfo</code>.</p>
+          <p>If no access has been granted to any local devices and this <code>InputDeviceInfo</code>
+          has been filtered with respect to unique identifying information (see above description of <code>
+          enumerateDevices()</code> result), then this method returns <code>null</code>.
         </dd>
       </dl>
 
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2329,7 +2329,8 @@
                   a copy of <var>resultList</var>, and all its elements, where
                   the <code><a href=
                   "#widl-MediaDeviceInfo-label">label</a></code> member is the
-                  empty string.</p>
+                  empty string and the <code><a href=
+                  "#widl-MediaDeviceInfo-capabilities">capabilities</a></code> member is <code>null</code>.</p>
                 </li>
 
                 <li>
@@ -2360,7 +2361,7 @@
 
         <p>If access has been granted for a media device, the
         <code><a>MediaDeviceInfo</a></code> dictionary will contain the
-        deviceId, kind, label, and groupId.</p>
+        deviceId, kind, label, capabilities and groupId.</p>
       </section>
     </section>
 
@@ -2402,6 +2403,19 @@
           <p>A label describing this device (for example "External USB
           Webcam"). If the device has no associated label, then this attribute
           MUST return the empty string.</p>
+        </dd>
+
+        <dt>readonly attribute MediaTrackCapabilities capabilities</dt>
+
+        <dd>
+          <p>Returns a default set of capabilities describing the device's audio or video track
+          (according to its <code>kind</code> value), in the absence of any user-supplied constraints.
+          For input devices these capabilities will be identical to those that would have been obtained by calling
+          <code>getCapabilities()</code> on the first <code><a>MediaStreamTrack</a></code> of this type in a
+          <code><a>MediaStream</a></code> returned by <code>getUserMedia({deviceId: <i>id</i>})</code>
+          where <i>id</i> is the value of the <code>deviceId</code> attribute of this <code>MediaDeviceInfo</code>.
+          For output devices, the capabilities describe a hypothetical MediaStreamTrack whose characteristics
+          exactly match those of the device.</p>
         </dd>
 
         <dt>readonly attribute DOMString groupId</dt>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2333,8 +2333,7 @@
                   a copy of <var>resultList</var>, and all its elements, where
                   the <code><a href=
                   "#widl-MediaDeviceInfo-label">label</a></code> member is the
-                  empty string and the <code><a href=
-                  "#widl-MediaDeviceInfo-capabilities">capabilities</a></code> member is <code>null</code>.</p>
+                  empty string.</p>
                 </li>
 
                 <li>
@@ -2449,7 +2448,7 @@
         <dd>
           <p>Returns a <code><a>MediaTrackCapabilities</a></code> object describing the device's audio or video track
           (according to its <code>kind</code> value), in the absence of any user-supplied constraints.
-          For input devices these capabilities will be identical to those that would have been obtained by calling
+          These capabilities will be identical to those that would have been obtained by calling
           <code>getCapabilities()</code> on the first <code><a>MediaStreamTrack</a></code> of this type in a
           <code><a>MediaStream</a></code> returned by <code>getUserMedia({deviceId: <i>id</i>})</code>
           where <i>id</i> is the value of the <code>deviceId</code> attribute of this <code>MediaDeviceInfo</code>.</p>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2405,17 +2405,18 @@
           MUST return the empty string.</p>
         </dd>
 
-        <dt>readonly attribute MediaTrackCapabilities capabilities</dt>
+        <dt>readonly attribute Object capabilities</dt>
 
         <dd>
-          <p>Returns a default set of capabilities describing the device's audio or video track
+          <p>For input devices, contains a <code><a>MediaTrackCapabilities</a></code> object describing the device's audio or video track
           (according to its <code>kind</code> value), in the absence of any user-supplied constraints.
           For input devices these capabilities will be identical to those that would have been obtained by calling
           <code>getCapabilities()</code> on the first <code><a>MediaStreamTrack</a></code> of this type in a
           <code><a>MediaStream</a></code> returned by <code>getUserMedia({deviceId: <i>id</i>})</code>
-          where <i>id</i> is the value of the <code>deviceId</code> attribute of this <code>MediaDeviceInfo</code>.
-          For output devices, the capabilities describe a hypothetical MediaStreamTrack whose characteristics
-          exactly match those of the device.</p>
+          where <i>id</i> is the value of the <code>deviceId</code> attribute of this <code>MediaDeviceInfo</code>.</p>
+
+          <p>For output devices, this attribute will always be <code>null</code>. Future changes to this API will address
+          the acquisition of output device capabilities.</p>
         </dd>
 
         <dt>readonly attribute DOMString groupId</dt>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2248,6 +2248,10 @@
           User Agent's available media input and output devices if enumeration
           is successful.</p>
 
+          <p>Elements of this sequence which represent input devices will be of type
+          <code><a>InputDeviceInfo</a></code> which extends <code><a>MediaDeviceInfo</a></code>
+          with input-specific elements.</p>
+
           <p>Camera and microphone sources <em class="rfc2119">should</em> be
           enumerable. Specifications that add additional types of source will
           provide recommendations about whether the source type should be
@@ -2361,7 +2365,7 @@
 
         <p>If access has been granted for a media device, the
         <code><a>MediaDeviceInfo</a></code> dictionary will contain the
-        deviceId, kind, label, capabilities and groupId.</p>
+        deviceId, kind, label, and groupId.</p>
       </section>
     </section>
 
@@ -2405,19 +2409,6 @@
           MUST return the empty string.</p>
         </dd>
 
-        <dt>readonly attribute Object capabilities</dt>
-
-        <dd>
-          <p>For input devices, contains a <code><a>MediaTrackCapabilities</a></code> object describing the device's audio or video track
-          (according to its <code>kind</code> value), in the absence of any user-supplied constraints.
-          For input devices these capabilities will be identical to those that would have been obtained by calling
-          <code>getCapabilities()</code> on the first <code><a>MediaStreamTrack</a></code> of this type in a
-          <code><a>MediaStream</a></code> returned by <code>getUserMedia({deviceId: <i>id</i>})</code>
-          where <i>id</i> is the value of the <code>deviceId</code> attribute of this <code>MediaDeviceInfo</code>.</p>
-
-          <p>For output devices, this attribute has the value <code>null</code>.</p>
-        </dd>
-
         <dt>readonly attribute DOMString groupId</dt>
 
         <dd>
@@ -2429,7 +2420,6 @@
 
         <dt>serializer = { attribute }</dt>
       </dl>
-
       <dl class="idl" title="enum MediaDeviceKind">
         <dt>audioinput</dt>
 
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -2404,6 +2405,19 @@
...
+        <dt>readonly attribute MediaTrackCapabilities capabilities</dt>
+
+        <dd>
+          <p>Returns a default set of capabilities describing the device's audio or video track
+          (according to its <code>kind</code> value), in the absence of any user-supplied constraints.
+          For input devices these capabilities will be identical to those that would have been obtained by calling
+          <code>getCapabilities()</code> on the first <code><a>MediaStreamTrack</a></code> of this type in a
+          <code><a>MediaStream</a></code> returned by <code>getUserMedia({deviceId: <i>id</i>})</code>
+          where <i>id</i> is the value of the <code>deviceId</code> attribute of this <code>MediaDeviceInfo</code>.
+          For output devices, the capabilities describe a hypothetical MediaStreamTrack whose characteristics
```
stefhak: I am a bit unsure about the applicability to output devices. Did we ever discuss that?

joeberkovitz: I think we missed this in our earlier discussion. However, there is presently no output analogue to getUserMedia() that would hypothetically supply the capabilities of an output device. Without some kind of definition similar in spirit to the above, we seem to have no way to determine the capabilities of output devices until the Audio Output API is fleshed out.

stefhak:
I don't think we should add ways to determine capabilities of output devices to this spec. Let's focus on the input devices, at least for now.

joeberkovitz:
@stefhak Output devices are more central to the Web Audio API than input devices -- one can't make any sound in Web Audio without an output device.

I have a few questions: Why do you feel it's best to put this off? And when/where do you think we should address them, if not now? Would we change `enumerateDevices()` down the road or do you think this becomes a brand new API, only for output?

stefhak:
After re-reading http://w3c.github.io/mediacapture-main/ I think I was wrong in my initial reaction regarding output devices. enumerateDevices does indeed enumerate output audio devices, and adding capabilities to them does not seem like that big a deal in that perspective. 

I'm not sure about the language ("hypothetical MediaStreamTrack") in some parts. Additionally, there is text now about only devices that the user has approved use of will described. Are all output devices automatically in this category?

At the editor's meeting tomorrow I will ask one of the editors to review this PR.

joeberkovitz:
Thanks. A couple of quick responses:

The language of "hypothetical MediaStreamTrack matching the output device characteristics" was included because my understanding is that MediaStreamTracks are not creatable for output devices: they are defined in terms of a source only. Perhaps there is a better way to phrase this.

The device approval policy was intended to track the same permission policy for the rest of enumerateDevices() filterable information, which currently states that once the user has granted access to *any* device, this filterable information is obtainable for *all* devices.

I continue to think that some way of requesting permission from the user to obtain info on available devices would clarify this area, and someone proposed that on the list not long ago.

jan-ivar:
To me, this begs a lot of questions:

* `navigator.mediaDevices` has `getUserMedia()`, but no API yet for addressing outputs AFAIK, so what will these Capabilities be used for?

* **Permission:** Why must I share my mic (or camera) with a sound-producing site/game for it to access info about (not even actual control-access to) speakers? That doesn't make any sense to me. Also, if I don't have a camera or microphone, how can a site gain info about my speakers/output-recording devices?

* **Constraints:** *Capabilities* is one third of the *Constrainable* pattern (*Capabilities*/*Settings*/*Constraints*), yet there's no Constrainable API for output devices, so is this reuse non-constraints-related? Also, as @joeberkovitz points out, *MediaTrack*Capabilities, are source capabilities, not sink capabilities. Should we really reuse these for cross-purposes?

-
    dictionary MediaTrackCapabilities {
             (double or DoubleRange) volume;
             (long or LongRange)     sampleRate;
             (long or LongRange)     sampleSize;
             sequence<boolean>       echoCancellation;
             (double or DoubleRange) latency;
             (long or LongRange)     channelCount;
             DOMString               deviceId;
             DOMString               groupId;
    };

We're surely stretching definitions here.

burnburn:
I looked at this from an editorial/spec consistency standpoint and agree with Jan-Ivar that this crosses the line into more output info (and ultimately control) than the rest of the specification supports.  Because output issues can be complex and are definitely different in some cases from input issues, all output work (with the notable exception I'll get to below) is being done in the separate Audio Output specification.
The one exception is that we do enumerate output devices.  This exception only occurred because of the absolute failure of WebRTC applications where a user donned a headset and could not choose for the audio to come out of the headset when using the headset microphone.

I think going beyond this is too far and that it would be better for the Audio Output spec to add any additional info for output devices.

Then again, I personally think capabilities info in enumerateDevices is too far, but that's just my personal opinion (not as an editor).

jan-ivar:
@burnburn thanks for explaining the reason for the exception.

Having an Audio Output spec add this makes sense to me FWIW. It might be worth pondering whether such a spec would be likely to extend (with the `partial` webidl keyword) either `MediaTrackCapabilities` or `MediaDeviceInfo`. If the latter, we may want to rename `info.capabilities` to `info.inputCapabilities` now to anticipate this (assuming we're still going ahead). *

*) Because `(MediaTrackCapabilities or OutputCapabilities) capabilities` wont work, as two dictionaries are not [distinguishable](https://heycam.github.io/webidl/#dfn-distinguishable) from each other in webidl today.

alvestrand:
The logic of how we got here is kind of a step by step process.

- We want to list both input and output devices. Sometimes we even want to indicate groupings between them. So we create one interface to get info on them: enumerateDevices.
- We want to be able to get capabilities of input devices without actually opening the device. So we add capabilities to the info returned by enumerateDevices.
- For some properties (example: channel count), it's obvious how they would apply to output devices. It seems strange to need to go to a second API to get information about the properties of the output devices. Thus, it seems logical to add this too to information returned by enumerateDevices.

Each step seems logical. But if we're unhappy with the state at the end, we should think about which step is the one that doesn't hang together.

jan-ivar:
It seems to me we've forgotten to define output devices. Relying on their similarity to input devices, is the weak link in this reasoning imho.

joeberkovitz:
I would agree with @jan-ivar that there is a substantial gap in the media capture world right now regarding output devices, and that "piggybacking" a definition of output capabilities on top of input is a shaky approach.

I wouldn't have chosen this except that my understanding from the chairs has been that they want to defer discussions of how to fill this gap, until future consideration of the Audio Output API (which spec is really a kind of prototype that needs considerable revision to serve the most important use cases).

I suppose one question for the group here is whether deferral is desirable, if it leads to a device enumeration API that cannot expose a unified view of input and output devices.

alvestrand:
If it's "cannot expose now", it seems like an acceptable thing to me.
If it's "cannot expose ever", I'm more worried.

joeberkovitz:
I think that my current approach is to redo this change and rename the new attribute `inputCapabilities`, and indicate that it is always `null` for output devices. This avoids the issue that could potentially lead to trouble downstream, namely an assumption that the current interface definitions for capabilities would be usable across both input and output devices.

If no one objects I'll close this PR and create a new one along these lines.

stefhak:
@joeberkovitz to me that sounds like a sensible next step.

alvestrand:
Defining capabilities for output devices seems like a different question to be addressed, but the idea of exposing capabiilities in the device list seems good. If everyone agrees, we should merge this and add a new issue on output device caps.

alvestrand:
To @joeberkovitz ' comment: I think changing to "inputcapabilities" would be a mistake.
If we define output capabilities, they will either have the same form as for input, or not. If they have the same form, it seems strange to have a different field for them; if they don't, they can have a new field anyway.
Each device is already tagged as being an input or an output device.

joeberkovitz:
@alvestrand I'm fine with that approach, I suppose it has to be either input or output. As per our earlier discussion I believe an attempt to define output capabilities would be post-1.0 unless you want to resolve the other, larger API issues pertaining to output.

jan-ivar:
@alvestrand Extending `MediaTrackCapabilities` to include output capabilites sounds weird to me, if for not other reason than the name (will "media tracks" be used to represent outputs?)

So that would almost certainly leave us with `info.capabilities` and `info.outputCapabilites`, where the former is input capabilities. Are we fine with that?

stefhak:
@jan-ivar I think that would be fine.
@joeberkovitz could you update the PR to reflect this?

getusermedia.html
alvestrand: Seems like we should delete the sentence about output devices, and declare that we'll figure that one out later.

joeberkovitz:
@stefhak @jan-ivar I've changed this to remove support for output device capabilities, and also removed strong typing from the `capabilities` attribute in case we want to return different, output-oriented type in this attribute in the future. I thought this would be better than having both `capabilities` and `outputCapabilities` which seems strange, although we retain the potential to do that.

If this is a problem, we could leave the strong typing in.

stefhak:
With the exception of the part "Future changes to this API will address the acquisition of output device capabilities." this LGTM. I don't think that kind of forward looking predictions belong in a spec for implementers, that fits more into a group planning doc or similar.

joeberkovitz:
I've removed the statement about future spec intentions, and also changed the phrase "...will always have the value `null`" to simply "...has the value `null`" (since this was also forward-looking in a sense).

stefhak:
Thanks @joeberkovitz, this looks all good to me now. Let's see if any of the editors see any issue, or we can land this.

getusermedia.html
```diff
@@ -2404,6 +2405,19 @@
           MUST return the empty string.</p>
         </dd>
 
+        <dt>readonly attribute Object capabilities</dt>
```
jan-ivar: Did you mean lower-case `object`?

Even though the repercussions for using `object` are much less for read-only attributes than for anything else, they're not zero, and I still wince when I see it:
* For users it means reading prose to figure out what to expect, and possibly loads of defensive coding around every read of properties that really are meant to be invariant for all input devices.
* For implementers it means manually re-creating the copy-semantics offered by WebIDL binding code, that are absent with `object` (which uses the JS object reference semantic), which can lead to bugs.

I think we should strive to define our API types whenever possible and use `object` sparingly.

Also, in my experience (but not always), the WebIDL spec is written to guide spec writers into choosing patterns the Tag finds desirable, and away from ones it finds undesirable. One might take the absence of a way to specify a union of two dictionaries as such a hint. But see https://github.com/w3c/webrtc-pc/issues/275#issuecomment-135623998

joeberkovitz: @jan-ivar I did mean lowercase `object`. I realize this is undesirable, but you didn't respond to my earlier question in the pull request conversation as to whether this was a good idea or not.

Please see:
https://github.com/w3c/mediacapture-main/pull/211#issuecomment-135457504

jan-ivar: @joeberkovitz As I mentioned, I think we should strive to define our API types whenever possible and use `object` sparingly.

Using `object` today where we return an entirely known well-defined dictionary, just because we anticipate future changes, to me this falls under "forward looking predictions" which @stefhak says "don't [...] belong in a spec for implementers".

@alvestrand [says](https://github.com/w3c/mediacapture-main/pull/211#issuecomment-133334319) they're in the "same form", but that wont explain why input devices will invariantly return all capabilities while outdevices will not. I think calling it `inputCapabilities` will.

Having to define a new dictionary for output devices later, where some (but not all) of the members are the same, doesn't sound onerous to me, especially since these are not constraints.

joeberkovitz: Trying to understand what I will revise here since I am hearing two different proposals. @jan-ivar thinks we have separate strongly typed `inputCapabilities` and `outputCapabilities` attributes IIRC, while @alvestrand wants a single `capabilities` field.

I'll wait for some indication from the chairs including @stefhak  as I don't want to thrash back and forth on these edits further. I lean towards @jan-ivar's proposal of a single strongly typed field (`attribute MediaTrackCapabilities inputCapabilities`).

adam-be:
```
(MediaTrackCapabilities or OutputCapabilities) getCapabilities();
```
I can see a problem with union types being indistinguishable when the union is used as an input argument, but is it a problem when used as a function return value? The UA creates either type and then the app will know from the context what type it is.

jan-ivar:
@adam-be It shouldn't be a problem in theory, which is why I filed https://github.com/heycam/webidl/issues/57.

I think my issue would be we're talking about a future dictionary we haven't defined. I suppose if we defined `OutputCapabilities` now and left it empty, then future specs could use `partial dictionary OutputCapabilities` to add to it?

On the other hand, the value of this over just using `partial dictionary MediaTrackCapabilities` may seem marginal, except it solves the inconsistent naming perhaps.

I still worry the invariant becomes buried, and users will end up testing for every field (though arguably current implementations can be blamed a bit here).

adam-be:
To avoid creating an empty ```OutputCapabilites``` dictionary we could actually leave ```MediaDeviceInfo``` as is and derive a new ```InputDeviceInfo``` from it and let ```getCapabilites()``` live on that type with a simple return value. That way, output devices wouldn't have a ```getCapabilites()``` (they are of type ```MediaDeviceInfo```)  until we add the corresponding ```OutputDeviceInfo``` with its own ```getCapabilites()``` method.

jan-ivar:
> When you say "just using partial dictionary MediaTrackCapabilities", do you mean adding output capability members to that one then?

I'm not advocating doing that, but that's what I meant.

> derive a new InputDeviceInfo type from it and let getCapabilites() live on the new type with a simple return value.

I like that.

joeberkovitz:
@jan-ivar so with that last comment, are we back to the current PR being acceptable with its single `capabilities` field that will only be used for input?

jan-ivar:
@joeberkovitz I believe Adam is suggesting this:

    interface MediaDeviceInfo {
        readonly    attribute DOMString       deviceId;
        readonly    attribute MediaDeviceKind kind;
        readonly    attribute DOMString       label;
        readonly    attribute DOMString       groupId;
        serializer = {attribute};
    };

    interface InputDeviceInfo : MediaDeviceInfo {
        MediaTrackCapabilities          getCapabilities ();
    };

Note that [the webidl spec says](https://heycam.github.io/webidl/#idl-dictionaries): "Dictionaries MUST NOT be used as the type of an attribute or constant." - because they have copy-semantics.

It's a little odd perhaps that the capabilities are not serializable, but I can live with that.

adam-be:
Thanks for writing up the WebIDL @jan-ivar

Regarding serialization, I guess the capabilities won't be included unless we skip interfaces all together and go with dictionaries. But then we're back at the "union between dictionaries problem" again.

We could look into defining a serializer as an operation (supported in WebIDL according to the spec) or we could leave it to developers to add one in JavaScript.

joeberkovitz:
@jan-ivar @adam-be Can't the capabilities be serialized by defining `serializer = { inherit, attribute }` for `InputDeviceInfo`? I'm not a WebIDL maven, just reading the spec and wondering why this wouldn't work.

adam-be:
I believe the problem is that we need the value returned from the the ```getCapabilites()``` method to be included in the serialized result. It might work by defining an operation as a serializer because then you can run customs steps.

alvestrand:
Check: What requirement do we actually have on a serialization for either deviceinfo or capabilities?

If we don't have a specific requirement, we might want to leave this to applications.

adam-be:
I don't think we have such a requirement, and it's pretty easy to do in JavaScript.

jan-ivar:
I agree. As I said already I can live without the capabilities being serialized.

adam-be:
@joeberkovitz, do you feel that you've got all the info you need to update the PR accordingly?

joeberkovitz:
@adam-be I believe I would just update the WebIDL to include the new `InputDeviceInfo` as per @jan-ivar 's writeup, and explain that `InputDeviceInfo` is the concrete subtype of `MediaDeviceInfo` for input devices represented in the sequence returned by `enumerateDevices()`. Please correct me if this doesn't sound right.

adam-be:
:+1:

joeberkovitz:
@jan-ivar I think we should be good now

alvestrand:
Travis thinks there's a reference error in it, however.... and it's not the same one as the one on the main branch.
Can you take a look?

dontcallmedom:
The Travis issues actually look like the same as on master, so I don't think they are blocking (I'll take a stab at fixing the issues separately)

stefhak:
@jan-ivar @adam-be does this look good to you?

joeberkovitz:
It looks like gh-pages might have been reset on the repo to an earlier commit by someone. After fetching, my local gh-pages was suddenly ahead of the origin. I left this situation alone.

adam-be:
The Travis CI warnings seem to be unrelated. You could do a rebase if you want to get rid of them.

getusermedia.html
```diff
@@ -2329,7 +2333,8 @@
                   a copy of <var>resultList</var>, and all its elements, where
                   the <code><a href=
                   "#widl-MediaDeviceInfo-label">label</a></code> member is the
-                  empty string.</p>
+                  empty string and the <code><a href=
+                  "#widl-MediaDeviceInfo-capabilities">capabilities</a></code> member is <code>null</code>.</p>
```
adam-be: We don't have the capabilities member (attribute) anymore. I think you can simply remove this since the wanted behaviour is already documented in the description of getCapabilites().

getusermedia.html
```diff
@@ -2437,6 +2441,24 @@
...
     </section>
+    <section>
+      <h2>Input-Specific Device Info</h2>
+      <dl class="idl" title="interface InputDeviceInfo : MediaDeviceInfo">
+        <dt>MediaTrackCapabilities getCapabilities()</dt>
+
+        <dd>
+          <p>Returns a <code><a>MediaTrackCapabilities</a></code> object describing the device's audio or video track
+          (according to its <code>kind</code> value), in the absence of any user-supplied constraints.
+          For input devices these capabilities will be identical to those that would have been obtained by calling
```
adam-be: We can skip the "For input devices" now since it's always input devices.

joeberkovitz:
Thanks @adam-be, I should have caught those references. Should be good now.

getusermedia.html
```diff
@@ -2248,6 +2248,10 @@
           User Agent's available media input and output devices if enumeration
           is successful.</p>
 
+          <p>Elements of this sequence which represent input devices will be of type
```
jan-ivar: Use "that" with restrictive clause.

getusermedia.html
```diff
@@ -2248,6 +2248,10 @@
...
+          <code><a>InputDeviceInfo</a></code> which extends <code><a>MediaDeviceInfo</a></code>
+          with input-specific elements.</p>
```
jan-ivar: Not sure if elements is the right word. Can we just leave out "with input-specific elements"? Seems redundant.

getusermedia.html
```diff
@@ -2437,6 +2440,24 @@
         </dd>
       </dl>
     </section>
+    <section>
+      <h2>Input-Specific Device Info</h2>
```
jan-ivar: Lower case s (would seem follow the pattern in this document)

getusermedia.html
```diff
@@ -2437,6 +2440,24 @@
...
+
+        <dd>
+          <p>Returns a <code><a>MediaTrackCapabilities</a></code> object describing the device's audio or video track
+          (according to its <code>kind</code> value), in the absence of any user-supplied constraints.
+          These capabilities will be identical to those that would have been obtained by calling
+          <code>getCapabilities()</code> on the first <code><a>MediaStreamTrack</a></code> of this type in a
+          <code><a>MediaStream</a></code> returned by <code>getUserMedia({deviceId: <i>id</i>})</code>
+          where <i>id</i> is the value of the <code>deviceId</code> attribute of this <code>MediaDeviceInfo</code>.</p>
+          <p>If no access has been granted to any local devices and this <code>InputDeviceInfo</code> has been filtered
+          with respect to unique identifying information (see above description of <code>enumerateDevices()</code> result), then this method returns <code>null</code>.
```
jan-ivar: Please word-wrap this paragraph to at least stay within the github window, so we don't lose anything while reviewing. ;)

getusermedia.html
```diff
@@ -2437,6 +2440,24 @@
         </dd>
       </dl>
     </section>
+    <section>
+      <h2>Input-Specific Device Info</h2>
+      <dl class="idl" title="interface InputDeviceInfo : MediaDeviceInfo">
+        <dt>MediaTrackCapabilities getCapabilities()</dt>
+
+        <dd>
+          <p>Returns a <code><a>MediaTrackCapabilities</a></code> object describing the device's audio or video track
```
jan-ivar: Devices don't have tracks. Can you rephrase?

getusermedia.html
```diff
@@ -2437,6 +2440,24 @@
...
+          (according to its <code>kind</code> value), in the absence of any user-supplied constraints.
+          These capabilities will be identical to those that would have been obtained by calling
+          <code>getCapabilities()</code> on the first <code><a>MediaStreamTrack</a></code> of this type in a
```
jan-ivar: Write to the implementer. E.g. "These capabilities MUST match capabilities obtainable by calling `getCapabilities()` on all tracks of this type in a ...

jan-ivar:
lgtm on substance. some language nits.

joeberkovitz:
Thanks -- appreciate the pointers. Changes applied.

joeberkovitz:
OK, please let me know if there's a further next step I can take -- I don't have write access and can't merge this request.

stefhak:
@joeberkovitz this is now in the hands of the editors (and @adam-be is assigned). Hopefully it can be merged shortly. I'm removing the "submitter action needed" label.

adam-be:
This looks quite ready to me.

alvestrand:
With all the positive words, I'm hitting the merge button.

PR #213, opened 2015-07-07
Fix Not Found Error in gUM algorithm

burnburn:
It appears that the rewrite of the gUM algorithm to use SelectSettings inadvertently changed some of the errors.  Specifically, the NotFoundError should have been for the case where no suitable sources were available even before applying constraints, and the ConstraintNotSatisfiedError should have been used for the case where no suitable sources were found after applying constraints, meaning that at least one required constraint could not be satisfied.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2798,17 +2798,6 @@
                   <code><a>name</a></code> attribute has the value
                   <code>SecurityError</code>.</p>
                 </li>
-
-                <li>
-                  <p><em>Constraint Failure</em>: Let <var>error</var> be a new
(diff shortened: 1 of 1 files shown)


jan-ivar:
Doh! And thus constraint failure in Firefox fails with NotFoundError. :/ Good catch.

stefhak:
Yes, good catch. We should just merge this IMO.

burnburn:
Okay, I am merging.  Now for another attempt at merging #194 :)

stpeter:
Belated +1 - I looked at this earlier but neglected to reply. :(

PR #214, opened 2015-07-07
Create new ErrorEvent

burnburn:
This PR replaces #206.  The existing MediaStreamErrorEvent will be replaced with an OverconstrainedErrorEvent once PR #194 lands.  However, there is still a need for the ended event to be able to include an error for the cases when a track ends because of an error.  Adam and I believe this general ErrorEvent will be of use for this spec, and perhaps the WebRTC spec, anytime an event can occur as a result of an error.  Note that there is no requirement that the event always be an error.  The use of a generic ErrorEvent object is analogous to how we use the generic Event object today.


stefhak:
LGTM.

stpeter:
Double-checked and judged ready to merge.

PR #216, opened 2015-07-09
Define MediaStreamError exclusively for legacy gUM callback interface use

burnburn:
The legacy callback-based gUM interface needs an error callback, for which it needs an error type value.  This typedef makes clear that we now only have two such kinds of errors:  DOMException and OverconstrainedError.
This is in partial fulfillment of Issue #162.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3104,6 +3104,8 @@
           Error in obtaining a <code><a>MediaStream</a></code> as described in the failure steps of the <a href="#widl-NavigatorUserMedia-getUserMedia-void-MediaStreamConstraints-constraints-NavigatorUserMediaSuccessCallback-successCallback-NavigatorUserMediaErrorCallback-errorCallback"><code>navigator.getUserMedia()</code></a> algorithm.
         </dd>
       </dl>
+      <dl class="idl" title=
+      "typedef (DOMException or OverconstrainedError) MediaStreamError">
     </section>
 
     <section>


stefhak:
LGTM.

jan-ivar:
Remind me why we're breaking the legacy API?

burnburn:
@jan-ivar , is a typedef okay for this?

jan-ivar:
I don't think so, or rather the union is not ok. In order to be in a webidl union, types must be [distinguishable](https://heycam.github.io/webidl/#dfn-distinguishable), and I don't think these two are.

I'm not sure though since @bzbarsky and @domenic can't seen to agree what a DOMException is. ;) But then I'm not up on how our new errors are supposed to intermingle with regular errors, so I may be missing something.

For the legacy callbacks at least, could we leave them alone and require the UA to convert?

bzbarsky:
Yeah, this is broken in Web IDL right now.  Either `Error` should allow any error, including `DOMException`, or `Error` and `DOMException` should be distinguishable.  @heycam and @domenic and I need to sort this through.

Given this is a callback you're defining, you could probably use `object` in the IDL and prose and it would be ok for now.

burnburn:
@jan-ivar, would you be willing to suggest text and IDL acceptable to yourself that matches what bzbarsky is proposing?  You have been the strongest objector to use of object in IDL so far.

If you can't I'll magic up something, but I'd rather have you just Make It Right.

jan-ivar:
@burnburn Using `object` as @bzbarsky suggests seems fine for this legacy API. Being that it's only ever output from the browser, it doesn't have the security implications that `object` as an input has. I also can't think of another way.

jan-ivar:
I'll put something up.

burnburn:
Replaced by PR #226.

PR #217, opened 2015-07-09
remove DOMExceptions from error names table

burnburn:
Now that updates to the error names have been completed, can we just remove all the ones from this table that are actually DOMExceptions?

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2177,6 +2177,14 @@
     <section>
       <h3>Error names</h3>
 
+      <p class="note">There is a known issue with this table of error
+      names, because <code>MediaStreamError</code> has been removed
+      but is still needed in the legacy, callback-based getUserMedia
+      interface.  This issue will be fixed through resolution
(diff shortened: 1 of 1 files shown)


burnburn:
Posted to list for review and comment.

stefhak:
I think we should follow Anne's advice and remove the ones that are DOMExceptions.

stpeter:
Here is proposed text to place before the table:

      <p class="note">There is a known issue with this table of error names,
      because <code>MediaStreamError</code> has been removed but is still 
      needed in the legacy, callback-based getUserMedia interface.  This issue 
      will be fixed through resolution of <a 
      href="https://github.com/w3c/mediacapture-main/pull/216">PR #216</a>.
      </p>

stefhak:
That text LGTM.

burnburn:
@stpeter, please go ahead and add the text and merge the PR (or the other order, if that's easier.)

burnburn:
Looks like Peter's out, so I'll apply the change Peter suggests and merge this PR.

Issue #218, opened 2015-07-16
Persist deviceIds only after gUM permission grant.


jan-ivar:
[Discussed favorably on the list](https://lists.w3.org/Archives/Public/public-media-capture/2015Jul/0004.html), the idea is to limit exposure of persistent ids to the drive-by web.

stpeter:
Handled via manual merge: https://github.com/w3c/mediacapture-main/commit/1bc469f03d50dc9b0965ebac37a12a9eaa5bab4d

PR #219, opened 2015-07-16
Persist deviceIds only after gUM permission grant.

jan-ivar:
For issue #218. Limit deviceId stability to the current browsing session for the drive-by web. Ids are only persisted to disk if user grants camera or microphone.

This has been implemented in [Firefox Nightly](https://nightly.mozilla.org).

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2593,19 +2593,30 @@
           <p>A unique identifier for the represented device.</p>
 
           <p>All enumerable devices have an identifier that MUST be unique to
-          the application and persistent across browsing sessions. Unique and stable
-          identifiers let the application save, identify the availability of,
-          and directly request specific sources.</p>
+          the page's origin. As long as no local device has been attached to an
+          active MediaStreamTrack in a page from this origin, and no
+          <a href="#stored-permissions">persistent permission</a> to access
+          local devices has been granted to this origin, then this identifier
+          MUST NOT be reusable after the end of the current browsing session.</p>
 
-          <p>This identifier MUST be un-guessable by other applications to
-          prevent the identifier being used to correlate the same user across
-          different applications.</p>
+          However, if any local devices have been attached to an active
+          MediaStreamTrack in a page from this origin, or
+          <a href="#stored-permissions">persistent permission</a> to access
+          local devices has been granted to this origin, then this identifier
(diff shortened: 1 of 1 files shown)


jan-ivar:
Travis errors seem unrelated to this PR...

jan-ivar:
Travis errors seem unrelated to this PR...

dontcallmedom:
a combination of merging #220 and #216 (or some replacement of that one) would be needed to fix Travis failures. I'm not sure why we allowed master to get in a failing state here.

mhofman:
Looks great. One concern I have is that, as-is, UAs are technically allowed to reset persisted device identifiers independently of other persistent storages. This can cause a problem for the application logic relying on a saved id in cookies/localStorage, that no longer matches any existing deviceId.

jan-ivar:
@mhofman - true but unrelated to this PR, yes? FWIW the firefox implementation clears deviceIds at the same time it clears cookies.

Firefox also lets users delete cookies by age, or even individual cookies, so the guarantees you mention seem like they would be hard to keep.

mhofman:
@jan-ivar I agree there's nothing to do about individual cookies being reset manually, but I was wondering if we should provide some guidance for UAs to NOT provide an option in their "normal reset flow" to separately reset persisted ids outside of clearing cookies. Or maybe it just doesn't belong in the spec?
How is local storage reset specified in relations to cookies reset? I would think we would be in a very similar case here.

jan-ivar:
@mhofman Good point, as there are privacy reasons for clearing both as well. Maybe we could use something similar to [Treating persistent storage as cookies](https://html.spec.whatwg.org/multipage/webstorage.html#user-tracking)?

stefhak:
LGTM

stpeter:
I agree that this looks good and is an improvement. I suggest that we merge this text but look at it more holistically when we review the entire spec to make sure that we address the TAG's security questionnaire: https://w3ctag.github.io/security-questionnaire/

stpeter:
Merged manually because of conflicts: https://github.com/w3c/mediacapture-main/commit/1bc469f03d50dc9b0965ebac37a12a9eaa5bab4d#diff-ea76d38900f79cfae8f60e5f7cf16dd1

PR #226, opened 2015-08-05
typedef MediaStreamError as an object

jan-ivar:
From discussion in https://github.com/w3c/mediacapture-main/pull/216. I don't have any better ideas than this.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2031,7 +2031,7 @@
         <section>
           <h2>OverconstrainedError ( constraint, message )</h2>
 
-          <p>When the <code>OverconstrainedError</code> function is
+          <p>When the <code><dfn>OverconstrainedError</dfn></code> function is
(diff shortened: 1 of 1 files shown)


jan-ivar:
I matched the `OverConstrainedError` link to the existing use. Should they perhaps both link to [7.2](https://rawgit.com/w3c/mediacapture-main/8464a1db1f4beee6495e5ca8add81ad33a71a35f/getusermedia.html#overconstrainederror-object) instead?

burnburn:
Overall this looks good.  I need to check into which OverconstrainedError link to use at this point given other PRs but will do that.

burnburn:
Yes, it should link to 7.2 instead.  I believe the reference is #overconstrainederror-object.

jan-ivar:
There were a lot of links. Trying this instead.

Issue #204, opened 2015-06-29
Use [SameObject]


foolip:
Some attributes which return objects that appear to always be the same:
MediaStreamTrackEvent.track
MediaStreamErrorEvent.error (nullable, so I'm not sure if it's OK)
NavigatorUserMedia.mediaDevices

@heycam, is it OK to use [SameObject] on a nullable attribute?

heycam:
Per spec, no:

> The [SameObject] extended attribute MUST NOT be used on anything other than a read only attribute whose type is an interface type, an array type or object.

Maybe [SameObject] should become something that indicates that whatever value is returned will always be the same, regardless of whether it's an object.  Gecko's bindings generators support a [Constant] which means this.

foolip:
Hmm, OK. Do you want a spec bug suggesting that change to keep track of it?

Pending that, it seems like MediaStreamTrackEvent.track and NavigatorUserMedia.mediaDevices could be fixed.

stefhak:
Message sent to list https://lists.w3.org/Archives/Public/public-media-capture/2015Jul/0002.html

stefhak:
There was some discussion, and the conclusion so far seem to be "this is not harmful, let's do it".

adam-be:
Proposed fix: https://github.com/w3c/mediacapture-main/pull/228

foolip:
:+1:

burnburn:
PR is merged.

PR #228, opened 2015-08-10
Use WebIDL [SameObject] extended attribute (fix for issue #204)

adam-be:


alvestrand:
LGTM

stpeter:
The pointer for the [WEBIDL] reference is the http://www.w3.org/TR/WebIDL/ but [SameObject] is not defined in the approved WebIDL CR, only in the Editor's Draft for WebIDL Second Edition. Do we need to change the pointer to https://heycam.github.io/webidl/ and is that a stable reference?

PR #235, opened 2015-08-27
Remove static like definitions of attributes and methods

adam-be:
Fixed for MediaStream and MediaStreamTrack

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2718,7 +2718,7 @@
     media input devices available to the User Agent.</p>
 
     <p>When on an insecure origin [[mixed-content]], User Agents are encouraged
-    to warn about usage of <code>MediaDevices.getUserMedia</code>,
+    to warn about usage of <code>navigator.mediaDevices.getUserMedia</code>,
     <code>navigator.getUserMedia</code>, and any prefixed variants in their
     developer tools, error logs, etc. It is explicitly permitted for User
     Agents to remove these APIs entirely when on an insecure origin, as long as
@@ -2735,7 +2735,7 @@
 
         <p>First, the official definition for the getUserMedia() method, and
         the one which developers are encouraged to use, is now at <a href=
-        "#dom-mediadevices-getusermedia">MediaDevices.getUserMedia()</a>. This
+        "#dom-mediadevices-getusermedia">MediaDevices</a>. This
         decision reflected consensus as long as the original API remained
         available here under the Navigator object for backwards compatibility
         reasons, since the working group acknowledges that early users of these
@@ -2748,8 +2748,8 @@
 
         <p>Second, the decision to change all other callback-based methods in
         the specification to be based on Promises instead required that the
-        Navigator.getUserMedia() definition reflect this in its use of the
-        MediaDevices.getUserMedia() method. Because Navigator.getUserMedia() is
+        navigator.getUserMedia() definition reflect this in its use of the
+        method on MediaDevices with the same name. Because navigator.getUserMedia() is
         now the only callback-based method remaining in the specification,
         there is ongoing discussion as to a) whether it still belongs in the
         specification, and b) if it does, whether its syntax should remain
@@ -2778,12 +2778,14 @@
           <p>The <var>successCallback</var> will be invoked with a suitable
           <code><a>MediaStream</a></code> object as its argument if the user
           accepts valid tracks as described in <a href=
-          "#dom-mediadevices-getusermedia">MediaDevices.getUserMedia()</a>.</p>
+          "#dom-mediadevices-getusermedia">getUserMedia()</a> on
+          <code><a>MediaDevices</a></code>.</p>
 
           <p>The <var>errorCallback</var> will be invoked if there is a failure
           in finding valid tracks or if the user denies permission, as
           described in <a href=
-          "#dom-mediadevices-getusermedia">MediaDevices.getUserMedia()</a>.</p>
+          "#dom-mediadevices-getusermedia">getUserMedia()</a> on
+          <code><a>MediaDevices</a></code>.</p>
 
           <p>When the <code id=
           "dom-navigatorusermedia-getusermedia">getUserMedia()</code> method is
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM.

adam-be:
Pushed an additional commit that removes wording like MediaDevices.getUserMedia().

stefhak:
Does anyone have a problem with this PR? If not I think we should merge it. As I read it the changes are only editorial.

getusermedia.html
```diff
@@ -449,7 +449,7 @@
           to that string. A good practice is to use a UUID [[rfc4122]], which
           is 36 characters long in its canonical form.</p>
 
-          <p>The <dfn id="dom-mediastream-id"><code>id</code></dfn> attribute
+          <p>The <dfn data-lt-noDefault id="dom-mediastream-id"><code>id</code></dfn> attribute
```
jan-ivar: Fyi, there's another 'id' in the paragraph above this one that you missed (the two now look different). I also wonder if this could be rewritten to avoid the double mention (let alone single mention).

getusermedia.html
```diff
@@ -629,9 +627,10 @@
       User Agents to manipulate media in whatever fashion is most suitable on
       the user's platform.</p>
 
-      <p>A script can indicate that a track no longer needs its source with the
+      <p>A script can indicate that a <code><a>MediaStreamTrack</a></code> no
```
jan-ivar: The paragraph directly above this one uses "MediaStreamTrack object". Seems inconsistent.

getusermedia.html
```diff
@@ -893,7 +893,7 @@
 
           <dd>
             <p>The <dfn id=
-            "dom-mediastreamtrack-kind"><code>MediaStreamTrack.kind</code></dfn>
+            "dom-mediastreamtrack-kind"><code>kind</code></dfn>
             attribute MUST return the string "<code>audio</code>" if the object
             represents an audio track or "<code>video</code>" if object
```
jan-ivar: "the object" (not your patch but nearby)

getusermedia.html
```diff
@@ -953,8 +954,8 @@
           <dt>readonly attribute boolean muted</dt>
 
           <dd>
-            <p>The <dfn id=
-            "dom-mediastreamtrack-muted"><code>MediaStreamTrack.muted</code></dfn>
+            <p>The <dfn data-lt-noDefault id=
+            "dom-mediastreamtrack-muted"><code>muted</code></dfn>
```
jan-ivar: no-op link

getusermedia.html
```diff
@@ -1014,8 +1015,7 @@
           <dd>
             <p>Clones the given <code><a>MediaStreamTrack</a></code>.</p>
```
jan-ivar: (nearby): s/the given/this/ (clone is given no arguments)

getusermedia.html
```diff
@@ -2748,8 +2748,8 @@
 
         <p>Second, the decision to change all other callback-based methods in
         the specification to be based on Promises instead required that the
-        Navigator.getUserMedia() definition reflect this in its use of the
-        MediaDevices.getUserMedia() method. Because Navigator.getUserMedia() is
+        navigator.getUserMedia() definition reflect this in its use of the
+        method on MediaDevices with the same name. Because navigator.getUserMedia() is
```
jan-ivar: maybe "method on navigator.mediaDevices with the same name." ?
or even "navigator.mediaDevices.getUserMedia()." ?

getusermedia.html
```diff
@@ -2778,12 +2778,14 @@
           <p>The <var>successCallback</var> will be invoked with a suitable
           <code><a>MediaStream</a></code> object as its argument if the user
           accepts valid tracks as described in <a href=
-          "#dom-mediadevices-getusermedia">MediaDevices.getUserMedia()</a>.</p>
+          "#dom-mediadevices-getusermedia">getUserMedia()</a> on
```
jan-ivar: Not a comment on this patch, but a general comment I don't know where else to put:

I see we commonly refer to a function `foo` as `foo()` in this spec. None of the cases in this patch trouble me, but other use like "Subsequent calls to getUserMedia()" is only unambiguous by the fact that zero arguments are invalid for `getUserMedia`. Just wanted to mention it. All uses in this patch (`clone` and `stop`), except `getUserMedia`, happen to have zero arguments.

adam-be: I think the parentheses on function names improves readability even though you cannot copy paste the text as a proper call into your code.

Looking quickly at other specs, it seem to be quite common. The XHR spec even has it section names when they document functions. [1]

[1] https://xhr.spec.whatwg.org/

getusermedia.html
adam-be: Are you referring to the id link in the paragraph above? Since it's a link, and the highlighted text is the definition, they need to look different.

getusermedia.html
adam-be: Fixed

getusermedia.html
adam-be: Let's fix that as well (I think "this object" is better than "the object")

getusermedia.html
adam-be: Can you clarify what you mean here?

getusermedia.html
adam-be: Fixed

getusermedia.html
adam-be: Lets use the same language for both (you second proposal).

adam-be:
@jan-ivar, can you take a look at the latest changes?

getusermedia.html
jan-ivar: Fair enough, and thanks for the info.

jan-ivar:
lgtm!

getusermedia.html
jan-ivar: (whoops, missed some comments. Thanks github and outdated diff!)

Makes sense now. I was probably expecting this paragraph (with definition) to come before the other.

getusermedia.html
jan-ivar: nm. I just learned about data-lt-noDefault.

Issue #222, opened 2015-07-28
Clarify whether groupId is origin-unique


jyasskin:
In #202, @jan-ivar pointed out that [`MediaDeviceInfo.groupId`](https://w3c.github.io/mediacapture-main/#widl-MediaDeviceInfo-groupId) isn't explicitly specified to be origin-unique. However, the [`enumerateDevices`](https://w3c.github.io/mediacapture-main/#widl-MediaDevices-enumerateDevices-Promise-sequence-MediaDeviceInfo) algorithm sets `groupId` to "a newly generated unique identifier" if no device in _oldList_ is in the same physical device. I believe that has to be origin-scoped in order for `deviceId` to be origin-scoped.

_oldList_ is bound by
> If this method has been called previously within this browsing session, let oldList be the list of MediaDeviceInfo objects that was produced at that call (resultList); otherwise, let oldList be an empty list.

And then objects from it are used to populate the return value of `enumerateDevices()`:
> If device is represented by a MediaDeviceInfo object in oldList, append that object to resultList...

That's a little confused: If the method was called in a different execution environment in the same browsing session, this returns a MediaDeviceInfo object from a different global object, so `instanceof` checks won't work. But you can't just limit to previous calls in the current execution context or `groupId` won't be shared within the origin.

alvestrand:
In my opinion, it's obviously origin-unique, but I could be wrong.

jan-ivar:
Here's the trouble I have reading it: `enumerateDevices` is a global method, and in its [processing steps](https://w3c.github.io/mediacapture-main/#dom-mediadevices-enumeratedevices) I find no mention of origin-uniqueness, unless there is some definition of "browsing session" that includes origin-specific context. I thought it just meant "for as long as the browser is running".

In other words, if not for the requirement given under [deviceId](https://w3c.github.io/mediacapture-main/#widl-MediaDeviceInfo-deviceId), I think one could be forgiven for thinking the list returned was global.

Absent similar wording under [groupId](https://w3c.github.io/mediacapture-main/#widl-MediaDeviceInfo-groupId), it seems at least difficult to deduce that it must be origin-unique as well.

stefhak:
I looked into "browsing session" a bit (or rather, @dontcallmedom helped me), and apparently it is not defined. But it is OTOH used quite a bit (usage in e.g.: http://www.w3.org/TR/geolocation-API/, 
[...]

PR #240, opened 2015-09-03
specify that groupId is origin-unique

stpeter:
Intended to fix https://github.com/w3c/mediacapture-main/issues/222


stefhak:
LGTM

PR #256, opened 2015-10-07
stop using IANA registry

fluffy:
PR for discussion of what we should do


stefhak:
This PR does not seem to be in line with what we decided [1] to to given the vote result.
[1] https://lists.w3.org/Archives/Public/public-webrtc/2015Oct/0007.html

PR #258, opened 2015-10-14
Reorganize definition of constrainable properties without a registry

dontcallmedom:
Remove IANA registry
Move definitions of MediaStreamTrack constrainable properties to subsection of MediaStreamTrack section
Remove references to registered values and instead make references to properties


dontcallmedom:
I took some liberty from the original proposed follow-up
> the text in section 11.2 will be replaced with “See section 14 for constraints defined by this specification. Other specifications may define additional constraints.

because it didn't seem to be very logical to separate so much the definition of constrainable properties for MediaStreamTrack from the definition of MediaStreamTrack itself.

PR #260, opened 2015-10-14
Update privacy section with lessons from TAG questionnaire

### LINKED ISSUE #212: Privacy/security review questions
The W3CTAG is asking new specs to include a section with answers to the questions at https://w3ctag.github.io/security-questionnaire/. For an example, see https://github.com/w3ctag/security-questionnaire/issues/1. 

If possible, I'd encourage expanding the Privacy and Security Considerations section to include answers to all of the questions in the questionnaire. (Or add an appendix with the answers.)

dontcallmedom:
close #212

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -4331,9 +4331,12 @@ var gotten = navigator.mediaDevices.getUserMedia({video: {
     explicit user consent before sharing it.</p>
 
     <p>Without authorization (to the "drive-by web"), it offers the ability to
-    tell how many devices there are of each class. The identifiers for the
+    tell how many devices there are of each class, and how they are grouped
+    together (e.g. a microphone and camera belonging to a single Web cam). The
+    identifiers for the
     devices are designed to not be useful for a fingerprint that can track the
-    user between origins, but the number of devices adds to the fingerprint
+    user between origins, but the number and grouping of devices adds to the
+    fingerprint
     surface. It recommends to treat the per-origin persistent identifier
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -4348,6 +4353,19 @@
...
+    been granted authorization, and thus potentially provide a way to track a
+    given device across time and origins.</p>
+
+    <p>Once a developer gains access to a media stream from a capture device,
+    the developer also gains access to detailed information about the device,
+    including its range of operating capabilities (e.g. available resolutions
+    for a camera). These operating capabilities are for the most part
+    persistent across browsing sessions and origins, and thus provide a way to
+    track a given device across time and origins.</p>
+
```
alvestrand: Note: For video devices, getting a video stream from the device will almost always give a reliable fingerprint of the device (for instance by dead pixel detection). I think we've agreed in the past that once you have a video stream from a camera, fingerprint defense is a lost cause. Should we say so explicitly in the text?

dontcallmedom: I've added a paragraph to that effect.

getusermedia.html
```diff
@@ -4348,6 +4353,23 @@
     interaction until permission is given - either by offering a way to say a
     "persistent NO" or by not using a modal permissions dialog.</p>
 
+    <p>When authorization is given, application developers gain access to the
```
alvestrand: Should probably say "When authorization to any media device is given", or "When authorization to a single media device is given".

PR #261, opened 2015-10-14
Warn against UUID fingerprinting

### LINKED ISSUE #255: Generation of IDs needs consideration
The current spec says:

"When a MediaStream object is created, the User Agent must generate an identifier string, and must initialize the object's id attribute to that string. A good practice is to use a UUID [rfc4122], which is 36 characters long in its canonical form."

Several of the UUID forms in RFC 4122 leak information that allow tracking, such as the MAC address of the host.

Suggested fix: Add to the above:

"To avoid fingerprinting, Implementations SHOULD use the forms in section 4.4 or 4.5 of RFC 4122 when generating UUIDs."

dontcallmedom:
close #255
from @alvestrand proposed text


alvestrand:
lgtm

Issue #250, opened 2015-09-21
PING: Refer to security-arch on the MUST inside Best Practice for HTTPS


alvestrand:
The "best practice" on storing permissions in section 10.6 "Implementation suggestions" has a MUST in it. It should be rephrased to make it clear that it's saying that security-arch section 5.2 has the MUST, and that it's quoting it, not establishing a new requirement.

This also requires adding security-arch to references, and making the reference normative.

(The alternative is to move it to a place where normative text is appropriate.)

alvestrand:
Fixed in #266

alvestrand:
The fix for #250 did not make it into #266 (it was in #265). Line 3273.

aboba:
Sorry, fixed in PR 278:
https://github.com/w3c/mediacapture-main/issues/250

alvestrand:
Done in #278

Issue #249, opened 2015-09-21
PING: Document tradeoff for non-HTTPS usage of getUserMedia


alvestrand:
From Nick Doty's mail on behalf of PING:

"You've heard from the TAG already about whether use of the API ever makes sense in unprivileged contexts. That is, when the user is asked for permission to access their camera, do they understand that they're granting this permission to all network attackers as well as the site they think they're talking to? I suspect this PING email thread is not going to change your minds about that already discussed topic. However, it would be worthwhile to note this security threat in the security considerations section and to note for user agent implementers the difficulty for this permission prompt."

This does not suggest a technical change in when getUserMedia is permitted, but does suggest that section 13 (security and privacy) should have some text explaining the reasoning behind the current spec.

alvestrand:
Fixed in #266

PR #265, opened 2015-10-23
Refer to security-arch for HTTP (and HTTPS) Best Practice (Issues 249 and 250)

aboba:
Addresses the following issues: 

https://github.com/w3c/mediacapture-main/issues/249
https://github.com/w3c/mediacapture-main/issues/250

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -4445,6 +4445,14 @@ var gotten = navigator.mediaDevices.getUserMedia({video: {
       audio/video streams to an attacker <code>EvilSpy</code> by following a
       link or being redirected to
       <code>https://webrtc.example.org/?user=EvilSpy</code>.</p>
+
+      <p>Although [[!RTCWEB-SECURITY-ARCH]] Section 5.2 indicates that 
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -4444,6 +4445,14 @@
...
       link or being redirected to
       <code>https://webrtc.example.org/?user=EvilSpy</code>.</p>
+
+      <p>Although [[!RTCWEB-SECURITY-ARCH]] Section 5.2 indicates that 
+      implementations may refuse all access permissions for HTTP origins,
+      it recommends that implementations allow one-time camera/microphone 
+      access. While allowing one-time access for HTTP origins is 
+      convenient, this makes it possible for an attacker to obtain access
+      to the camera/microphone of an unsuspecting user.  Therefore this 
+      cannot be considered a sound security practice.</p>
```
alvestrand: We don't have consensus on this paragraph AFAIK - the access in this case would cause a permission prompt to pop up, which the WG has previously thought adequate. Suggest removing from this PR.

Issue #242, opened 2015-09-08
Management of track members in a stream


martinthomson:
The spec is currently pretty lean on detail when it comes to where the set of track is managed.  There are multiple actors that might want to do that.  We should be clear on what the rules of engagement are.

martinthomson:
The primary case of concern here is that the spec doesn't recognize that the set of tracks in a stream might be manipulated by things that are out of the control of an application.  In particular, `RTCPeerConnection` and it's session negotiation.

I don't think that this requires much, but `removeTrack` (or maybe **remove a track** from a `MediaStream`) isn't specified at all.

alvestrand:
@aboba will take care of this

aboba:
Section 4.2.3 defines the basic operation of removeTrack().  

We could add more detail in Section 4.2, such as the following: 

To remove a track to a MediaStream, the User Agent must run the following steps:

1. Let track be the MediaStreamTrack in question and stream the MediaStream object to which track is to be removed.

2. If track is not in the stream's track set, then abort these steps.

3. Remove the track from the stream's track set.

alvestrand:
The spec for the  removetrack event is together with addtrack in section 4.4 (casing confused me a bit).
Seems that we have the spec, it might need to be easier to find somehow.

alvestrand:
See #276 for the issue about which spec talks about event firing.

PR #266, opened 2015-10-23
Add detailed steps for removeTrack() and fix a security considerations issue.

aboba:
Fix for:
https://github.com/w3c/mediacapture-main/issues/242
https://github.com/w3c/mediacapture-main/issues/249
https://github.com/w3c/mediacapture-main/issues/250

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -423,6 +423,29 @@
           "#track-set">track set</a>.</p>
         </li>
       </ol>
+      
+            <p>To <dfn id="stream-remove-track">remove a track</dfn> from a
+      <code><a>MediaStream</a></code>, the User Agent MUST run the
+      following steps:</p>
+
+      <ol>
+        <li>
+          <p>Let <var>track</var> be the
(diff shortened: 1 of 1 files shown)


stefhak:
This looks like a good clarification to  me, we already have similar steps for addTrack.

getusermedia.html
```diff
@@ -423,6 +423,29 @@
...
+        </li>
+
+        <li>
+          <p>If <var>track</var> is not in <var>stream's</var> <a href=
+          "#track-set">track set</a>, then abort these steps.</p>
+        </li>
+
+        <li>
+          <p>Remove <var>track</var> from <var>stream</var>'s <a href=
+          "#track-set">track set</a>.</p>
```
alvestrand: Do we need to say "if this was not done by calling removeTrack, queue a task to fire the removetrack event", just like we have for addTrack?

alvestrand: Turns out that the language I was remembering is off in webrtc, not here. It should probably be here, but that's worth another PR.

Issue #232, opened 2015-08-19
clone a track: "inherit" not defined


annevk:
It's not clear what this means.

stpeter:
Perhaps this change makes sense...

OLD
Let trackClone inherit track's underlying source, kind, label, readyState, and enabled attributes, as well as its currently active constraints.

NEW
Set the source, kind, label, readyState, enabled attributes, and constraints for trackClone to the values currently active for track.

annevk:
Yeah that makes more sense, though it's still somewhat wrong, since "IDL attributes" are actually getters and setters. So you'd have to copy over the internal state and not muddle with the publicly-facing stuff.

adam-be:
Sure the IDL attributes are getters and setters, but it becomes quite verbose to define an internal representation for each and every attribute when it represents a simple value.

I recall often seeing spec text saying something like:
"Change the foo attributes value to 'bar'"
and
"The foo attribute MUST return the value to which it was last set by the User Agent."

That doesn't mean that it's 100% correct, but that it seems like a fairly common shortcut.

annevk:
It is, but it's also wrong. I've filed a bug on IDL to start including the notion of internal slots to make this easier on specification authors.

domenic:
+1 internal slots. Until IDL has support, see for example https://domenic.github.io/streaming-mediastreams/#media-stream-recorder which is my latest attempt at this. Yes it can be a bit verbose, but the gain in precision is very helpful to implementers and readers. Setting a readonly attribute is really wrong.

alvestrand:
Adam will do what stpeter suggested.

alvestrand:
Fixed by #270

PR #270, opened 2015-11-05
Clarify how a cloned track is initialized (remove 'inherit' language) (Issue  #232)

adam-be:
Proposed fix for #232

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -671,14 +671,23 @@
         </li>
 
         <li>
-          <p>Let <var>trackClone</var> inherit <var>track</var>'s underlying
-          source, <code><a href=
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -671,14 +671,23 @@
...
+        </li>
+
+        <li>
+          <p>Let <var>trackClone</var>'s underlying source be the source of
+          <var>track</var>.</p>
+        </li>
+
+        <li>
+          <p>Set <var>trackClone</var>'s constraints to the active constrains
+          of <var>track</var>.</p>
```
alvestrand: Since constraints are dictionaries, this means that the constraints are copied, not linked. Is that obvious enough?

PR #272, opened 2015-11-14
Fix applyConstraints steps to avoid unused var and bail on reject. 

jan-ivar:
The steps never bail after rejecting the promise, causing *newConstraints* to be applied regardless of failure. I also removed the use of the undefined *existingConstraints* variable.


alvestrand:
LGTM

PR #273, opened 2015-11-17
Mention stored permissions in text about stopping sources (fixes #269)

### LINKED ISSUE #269: Inconsistency in releasing permissions for a source
In section http://w3c.github.io/mediacapture-main/getusermedia.html#mediastreamtrack:

"A script can indicate that a track no longer needs its source with the MediaStreamTrack.stop() method. When all tracks using a source have been stopped, the given permission for that source is revoked and the source is stopped.......

If there is no stored permission to use that source, the User Agent should also remove the "permission granted" indicator for the source."

I believe the second sentence is correct, and the first sentence is not: If there is a stored permission, that is not affected by closing, but if there is no stored permission, the permission is revoked when the device is not used. Both sentences should use the same wording, and possibly merge.

adam-be:


alvestrand:
LGTM

PR #284, opened 2015-11-25
Clarify that constraint in overconstrainederror is defined in selectsettings

### LINKED ISSUE #248: Conflicting information about OverconstrainedError.constraint
[7.4 Error names](http://w3c.github.io/mediacapture-main/getusermedia.html#error-names) says:

> Name | Description | Note
> ---------|-----------------|-------
> OverconstrainedError | One of the mandatory Constraints could not be satisfied. | The constraint attribute is set to the name of the constraint *that caused the error*

while [10.2.1 getUserMedia](http://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-getusermedia) says:
> If *finalSet* is the empty set, let constraint be any required constraint whose fitness distance was infinity for all settings dictionaries examined while executing the **SelectSettings** algorithm

This is not the same thing. For example, if the following fails with `OverconstrainedError`:

    { width: { min: 1280 }, frameRate: { min: 60 }, facingMode: { exact: "user" } }

then it's because there are no resolutions at or above 1280 pixels wide that *also* do 60Hz *and* face the user, even though there may be plenty of 1280x720+ resolutions, 60Hz modes, and cameras that face the user.

The first definition fails because there is no single "constraint that caused the error".

The second definition works, except it returns "", because none of the three constraints have infinite fitness distance from *all* the settings dictionaries.

I need to know which one to implement.

### LINKED ISSUE #257: Non-existent constraints attribute in 4.3.2
[4.3.2 Tracks and Constraints](http://w3c.github.io/mediacapture-main/getusermedia.html#tracks-and-constraints) says:

"Each track maintains an internal version of the Constraints structure, namely a mandatory set of constraints (no duplicates) and an optional ordered list of individual constraint objects (may contain duplicates). The internal stored constraint structure is exposed to the application by the constraints attribute, and may be modified by the applyConstraints() method."

This paragraph seems redundant and outdated. There is no constraints attribute anymore, the internal description seems old and [getSettings](http://w3c.github.io/mediacapture-main/getusermedia.html#widl-ConstrainablePattern-getConstraints-Constraints) seems to cover everything else.

Also, the undefined term *Constraints structure* is used throughout, where I think we mean *Constraints dictionary*.

dontcallmedom:
close #248

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2584,7 +2584,8 @@
             satisfied.</td>
 
             <td>The <code>constraint</code> attribute is set to the name
-            of the constraint that caused the error</td>
+            of the constraint (if one specifically) that caused the error
+            as described in the <a>SelectSettings</a> algorithm</td>
           </tr>
         </tbody>
       </table>


burnburn:
I think a slightly more precise change would be 

"of the constraint (if one specifically) that caused the error"  to
"of any one of the constraints that caused the error, if any".

burnburn:
Based on the call today I will do a manual merge that includes the change I proposed (which the chairs/editors agreed to).

PR #286, opened 2015-12-03
Fix facingMode in Capabilities example

adam-be:
The dictionary member is a DOMString [1]

https://w3c.github.io/mediacapture-main/#media-track-capabilities

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1288,9 +1288,14 @@
 
           <dd />
 
-          <dt>DOMString facingMode</dt>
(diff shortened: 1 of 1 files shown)


alvestrand:
The examples need to show that capabilities can be an array of DOMString, but "user/environment" sounds kind of weird, since they're pointing in opposite directions.
Changing the example to "user" and "left" would make it more realistic (those are facing the same way, and could possibly be combined).

adam-be:
Updated. Ping @alvestrand.

alvestrand:
lgtm

stefhak:
LGTM.

PR #287, opened 2015-12-03
MediaTrackCapabilities echoCancellation member should be a boolean

adam-be:


alvestrand:
Capabilities needs to be sequence<boolean> so that "true", "false" is allowed.

PR #289, opened 2015-12-04
 Fix constraint description and add capabilty description for echoCancellation

adam-be:


getusermedia.html
```diff
@@ -1306,7 +1306,13 @@
 
           <dt>sequence&lt;boolean&gt; echoCancellation</dt>
 
-          <dd />
+          <dd>
+            <p>If the User Agent cannot do echo cancellation a single
+            <code>false</code> is reported. If echo cancellation cannot be
+            turned off, a single <code>true</code> is reported. If the script
+            control the feature, the User Agent reports a list with both
```
jan-ivar: Did you mean "can control" or "controls"?

adam-be: The first. I'll fix it.

stefhak:
LGTM.

getusermedia.html
```diff
@@ -1306,7 +1306,13 @@
 
           <dt>sequence&lt;boolean&gt; echoCancellation</dt>
 
-          <dd />
+          <dd>
+            <p>If the User Agent cannot do echo cancellation a single
+            <code>false</code> is reported. If echo cancellation cannot be
```
adam-be: I comment on my own patch here just to raise a question. Should unsupported features even be present in the capabilities dictionary? We can't require that since it would break forward compatibility with future features.

Ping @burnburn

alvestrand: Good question. I like this text because it defines what the status is instead of just saying "it's unsupported" (which could be either).
In this case, "unsupported" should mean "I know that echo cancellation is not present", so "false" makes sense.

PR #295, opened 2015-12-14
Resize images to fit in 800px width of new TR style

dontcallmedom:
per https://lists.w3.org/Archives/Public/public-media-capture/2015Dec/0042.html


burnburn:
I think Travis sent me the original PPT for the images a couple of years ago.  They should probably go into github so it's easier to make changes.

Issue #293, opened 2015-12-10
Is there any sense in having defaults for SupportedCapabilities?


alvestrand:
http://w3c.github.io/mediacapture-main/getusermedia.html#media-track-supported-constraints says that MediaTrackSupportedCapabilities has default value of "true" for everything.

This dictionary is only returned - it's never an argument to anything.
Does having defaults even make sense here?

alvestrand:
This is really an editorial fix, it doesn't matter to code (as far as I can tell).
The only advantage of deleting defaults is that it doesn't cause people who look for "default boolean to true" and think it's an issue irritation.

alvestrand:
also see https://github.com/w3c/webrtc-pc/issues/375 about "true" as a default for Boolean.

adam-be:
From the WebIDL spec [1]:
"Dictionary members can also optionally have a default value, which is the value to use for the dictionary member when passing a value to a platform object that does not have a specified value."

"Warning
As with operation argument default values, is strongly suggested not to use of true as the default value for boolean-typed dictionary members, as this can be confusing for authors who might otherwise expect the default conversion of undefined to be used (i.e., false)."

My interpretation is that we can't use default values to describe what we want here since the dictionary is only returned. We need to describe in prose how to initialize the dictionary.

[1] http://heycam.github.io/webidl/#dfn-dictionary-member-default-value

jan-ivar:
Having default values absolutely make sense, because once implemented, it always returns true.

There are real beneficial effects here: The default in webidl causes webidl binding code to furnish the dictionary with true values implicitly whenever an instance is created, whether returning dictionaries to content or taking them as arguments, doesn't matter. This works in Mozilla's webidl compiler, and I would venture it works everywhere given the limited number of ways to implement this.

If anything, I would remove the prose instead, since English is inferior to WebIDL here. In fact, zero code is needed in the Firefox implementation. Take a look, it's pretty neat: http://mxr.mozilla.org/mozilla-central/source/dom/media/MediaDevices.h#37

adam-be:
If the default values work for output dictionaries as well, then I agree that it makes sense. When I checked the WebIDL spec I didn't think that was particularly clear.

Looking again at the spec, if you combine the fact that "Dictionary members with default values are always considered to be present." [1] with the text about how a Dictionary is converted to an ECMAScript Object [2], then it makes sense.

I'll update PR #300.

[...]

PR #301, opened 2016-01-07
getSupportedConstraints(): Remove prose about dictionary member default values (fixes: #293)

adam-be:
Replaces PR #300


stefhak:
LGTM.

Issue #296, opened 2015-12-16
How to handle known but unsupported constraints


adam-be:
See list discussion:
https://lists.w3.org/Archives/Public/public-media-capture/2015Dec/0056.html

alvestrand:
When the constraint is known but useless (echoCancellation, only "false" supported), the text needs to say that it's supported, but only with one value. @adam-be  will check that the text is consistently clear on this subject.

adam-be:
I tried to clarify this a bit in PR #307.

jan-ivar:
I think there is value in having UAs not return `true` in supportedConstraints until they can reliably implement it. The reason is that for a constraint `foo`:

As a web-dev, if browser X returns `false` in supportedConstraints then I know the browser does not support it, and I can tell the user their browser is the problem, not (necessarily) their hardware.

If browser X returns `true` in supportedConstraints but then I get a single zeroish value in all getCapabilities(), then it would be nice if I could assume that the browser does support it, and I could unequivocally tell the user their browser is fine, but that they need to run out and buy more hardware.

adam-be:
So what you're saying @jan-ivar is that you think it's OK to return a single "off" as a capability if the reason a feature is not supported is out of the browser's control. Have I understood you correctly?

jan-ivar:
It's hard to infer a general rule from the `echoCancellation` constraint, given how complicated it is, conflating both feature vs. value (boolean), and inherent-property vs. control-setting (there are mics with controllable hardware aec, mics with hardware aec that cannot be turned off, mics without aec, and browsers that do software aec), but I'll try.

Say my system has two mics, `micA` with hardware aec ***that cannot be turned off***, and `micB` without aec.

If a browser claims to support the `echoCancellation` constraint, and the browser

1. does ***not*** support software aec, then I would expect `devInfoMicA.getCapabilities().echoCancellation` to yield `[ true ]`, and `devInfoMicB.getCapabilities().echoCancellation` to yield `[ false ]`.

2. If it ***does*** support software aec, then I might expect
`devInfoMicA.getCapabilities().echoCancellation` to yield `[ true ]`, and `devInfoMicB.getCapabilities().echoCancellation` to yield `[ false, true ]`.

3. If it ***does*** support software aec, but ***cannot glean hardware capabilities***, then I might settle for
[...]

PR #307, opened 2016-01-21
Add text about Capabilities and features that can only in mode 'off' (issue: #296)

adam-be:


adam-be:
Travis seems to be in a bad mood today. The line wrap test succeeds locally for me.

getusermedia.html
```diff
@@ -3569,7 +3569,9 @@
       constrainable properties and values that the User Agent supports. Note
       that a User Agent MAY support only a subset of the properties defined in
       the Web platform, and MAY support a subset of the set values for those
-      properties that it does support. Note that Capabilities are returned from
+      properties that it does support. A constrainable property, representing a
+      particular feature, is considered supported even if the only allowed
```
martinthomson: I think that you need to be much clearer here.  

> An implementation might implement a particular constrainable property for one type of source and not another.  On sources where the property cannot be used, a constraint MUST be supported, but the implementation MAY limit that support to a value where the feature is disabled, or even a value that cannot be expressed.  For instance, a `width` constraint does not apply to audio sources, but this would be reported as being supported for audio sources.  Attempting to set a `width` constraint on an audio source would result in the constraint not being met.

Nit: please use paragraphs for new concepts.   This document is very hard to read as it is.

jan-ivar: I'm not sure what we're trying to say here. In my mind, a `width` property is completely orthogonal to audio sources, so audio sources will do best to ignore it completely IMHO. I didn't think the issue here was about diversity of source types, but rather about whether it would be ok for UAs to acknowledge constraints it had dud implementations of, something I think is ill advised, as I mention in https://github.com/w3c/mediacapture-main/issues/296#issuecomment-173726112.

martinthomson: This is about dealing with [getSupportedConstraints()](https://w3c.github.io/mediacapture-main/#widl-MediaDevices-getSupportedConstraints-MediaTrackSupportedConstraints).

Since the value there is global and not separated by media device type, this rule is required.

jan-ivar: Why does the fact that `getSupportedConstraints()` has a shared namespace mean that audio sources have to respond one way or another to `width`? - I don't consider `getSupportedConstraints()` a discovery mechanism. People already need to know what `width` is, what its syntax is, and what types it is meant for.

martinthomson: Right, my mistake.  So change the width example to something less obvious, like AEC.  An audio device that doesn't support echoCancellation will support only a value of false for that constraint.

jan-ivar: Right. I think what's confusing in this text is it talks about what "the User Agent supports", which is misleading, and makes it sound like we're talking about `getSupportedConstraints()`. We're not. This is about the Capabilities dictionary used in `track.getCapabilities()`, so this text should be all about what a specific *source* supports. That is the proper context that's not coming through clearly enough here IMHO.

So I think what might need clarification here is the word "supported" and what it relates to. We seem to shift between what the UA "supports" and what a source "supports".

One solution might be to avoid the word "support" entirely here, and just talk about what capabilities (or lack of capabilities) a device (aka source) has.

jan-ivar: The point in the text that the content of the dictionary may vary from browser to browser, is well taken, I just find it a bit confusing that there's no mention of the fact that the real purpose of this dictionary is to glean what individual sources can and can't do.

adam-be: Yes, the word "supported" in this text is not the same the "Supported" in ```getSupportedConstraints()``` (which perhaps should have been ```getKnownConstraintNames()```.

I agree that this text would benefit from talking more about what the specific source, that's backing up the track, supports. Someone willing to take a stab at it?

jan-ivar: I actually think `getSupportedConstraints` was named right, because IMHO it would be a mistake to return `true` for a constraint the browser didn't fully implement. `false` is the most informative to the web developer here, pointing (correctly) to the browser as the limitation, rather than any of their hardware devices (which is how I interpret a constraint failure).

adam-be: I find it a bit strange that the browser could detect that the hardware is capable enough, but the browser itself doesn't have the means to control the hardware features. Since it's the browser that's the messenger, I would assume that information about such hardware features could never reach the script.

jan-ivar: @adam-be I'm not following, could you rephrase? Are you saying there are no microphones with hardware aec that cannot be turned off, or something else?

adam-be: @jan-ivar, no. I was referring to your comment above about the ```getSupportedContraints``` method returning ```false``` when the browser is the limitation. Can you give an example when that would occur?

jan-ivar: Sorry, I meant `undefined`.

adam-be:
This one has been overtaken by #315

Issue #268, opened 2015-10-29
Iframe sandboxing options for gUM


martinthomson:
The fullscreen API has an option for sites to allow iframes to request fullscreen.  We should discuss having that sort of option and what the default would be (allow or deny).

n.b., fullscreen prevents use of the API in an iframe by default

martinthomson:
A simple option here would be to add an "allow-usermedia" tag to [the iframe sandbox attribute](http://www.w3.org/TR/html5/embedded-content-0.html#attr-iframe-sandbox).  That would make this available by default (less breakage), but make it possible to restrict what is possible in the sandbox.

stefhak:
I like that proposal, I think the ability for the site to disallow iframes to access camera and microphone is important.

alvestrand:
Would this mean that "allow-usermedia=no" would disallow iframes' access, while specifying "allow-usermedia=yes" or leaving it off would both allow it?

adam-be:
@alvestrand, from what I've seen it's only the presence of the token, no corresponding value, that activates the feature. But shouldn't it be "disallow-usermedia" to get the behaviour with default on?

stefhak:
@martinthomson any comment to the above?

martinthomson:
I don't know enough about how people use this to know whether to default to on or off.  I think that @adam-be is right about the label form, but the fact that there are no other labels of the form "disallow-*", we would need to be very careful about adding that.

adam-be:
I think the main question here is what the default should be. Then we can figure out how to enforce that; "allow-usermedia" or something else.

As @martinthomson wrote above, default on means less breakage. But it seems that most (read as all) iframe sandbox features are default off and the script opts in.

dontcallmedom:
there are two types of sandboxing:
* [`<iframe sandbox="allow-foo allow-bar">`](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-iframe-sandbox)
* [`<iframe allow-foo>`](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-iframe-allowfullscreen)

The first one has for effect that by default, you get an iframe that has many features cut-off, and get specific features re-enabled via the keywords `allow-foo` and `allow-bar` in my example. The currently recognized features are:  `allow-forms, allow-modals, allow-pointer-lock, allow-popups, allow-popups-to-escape-sandbox, allow-same-origin, allow-scripts, and allow-top-navigation`.

[...]

PR #313, opened 2016-02-04
Extend iframe with a new allowusermedia attribute (issue: #268)

adam-be:


stefhak:
@adam-be please ping the list and ask people to review.

getusermedia.html
```diff
@@ -2877,6 +2888,36 @@
       </div>
     </section>
     <section>
+      <h2>User Media in an IFrame</h2>
+      <p>There are some special circumstances when an [[!HTML51]]
+      <code><a href="https://www.w3.org/TR/2015/WD-html51-20150506/semantics.html#the-iframe-element">
+      iframe</a></code> wants to use user media. The iframe needs explicit
```
martinthomson: s/use/access/ ?  "use user" doesn't flow very nicely

getusermedia.html
```diff
@@ -2877,6 +2890,38 @@
       </div>
     </section>
     <section>
+      <h2>User Media in an IFrame</h2>
+      <div class="note">The topic of this section is under discussion and may
+      be subject to change.</div>
+      <p>There are some special circumstances when a cross-origin [[!HTML51]]
```
alvestrand: Delete "special". I think it'll be fairly common.

getusermedia.html
```diff
@@ -2877,6 +2890,38 @@
       </div>
     </section>
     <section>
+      <h2>User Media in an IFrame</h2>
+      <div class="note">The topic of this section is under discussion and may
+      be subject to change.</div>
```
alvestrand: Suggest alternate text: NOTE: A general approach to the subject of passing permissions to iframes is being discussed in other fora. The result of that may give other means of passing permissions to an iframe.

getusermedia.html
stefhak: I'd propose to go one step further. Don't talk about circumstances, just describe that the embedding page must explicitly permit for a cross-origin iframe to be able to request user media.

getusermedia.html
stefhak: Given Martin's comment about not being sure that 'require delegation' will fly we should perhaps stick with the original note.

adam-be: We can always update the note to reflect the ongoing discussion. I think the current one works for now.

adam-be:
Anyone opposed to merging this one and continue to work on it upstream?

alvestrand:
Let's merge it.

stefhak:
+1, let's merge this.

PR #324, opened 2016-03-08
Minor respec udpates and fixes

vivienlacourba:
Note that we won't be able to use automatic publication for this drafts as it is already past WD status (and [Echidna does not handle the more mature document statuses](https://github.com/w3c/echidna/wiki/FAQ#can-i-publish-a-fpwd-with-echidna-a-wg-note-a-rec)).


getusermedia.js
```diff
@@ -67,7 +67,7 @@ var respecConfig = {
    // This is important for Rec-track documents, do not copy a patent URI from a random
    // document unless you know what you're doing. If in doubt ask your friendly neighbourhood
    // Team Contact.
-   wgPatentURI:   ["http://www.w3.org/2004/01/pp-impl/47318/status","http://www.w3.org/2004/01/pp-impl/43696/status"],
+   wgPatentURI:   ["https://www.w3.org/2004/01/pp-impl/47318/status","https://www.w3.org/2004/01/pp-impl/43696/status"],
```
dontcallmedom: Are you sure Echidna will recognize these?

dontcallmedom:
if you can confirm that echidna will accept the `https` URL, this looks good to me.

While Echidna doesn't support automated publication for CR and later, we're not there yet, and there may well be one or several WDs before we go there.

getusermedia.js
vivienlacourba: @deniak tells me this works with Specberus see: https://github.com/w3c/specberus/blob/master/lib/rules/sotd/pp.js#L104

vivienlacourba:
@dontcallmedom as we already published a LC document @deniak told me we will not be able to automatically publish new WDs as there is an impact on IPP that is not handled today.

vivienlacourba:
per discussion with @dontcallmedom switching to Process 2015 (which is also a requirement to use automatic publication)

PR #326, opened 2016-03-11
Persist device IDs like they were cookies, 

### LINKED ISSUE #322: Browsing sessions and device IDs
We've had a fair bit of back-and-forth about this at Mozilla (I hope to have https://bugzilla.mozilla.org/show_bug.cgi?id=1223773 opened so people can see that discussion, but it's still locked down).

The definition of "browsing session", which was created after the term was used in gUM, is a poor fit for this use case.  In that bug, @jan-ivar was motivated to go to somewhat extraordinary efforts to be compliant with the definition we have (see aforementioned bug).  That would be OK, except that it didn't really address a security issue.

Architecturally speaking, sites need to have a consistent view of any persistent state that a browser retains for them.  This includes device identifiers, which are, for better or worse, a form of persistent identifier.  Requiring identifiers to be discarded at specific points in time, as the current text does, and on a different timeline to other persisted state, can harm that view unnecessarily.  There is no security or privacy advantage to be gained by throwing away this state, but site integrity can be adversely affected.

I suggest that "browser session" isn't a good fit for what we want to have for the lifetime of device identifiers.

We've been around this problem a few times, and I think that ultimately we will find it hard to agree on a single lifetime for these identifiers.  However, I think that we can and should set expectations.  I propose:

1. That the minimum time over which the identifiers persist, absent user override, is a browsing session.  That is, unless a user intervenes, device identifiers are constant for a given origin for the duration of a browsing session.

2. That browsers be permitted to persist those identifiers after that, but they not be required to.

This allows a browser to drop identifiers on shutdown (whatever that means), or to run a timer, or to watch for the last window open to an origin to close and nuke at that point.  It also allows for different policies during normal and "private" browsing.

Ideally, we would persist these identifiers just like other origin-specific state, but the cost of implementing that vs. the benefit was considered and I don't think that we wanted to invest that effort.  I think that we would if we all decided that was the right thing to do.  I think that it is, but consider the value to be dismally low.  Thus, I did not propose this as a third part to the above proposal, but would support that if others do.

martinthomson:
...don't destroy at the end of a browsing session.

Closes #322

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2396,12 +2396,8 @@
         <dd>
           <p>A unique identifier for the represented device.</p>
           <p>All enumerable devices have an identifier that MUST be unique to
-          the page's origin. As long as no local device has been attached to an
-          active MediaStreamTrack in a page from this origin, and no <a href=
-          "#stored-permissions">persistent permission</a> to access local
-          devices has been granted to this origin, then this identifier MUST
-          NOT be reusable after the end of the current browsing session.</p>
-          <p>However, if any local devices have been attached to an active
+          the page's origin.</p>
+          <p>If any local devices have been attached to an active
           MediaStreamTrack in a page from this origin, or <a href=
           "#stored-permissions">persistent permission</a> to access local
           devices has been granted to this origin, then this identifier MUST be


stefhak:
LGTM.

jan-ivar:
This PR falls short of its subject. Cookies have lots of privacy-protecting properties that seem absent in this spec, like visibility, tool/extension-support, expiry, third-party blocking policies, and last but not least, laws and regulations in some countries. As an implementer, reading that something should be "like cookies", begs a lot of questions about desired behavior.

It seems like we have two paths to choose from:

 1. Make `deviceId`s more like cookies (and address the things I mention above).
 2. Recognize that `deviceId`s are *not* cookies, and stop leaning on that similarity argument.

martinthomson:
Let's go through the list (is it exhaustive?):

* visibility - sure; UA problem
* tool/extension support - sure; again, UAs need to handle that, which could be addressed by storing them like cookies
* expiry - not relevant; maximum cookie lifetime is effectively forever
* third-party blocking policies - I'm not 100% on the details here, but I believe that this only applies to httpOnly cookies; if it doesn't, then (as implementers who enabled this unstandardized feature) we should do whatever DOM local storage, or indexedDB, or <raft of other stateful features>
* laws - I should say this, but I will; laws that identify cookies and assign them some sort of special status over all of the myriad other ways in which tracking might be enabled in a browser are daft

Yes, device identifiers are not cookies.  They are less capable than cookies.  They also have less baggage than cookies.

However, if you want to take on the crusade against origin-bound, persistent state in the browser, I'd suggest that this isn't the hill upon which to die.  Take them all on and work out how you might properly manage state.

jan-ivar:
I want to take path # 2, that's my point.

jan-ivar:
In other words, saying `deviceId`s are just like cookies - therefore persist away - glosses over these privacy-respecting differences. They're different beasts with different privacy trade-offs unfortunately.

alvestrand:
Speaking as contributor:
My opinion is that the rules around deviceIDs should ensure that they are no worse than cookies. If we can make them better than cookies, that is a Good Thing - but if an attacker can achieve exactly the same thing with cookies as with device IDs, I'm not going to bother defending against the same attack using device IDs; it doesn't improve the user's security.

That said, I think we should close this PR - the arguments presented on the list show why persisting them when unused is a bad idea.

stefhak:
I agree, deviceId is different from cookies.

PR #328, opened 2016-03-14
A stab at fixing deviceId perstisting (#322).

stefhak:
Alternative to #326

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2434,6 +2434,19 @@
           devices have the same group identifier if they belong to the same
           physical device; for example a monitor with a built-in camera and
           microphone.</p>
+          <p>This identifier MUST be unique to
+          the page's origin. As long as no local device has been attached to an
+          active MediaStreamTrack in a page from this origin, and no <a href=
+          "#stored-permissions">persistent permission</a> to access local
+          devices has been granted to this origin, then this identifier MUST
+          NOT be reusable after the end of the current browsing session.</p>
+          <p>However, if any local devices have been attached to an active
+          MediaStreamTrack in a page from this origin, or <a href=
+          "#stored-permissions">persistent permission</a> to access local
+          devices has been granted to this origin, then this identifier MUST be
+          persisted across browsing sessions.</p>
+
+
           <p class="fingerprint">Since <code>groupId</code> may persist across
           browsing sessions and to reduce its potential as a fingerprinting
           mechanism, <code>groupId</code> is to be treated as other persistent


getusermedia.html
```diff
@@ -3286,7 +3299,7 @@
                   ideal|/max(|actual|,|ideal|)
 </pre>
             </li>
-            <li>For all string and enum non-required constraints (sourceId,
+            <li>For all string and enum non-required constraints (deviceId,
```
martinthomson: Unrelated change.

getusermedia.html
```diff
@@ -2434,6 +2434,19 @@
           devices have the same group identifier if they belong to the same
           physical device; for example a monitor with a built-in camera and
           microphone.</p>
+          <p>This identifier MUST be unique to
+          the page's origin. As long as no local device has been attached to an
+          active MediaStreamTrack in a page from this origin, and no <a href=
+          "#stored-permissions">persistent permission</a> to access local
+          devices has been granted to this origin, then this identifier MUST
+          NOT be reusable after the end of the current browsing session.</p>
```
martinthomson: This relies on the definition of browsing session still, and includes the offensive MUST clear requirement. I do not understand how this helps.

stefhak: It makes us treat groupId in the same way as deviceId (in the current spec). 
If we decide to not rely on browsing session this PR makes no sense.

getusermedia.html
stefhak: Yes, unrelated.

getusermedia.html
jan-ivar: `groupId` was never persisted before AFAIK, and I don't understand why we need to persist it now. In my mind, it exists solely as a grouping construct. That is, it merely tells you whether two devices belong to the same hardware or not.

In the rare case that someone needs to remember a group, they'd merely need to persist a device in the group.

jan-ivar: This is explained in http://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-enumeratedevices step 2. 3. 3.

stefhak: > groupId was never persisted before AFAIK
If we never persist groupId this PR makes no sense. But it is said in this section of the spec that 
```
Since groupId may persist across browsing sessions and to reduce its potential as a fingerprinting mechanism, groupId is to be treated as other persistent storage mechanisms such as cookies...
```
and further in the Privacy and Security section
```
It recommends to treat the per-origin persistent identifier deviceId and groupId as other persistent storages (e.g. cookies) are treated.
```

stefhak: > This is explained in http://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-enumeratedevices step 2. 3. 3.

If you read 2.2 and 2.3.1 it is pretty clear the ```groupId``` is persisted.

jan-ivar: 2.2 says (emphasis mine) "If this method has been called previously ***within this browsing session***, let oldList be the list of MediaDeviceInfo objects that was produced at that call (resultList); otherwise, let oldList be an empty list.".

That means `oldList` is an empty list across across browsing sessions, i.e. not persisted.

jan-ivar: > Since groupId may persist across browsing sessions and to reduce its potential as a fingerprinting mechanism, groupId is to be treated as other persistent storage mechanisms such as cookies...

Hmm, it does say that quite explicitly, yet it contradicts 2.2. I wonder when this text was added...

stefhak: @jan-ivar to me it seems that never persisting groupId between browsing sessions would be the best solution (and then make that clear and update the text regarding fingerprinting rather than doing the change this PR proposes)

jan-ivar: @stefhak I agree. The "persist groupId" text was added less than 5 months ago in https://github.com/w3c/mediacapture-main/commit/87b835866ea45682b5294f4393dc0024c9683e00 by @dontcallmedom .

The name of that commit is *"Specify that groupId is to be cleared when cookies are"*, which seems to assume `groupId` was already persisted, when it was not, so this looks like an editorial mistake.

jan-ivar: I'm still hunting down the Privacy section...

jan-ivar: The privacy section was added around the same time in 3f260fe9d744bc6c5ea2f8c96442bda5dcdc070a , so I'd chalk this up to a misunderstanding on @dontcallmedom 's part.

stefhak:
@jan-ivar thanks for digging in the history. It is pretty clear this PR is not the right thing to do, so closing.

PR #330, opened 2016-03-14
Clarify that group ID is non-persistent

stefhak:
as discussed in #328

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1217,12 +1217,12 @@
               <td><dfn>groupId</dfn></td>
               <td>DOMString</td>
               <td>
-                The origin and browsing session unique group identifier for the source of the
-                <a>MediaStreamTrack</a>. Two devices have the same group
-                identifier if they belong to the same physical device; for
-                example, the audio input and output devices representing the
-                speaker and microphone of the same headset would have the same
-                groupId.
+                The origin and browsing session unique group identifier for
+                the source of the <a>MediaStreamTrack</a>. Two devices have
+                the same group identifier if they belong to the same physical
+                device; for example, the audio input and output devices
+                representing the speaker and microphone of the same headset
+                would have the same groupId.
               </td>
             </tr>
           </tbody>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1205,12 +1205,13 @@
                 different for other origins. Some sort of GUID is recommended
                 for the identifier. Note that the setting of this property is
                 uniquely determined by the source that is attached to the
-                <a>MediaStreamTrack</a>. In particular, <a>getCapabilities()</a> will return only
-                a single value for deviceId. This property can therefore be
-                used for initial media selection with <a>getUserMedia()</a>.
-                However, it is not useful for subsequent media control with
-                <a>applyConstraints()</a>, since any attempt to set a different
-                value will result in an unsatisfiable <a>ConstraintSet</a>.
+                <a>MediaStreamTrack</a>. In particular,
+                <a>getCapabilities()</a> will return only a single value for
+                deviceId. This property can therefore be used for initial media
+                selection with <a>getUserMedia()</a>. However, it is not useful
+                for subsequent media control with <a>applyConstraints()</a>,
+                since any attempt to set a different value will result in an
+                unsatisfiable <a>ConstraintSet</a>.
               </td>
             </tr>
             <tr id="def-constraint-groupId">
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -1217,12 +1217,12 @@
               <td><dfn>groupId</dfn></td>
               <td>DOMString</td>
               <td>
-                The origin-unique group identifier for the source of the
-                <a>MediaStreamTrack</a>. Two devices have the same group
-                identifier if they belong to the same physical device; for
-                example, the audio input and output devices representing the
-                speaker and microphone of the same headset would have the same
-                groupId.
+                The origin and browsing session unique group identifier for
```
jan-ivar: I think we can remove mention of origin here since browsing sessions are inherently per-origin.

stefhak: I think you're right

stefhak:
I changed the title from "Make groupID non-persistent" to "Clarify that group ID is non-persistent" since I think this better reflects what the PR does. The core algorithm in section 9.2.2 is unchanged in this respect.
What this PR mainly does is to remove some text at other places that was quite recently added that talks about the fact that groupId is persisted (which it according to 9.2.2 never has been) is a problem from a privacy/fingerprinting perspective, and clarifies that groupId is not persisted.

burnburn:
This LGTM.  I believe we finally are capturing the following:
- that deviceIds must be persisted across browsing sessions when device permission has been given
- that they are removed in the same way as other persistent storage (such as cookies)
- that groupIds are never persisted

getusermedia.html
```diff
@@ -1217,12 +1217,20 @@
...
-                groupId.
+                The browsing session-unique group identifier for
+                the source of the <a>MediaStreamTrack</a>. Two devices have
+                the same group identifier if they belong to the same physical
+                device; for example, the audio input and output devices
+                representing the speaker and microphone of the same headset
+                would have the same groupId. Note that the setting of this property is
+                uniquely determined by the source that is attached to the
+                Track. In particular, <a>getCapabilities()</a> will return only
+                a single value for groupId. Since this property is not stable
```
martinthomson: <code>groupId</code> (twice above)

getusermedia.html
```diff
@@ -1217,12 +1217,20 @@
...
+                between browsing sessions its usefulness for initial media
+                selection with <a>getUserMedia()</a> is limited.
+                It is not useful for subsequent media control with
+                <a>applyConstraints()</a>, since any attempt to set a different
+                value will result in an unsatisfiable <a>ConstraintSet</a>
```
martinthomson: missing period

getusermedia.html
```diff
@@ -1217,12 +1217,20 @@
...
-                speaker and microphone of the same headset would have the same
-                groupId.
+                The browsing session-unique group identifier for
+                the source of the <a>MediaStreamTrack</a>. Two devices have
+                the same group identifier if they belong to the same physical
+                device; for example, the audio input and output devices
+                representing the speaker and microphone of the same headset
+                would have the same groupId. Note that the setting of this property is
+                uniquely determined by the source that is attached to the
+                Track. In particular, <a>getCapabilities()</a> will return only
```
martinthomson: s/Track/<code>MediaStreamTrack</code>/

getusermedia.html
stefhak: I have followed the same model as is in the current document (and how it is also done in the previous section for deviceId). @burnburn @adam-be is this something we should change?

getusermedia.html
stefhak: thanks for spotting.

getusermedia.html
stefhak: This was present already before, but I think you proposal makes sense and I will update the PR.

stefhak:
Updated considering Martin's comments. PTAL.

alvestrand:
LGTM

elagerway:
LGTM

adam-be:
LGTM

getusermedia.html
```diff
@@ -1205,24 +1205,34 @@
...
+                would have the same groupId. Note that the setting of this
+                property is uniquely determined by the source that is attached
+                to the <a>MediaStreamTrack</a>. In particular,
+                <a>getCapabilities()</a> will return only a single value for
+                groupId. Since this property is not stable between browsing
+                sessions its usefulness for initial media selection with
+                <a>getUserMedia()</a> is limited. It is not useful for
+                subsequent media control with <a>applyConstraints()</a>,
+                since any attempt to set a different value will result in an
+                unsatisfiable <a>ConstraintSet</a>.
```
jan-ivar: More specifically, an `OverconstrainedError`?

PR #340, opened 2016-04-18
Restrict emission of devicechange event to apps with list-devices perm

### LINKED ISSUE #333: Devicechange event firing potentially being a fingerprint
This is part of Nick Doty's feedback after reviewing how his comments made during LC have been addressed. What Nick says:
"I raised this issue in my initial set of comments, and while there was some discussion with Harald, I'm not sure it ever got documented as an issue or had any resolution. I provided a little more detail in my response here: https://lists.w3.org/Archives/Public/public-privacy/2015OctDec/0028.html

My concern is that firing a devicechange event simultaneously in different browsing contexts (including tabs or iframes not in the foreground, or in different browsers altogether, that have not asked for any permissions) creates a risk of unexpected correlation of browsing activity."

dontcallmedom:
My interpretation of what the discussion of #333 seems to lead to.

close #333

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3989,7 +3989,7 @@ below dictionary, but instead provide its own definition. See
     cases, the labels are persistent across browsing sessions and across
     origins that have also been granted authorization, and thus potentially
     provide a way to track a given device across time and origins.</p>
-    <p>When permission has been granted, the <code><a>devicechange</a></code>
+    <p>For origins to which permission has been granted, the <code><a>devicechange</a></code>
     event will be emitted across browsing contexts and origins each time a new
     media device is added or removed; user agents can mitigate the risk of
     correlation of browsing activity across origins by fuzzing the timing of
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM.

stefhak:
@dontcallmedom perhaps you could add something about fuzzing the eventfiring as @npdoty proposes in #333 ?

getusermedia.html
```diff
@@ -3984,6 +3989,11 @@
     cases, the labels are persistent across browsing sessions and across
     origins that have also been granted authorization, and thus potentially
     provide a way to track a given device across time and origins.</p>
+    <p>When permission has been granted, the <code><a>devicechange</a></code>
+    event will be emitted across browsing contexts and origins each time a new
```
stefhak: Should it not say something like "For origins to which permission has been granted" rather than "When permission has been granted"? The proposed text leads me to think that as soon as one origin has been granted permission the event will be emitted to all origins (regardless of have permission or not).

PR #341, opened 2016-04-20
Make ended event a simple event

### LINKED ISSUE #338: Make ended event a simple event
Per discussion in #337 the ErrorEvent is a problem in that it can't be implemented as specified, due to a name clash with HTML.

The only user of ErrrorEvent is to provide an optional error on the "ended" event; so far, nobody's suggested text that says when we would include such an error.

Suggestion: Make the "ended" event a simple event, and remove ErrorEvent from the spec.
There are no known implementations at the moment.

dontcallmedom:
close #338

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2177,7 +2177,7 @@
         </tr>
         <tr>
           <td><code id="event-mediastreamtrack-ended">ended</code></td>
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM.

burnburn:
LGTM

PR #343, opened 2016-04-20
Various fixes to permission language

### LINKED ISSUE #334: Is more needed regarding revocation?
This is part of Nick Doty's feedback after reviewing how his comments during LC have been addressed. What Nick says:
"I believe there were open questions on both user-side revocation and site-side revocation of persisted permissions.

On user-side revocation, Rigo had noted that RFC 7478 mandated user agents provide the capability for users to revoke permissions, and we weren't sure that had been translated into draft-ietf-rtcweb-security-arch. I think I may have dropped the ball on not creating a pull request on that point; if it's still useful for me to do so, let me know. The Media Capture spec assumes that user-side revocation is required, though I don't think it introduces any specific normative requirement.
	https://lists.w3.org/Archives/Public/public-media-capture/2015Oct/0061.html

I raised the concern that sites should also have a way to revoke persisted permissions that they may have received, as one way of limiting the risk to their users where they requested camera access in a way that might have just been one-time and subsequently had a security breach of some kind (like a reflected XSS attack). The wide review document suggests this is resolved via the Permissions API:

> We reworked our permission system to be based on the Permission API, where revokation is addressed

However, I couldn't find any references to the Permissions API in the Media Capture and Streams document. Is there any requirement or expectation that user agents that implement the Media Capture spec will also implement the corresponding Permissions API functionality? Or an example for how sites can use the Permissions API to query or revoke permissions using that API?

The current editor's draft of the Permissions API does include a PermissionsDescriptor for camera and microphone, and a method for revoking permissions. The Permissions API is at least under development for both Chrome and Firefox:
https://platform-status.mozilla.org/#permissions
Although I don't think the current Firefox patch includes camera and microphone among the supported permission names:
https://bugzilla.mozilla.org/show_bug.cgi?id=1105827

To the extent that it's unclear whether sites will be able to revoke their Media Capture permissions using the Permissions API, I remain concerned about this point. That said, maybe the fact that the Permissions API editor's draft has included it is a promising sign and the Privacy Interest Group could provide feedback on the Permissions API regarding this point."

dontcallmedom:
close #334

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -182,8 +182,7 @@
         <p>
           The terms <dfn>permission</dfn>, <dfn>retrieve the permission
             state</dfn>, <dfn id="request-permission">request
-            permission</dfn>, <dfn id="delete-permission">delete a
-            permission storage entry</dfn> and <dfn id="create-permission">
+            permission</dfn> and <dfn id="create-permission">
             create a permission storage entry</dfn> are
           defined in [[!permissions]].
           </p>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2422,12 +2422,12 @@
           <p>All enumerable devices have an identifier that MUST be unique to
           the page's origin. As long as no local device has been attached to a
           live MediaStreamTrack in a page from this origin, and no <a href=
-          "#stored-permissions">persistent permission</a> to access local
+          "#stored-permissions">stored permission</a> to access local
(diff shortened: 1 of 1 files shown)


alvestrand:
LGTM.
One outstanding item is what to do about open devices at permission revocation time; that issue needs raising separately.

PR #344, opened 2016-04-20
Add [CEReactions] annotation to allowUserMedia

domenic:
Since the allowUserMedia attribute can change the DOM in ways that may impact custom elements, it needs to be annotated with the [CEReactions] attribute. This is part of the larger effort at https://github.com/w3c/webcomponents/issues/186. [CEReactions] is defined in https://html.spec.whatwg.org/multipage/scripting.html#cereactions.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2949,7 +2949,7 @@
       <p>The iframe DOM interface is extended as described by the partial
       interface below.</p>
       <dl class="idl" title="partial interface HTMLIFrameElement">
-        <dt>attribute boolean allowUserMedia</dt>
+        <dt>[CEReactions] attribute boolean allowUserMedia</dt>
         <dd>
           <p>The allowUserMedia IDL attribute MUST [[!HTML51]] <code><a href=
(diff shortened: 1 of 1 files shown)


alvestrand:
Is there a corresponding definition of [CEReactions] in HTML 5.1?

dontcallmedom:
(the Travis CI failure comes from the current master being broken :(, not this specific PR)

dontcallmedom:
If we are to look for a W3C reference for the [CEReactions] extended attribute, the more logical place would be in WebIDL IMO.

That said, given that we are about to go to CR and this extended attribute is brand new, I think we should look at this during CR, not before.

domenic:
This is something implementers need to worry about now, not later. If the CR version of this spec is not going to be maintained, then at least the ED should be.

alvestrand:
Why do implementors need to worry about it now when they didn't have to worry about it two weeks ago?
The spec will be maintained.

alvestrand:
Closing until "later" comes around.

domenic:
Implementers need to worry about this now, as they are actively implementing custom elements. This is serious; please do not close. I am prepared to formally object.

Issue #339, opened 2016-04-14
originIdentifier is unused


jan-ivar:
With the move of *request permission* to the permissions API, the variable `originIdentifier` is unused in the [getUserMedia algorithm](http://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-getusermedia).

jan-ivar:
Also unsure why we're adding a normative dependency to the permissions API. Does that mean the permissions API is required to implement in order to implement getUserMedia?

alvestrand:
It gave us a language to talk about our permissions model without inventing new stuff.
The integration was a response to a "how do we relate" question.

alvestrand:
@martinthomson I think this language was proposed by you. Comments?

martinthomson:
I didn't write the text, and it's the first I've seen it.  It seems like we need to define this in terms of a [permissions storage identifier](https://w3c.github.io/permissions/#permission-storage-identifier).  Most likely by placing the top-level context in the optional bit.

dontcallmedom:
@adam-be was the one who added the text in 2515dc8ffd56bd179189282fa9f872d2d1641d3e fwiw

alvestrand:
Still 9 references that are not specific. Dom to take action to list these.

dontcallmedom:
@alvestrand 's comment was meant for https://github.com/w3c/webrtc-pc/issues/337

alvestrand:
Closed by #349

PR #349, opened 2016-04-25
Use 'originIdentifier' in getUserMedia() steps (issue #339)

adam-be:
Fix for issue #339

First commit is purely editorial so please review the second.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2797,11 +2797,10 @@
                   browsing context</a>'s origin.</p>
                 </li>
                 <li>
-		  <p><a>Request permission</a> for use of the devices,
-                  while considering
-		  all devices attached to a <code>live</code> MediaStreamTrack
-		  to have permission status "granted", resulting in a set of
-		  provided media.</p>
+                  <p><a>Request permission</a> for use of the devices, while
+                  considering all devices attached to a <code>live</code>
+                  MediaStreamTrack to have permission status "granted",
+                  resulting in a set of provided media.</p>
                   <p>The provided media MUST include precisely one track of
                   each media type in requestedMediaTypes from the
                   <var>finalSet</var>. The decision of which devices to choose
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2797,7 +2797,8 @@
                   browsing context</a>'s origin.</p>
                 </li>
                 <li>
-                  <p><a>Request permission</a> for use of the devices, while
+                  <p>For the origin identified by <var>originIdentifier</var>,
+                  <a>request permission</a> for use of the devices, while
                   considering all devices attached to a <code>live</code>
                   MediaStreamTrack to have permission status "granted",
                   resulting in a set of provided media.</p>


getusermedia.html
```diff
@@ -2797,11 +2797,11 @@
                   browsing context</a>'s origin.</p>
                 </li>
                 <li>
-		  <p><a>Request permission</a> for use of the devices,
-                  while considering
-		  all devices attached to a <code>live</code> MediaStreamTrack
-		  to have permission status "granted", resulting in a set of
-		  provided media.</p>
+                  <p>For the origin identified by <var>originIdentifier</var>,
+                  <a>request permission</a> for use of the devices, while
```
jan-ivar: I don't see how we're expected to verify anything about a "request permission" algorithm that doesn't exist yet.

dontcallmedom: https://w3c.github.io/permissions/#request-a-permission ?

getusermedia.html
```diff
@@ -2797,11 +2797,11 @@
...
+                  considering all devices attached to a <code>live</code>
+                  MediaStreamTrack to have permission status "granted",
+                  resulting in a set of provided media.</p>
```
jan-ivar: I also don't know when this language snuck in, but it doesn't match how Firefox works. The existence of live tracks has no bearing on whether users get prompted by `getUserMedia` in Firefox. You get prompted for each track you request, and that's it.

What was the point of introducing this language? Was there any discussion?

getusermedia.html
jan-ivar: Thanks, it's not in the TR which is what we link to, and we still have a [disclaimer](https://rawgit.com/adam-be/mediacapture-main/use-origin-identifier/getusermedia.html#request-permission) in our doc saying it doesn't exist. The link then seems like it should be ***request a permission***, not ***request permission***.

Also, maybe everyone else is aware, but this surprised me, that the other spec seems to follow a slightly different process for consensus than at least I was used to. Specifically, the editor of the other spec [says](https://github.com/w3c/permissions/issues/85#issuecomment-213616988) he doesn't "claim that the current text represents any sort of consensus."

I'm surprised we're adding such dependencies when we're supposed to be way past last call.

jan-ivar: Is there a way in this spec now for `getUserMedia` to succeed without granting persistent permission? What does ***request a permission*** return in that case?

See https://github.com/w3c/mediacapture-main/issues/350 for my take down of the *request permission* algorithm.

getusermedia.html
jan-ivar: Also, doesn't this literally say site B must be granted access if site A currently has a live MediaStreamTrack?

adam-be:
@jan-ivar, my interpretation is that your concerns are mainly with text that's not added by this PR (that you've also created a new issue for). And that shouldn't block us from merging this PR. Is that correct?

jan-ivar:
@adam-be Sounds like the algorithm referenced should be [request ***a*** permission](https://w3c.github.io/permissions/#request-a-permission), though I see no evidence that it takes an `origin` (though the document does talk about origin elsewhere now).

The level of meaninglessness seems neither increased nor diminished by this PR, so I have no problem with it merging.

PR #358, opened 2016-05-06
Getusermedia: Pull the permissions check into the per-type loop

### LINKED ISSUE #352: With single camera, calling getUserMedia twice causes PermissionFailure.
A [commit](https://github.com/w3c/mediacapture-main/commit/049fd00f87f9ed39a642826fb3ae521cf0e49bb2) from a little over a month ago broke the [`getUserMedia`](http://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-getusermedia) algorithm here (step 6.3):

> [Retrieve the permission state](http://w3c.github.io/mediacapture-main/getusermedia.html#dfn-retrieve-the-permission-state) for all candidate devices that are not attached to a `live` MediaStreamTrack in the current browsing context. Remove from the set any device for which the permission state is "denied".
> 
> If the set is now empty, jump to the step labeled *Permission Failure* below.

For users with only one camera, the `set` will always be empty here if a page already holds an active track when it calls `getUserMedia`, resulting in permission failure.

alvestrand:
Closes #352

Diff also moves the "overconstrained" call-out, since the loop can now exit either with "overconstrained" or with "permission denied".

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2711,36 +2711,42 @@
                       satisfy the constraints, by verifying that at least one
                       settings dictionary exists that satisfies the
                       constraints.</p>
+                      <p>If <var>finalSet</var> is the empty set, let
+			<var>constraint</var> be any required constraint whose
+			fitness distance was infinity for all settings
+			dictionaries examined while executing the
+			<code><a>SelectSettings</a></code> algorithm, let
+			<var>message</var> be either <code>undefined</code>
+			or an informative human-readable message, and reject
+			<var>p</var> with a new
+			<code><a>OverconstrainedError</a></code> created
+			by calling 
+			<code>OverconstrainedError(<var>constraint</var>,
+			  <var>message</var>)</code>, then abort these steps.
+		      </p>
+                      <p class="fingerprint">This error gives information about
+			what the underlying device is not capable of producing,
+			before the user has given any authorization to any
+			device,
+			and can thus be used as a fingerprinting surface.
+		      </p>
                     </li>
+		    <li>
+		    <p><a>Retrieve the permission state</a> for all
+		      candidate devices that are not attached to a
+		      <code>live</code>
+		      MediaStreamTrack in the current
+                      browsing context. Remove from <var>finalSet</var> any
+		      device for which the permission state is "denied".
+		    </p>
+		    <p>
+		      If <var>finalSet</var> is now empty, indicating that
+		      all devices of this type are in state "denied", jump
+		      to the step labeled <em>PermissionFailure</em> below.
+		    </p>
+		    </li>
                   </ol>
-                  <p>If <var>finalSet</var> is the empty set, let
-                  <var>constraint</var> be any required constraint whose
-                  fitness distance was infinity for all settings dictionaries
-                  examined while executing the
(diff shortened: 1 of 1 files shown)


alvestrand:
@jan-ivar please review

getusermedia.html
```diff
@@ -2711,37 +2711,43 @@
                       satisfy the constraints, by verifying that at least one
                       settings dictionary exists that satisfies the
                       constraints.</p>
+                      <p>If <var>finalSet</var> is the empty set, let
```
jan-ivar: *finalSet* is defined outside the for-loop, and will never be empty on subsequent runs of the loop, so this wont work.

Inside the loop I think we need to operate on *candidateSet* instead, the media type *T* subset. We could change the paragraph ahead of this one from:

* If the algorithm does not return `undefined`, add the track to *finalSet*

to:
* If the algorithm returns <code>undefined</code>, remove the track from <var>candidateSet</var>

Then we can safely say "If *candidateSet* is empty..." (and then remember to move tracks from *candidateSet* to *finalSet* at the end).

getusermedia.html
```diff
@@ -2711,37 +2711,43 @@
...
+                      <p class="fingerprint">This error gives information about
+			what the underlying device is not capable of producing,
+			before the user has given any authorization to any
+			device,
+			and can thus be used as a fingerprinting surface.
+		      </p>
                     </li>
+		    <li>
+		    <p><a>Retrieve the permission state</a> for all
+		      candidate devices that are not attached to a
```
jan-ivar: "all candidate devices" sounds like all of them, not just the subset of sources that produce media type T. Works but seems inefficient.

getusermedia.html
```diff
@@ -2711,37 +2711,43 @@
...
+		      <code>live</code>
+		      MediaStreamTrack in the current
+                      browsing context. Remove from <var>finalSet</var> any
+		      device for which the permission state is "denied".
```
jan-ivar: `finalSet` and `candidateSet` contain tracks, not devices.

Maybe: 'Remove from `candidateSet` all tracks from devices for which the permission state is "denied".'

getusermedia.html
```diff
@@ -2711,37 +2711,43 @@
...
+		    </p>
+		    <p>
+		      If <var>finalSet</var> is now empty, indicating that
```
jan-ivar: With *candidateSet* this should work.

getusermedia.html
```diff
@@ -2711,37 +2711,43 @@
...
+		      all devices of this type are in state "denied", jump
+		      to the step labeled <em>PermissionFailure</em> below.
+		    </p>
```
jan-ivar: And then at the end of the loop: "Add all tracks in *candidateSet* to *finalSet*.

alvestrand: You're showing that "finalSet" is used in two different meanings - once as "all the candidates for this type" and once for "all candidates of any type". Thanks - I'll update the PR.
(if they were the same, checking if finalSet was empty for the second type checked would give the wrong result.)

alvestrand:
Hope it looks better now.

elagerway:
lgtm

Issue #361, opened 2016-05-26
Example image text is hard to read


nigelmegitt:
The white text on coloured boxes in the diagrams in §5 is hard to read. Please could you check the colour contrast and text size for accessibility?

For example, the second diagram, boxes next to B and Y are hard to read (small fuzzy text); the white text on a yellow callout box in the third diagram is hard to read (low contrast). Both problems are significantly present in the 5th and 6th diagrams, in some cases both in the same box.

In some cases black on white text is very small too.

This is not a full accessibility review, just a couple of problems that I spotted quickly and hopefully can easily be resolved.

elagerway:
PR Merged

nigelmegitt:
Thanks, that's much clearer!

PR #365, opened 2016-06-09
Replace some PNG images with SVGs (issue #361)

adam-be:
I got the original silde set from @burnburn and converted those to SVGs. These should be easier to modify without having to export the result.

Proposed fix for issue #361


adam-be:
Rendered preview: http://rawgit.com/adam-be/mediacapture-main/svg-images-2/getusermedia.html#the-model-sources-sinks-constraints-and-settings

elagerway:
lgtm

Issue #359, opened 2016-05-11
MUST clear requirement for deviceId


martinthomson:
I apologize for not properly following the giant cluster of issues around  #322, #326, #328, and #330, but I don't believe that the main reason I opened #322 was properly addressed.

The offending text is this:

>  All enumerable devices have an identifier that MUST be unique to the page's origin. As long as no local device has been attached to a live MediaStreamTrack in a page from this origin, and no stored permission to access local devices has been granted to this origin, then **this identifier MUST NOT be reusable after the end of the current browsing session**.

As noted in #322, this is not what browsers currently do.  Nor is it what I believe that they should do.  They should do [what Harald says Chrome does](https://github.com/w3c/mediacapture-main/issues/322#issuecomment-193895840).  The other parts of the text are fine, but this isn't right.

I would be OK with "the user agent MAY generate a new identifier for every new realm created for the origin, see [[section with better details]]."

stefhak:
I may have gotten this totally wrong, but isn't the reason for mandating that deviceId is not reusable to avoid it being used for fingerprinting?
@martinthomson what precisely is the problem with the current text?

martinthomson:
The current text requires that the deviceId is cleared at a very precise time.  See the definition of browsing session.

The requirement to clear at that precise moment is a) a bad idea (see #322) and b) not what anyone does in practice.  @jan-ivar just reminded me of this, asking if I would object to him landing a change that made Firefox compliant.  That change is a lot of fairly complex code and I think we'd be better off taking Harald's view on this: treat it just like a cookie.

stefhak:
@jan-ivar also pointed out in #322 that the current spec text is a result addressing input given during the last call. This was settled June/July last year, and I think it has been in the spec since.
That the spec has said has said for a long time does not mean it is right, but personally I'm still a bit confused on why we would need to change (especially if it means we open up a difficult privacy discussion again).

martinthomson:
The text was added before the definition of browsing session was clear.

stefhak:
Ok, that's a change. I don't have a strong personal opinion, but it feels awkward to at this stage make changes that could open up a privacy discussion we've once closed unless we have very good reasons to do so.
I hope for more input (also from others) here.

stefhak:
I just did a small test with Chrome Canary, by using the developer console. It seems that Canary *does* generate new deviceIds if the tab is closed and a new tab on the same origin is opened.

alvestrand:
[...]

PR #373, opened 2016-07-08
Add pre-grant deviceId alternative.

jan-ivar:
Tries to capture the alternative we discussed at the last virtual face-to-face.

With all this info, I reordered the text a bit to cover uniqueness, stability, pre-grant behavior, and cookies in that order. I also:

 * Tried to clarify that ids persist after allowing gum just once.
 * Added visible cookie requirement to the UA alternative.
 * Tried to fix clearing logic to not orphan same-origin tabs (clear when last origin tab closes).
 * Restrict persistence for sites blocked from using cookies.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3074,7 +3074,7 @@ interface MediaDeviceInfo {
               local devices has been granted to this origin, then this
               identifier MUST be persisted. Unique and stable identifiers let
               the application save, identify the availability of, and directly
-              request specific sources from previous visits.</p>
+              request specific sources, across multiple visits.</p>
               <p>However, as long as no local device has been attached to a live
               MediaStreamTrack in a page from this origin, and no
               <a href="#stored-permissions">stored permission</a> to access
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3065,29 +3065,32 @@ interface MediaDeviceInfo {
             <dd>
               <p>A unique identifier for the represented device.</p>
               <p>All enumerable devices have an identifier that MUST be unique
-              to the page's origin. As long as no local device has been
-              attached to a live MediaStreamTrack in a page from this origin,
-              and no <a href="#stored-permissions">stored permission</a> to
-              access local devices has been granted to this origin, then this
-              identifier MUST NOT be reusable after the end of the current
-              browsing session.</p>
-              <p>However, if any local devices have been attached to a live
-              MediaStreamTrack in a page from this origin, or <a href=
-              "#stored-permissions">stored permission</a> to access local
-              devices has been granted to this origin, then this identifier
-              MUST be persisted across browsing sessions.</p>
-              <p>Unique and stable identifiers let the application save,
-              identify the availability of, and directly request specific
-              sources.</p>
-              <p>This identifier MUST be un-guessable by applications of other
-              origins to prevent the identifier being used to correlate the
-              same user across different origins.</p>
+              to the page's origin. This identifier MUST be un-guessable by
+              applications of other origins to prevent the identifier from being
+              used to correlate the same user across different origins.</p>
+              <p>If any local devices have been attached to a live
+              MediaStreamTrack in a page from this origin, or
+              <a href="#stored-permissions">stored permission</a> to access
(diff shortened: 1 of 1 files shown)


jan-ivar:
[Preview link](https://rawgit.com/jan-ivar/mediacapture-main/46a39f8219dfd9e29f5ca6cea44a8d0f5882c43b/getusermedia.html#dom-mediadeviceinfo-deviceid).

jan-ivar:
[Preview link](https://rawgit.com/jan-ivar/mediacapture-main/24b0ae23ee382269c0ff8045542f9e72d3b9d410/getusermedia.html#dom-mediadeviceinfo-deviceid).

jan-ivar:
For issue https://github.com/w3c/mediacapture-main/issues/359.

stefhak:
This mostly works for me. I wonder why you use plural in "...when other persistent storages are cleared."? And should "for this origin" be appended (if you clear all cookies stored for site A the deviceIds for site B should survive, right?)?

At the call Martin also brought up the question of what would happen if the user clears cookies for an origin while there is an active browsing session for that origin. Would a subsequent enumerateDevices() give other deviceIds if done in the same tab? What if a new tab is opened for the same origin, and enumerateDevices()  is called there? Do we need to spec that out, or can it be left to implementations?

jan-ivar:
> I wonder why you use plural in "...when other persistent storages are cleared."?

I didn't change that part.

>  what would happen if the user clears cookies for an origin while there is an active browsing session for that origin

Is how sites respond to clearing cookies spec'ed anywhere? What happens in Firefox is that if you clear cookies, then you get a fresh deviceId next time, even in existing tabs. Having anything survive a cookie clearing would be seen as a security bug IMHO. It was even one of your tests, wasn't it? There are other things like private browsing we don't spec at all, so I wouldn't we need to say anything more.

alvestrand:
@guidou can you take a look?

getusermedia.html
```diff
@@ -3065,29 +3065,32 @@
...
+              MediaStreamTrack in a page from this origin, and no
+              <a href="#stored-permissions">stored permission</a> to access
+              local devices has been granted to this origin, then the user agent
+              MAY clear this identifier once the last browsing session from this
+              origin has been closed.  If the user agent chooses not to clear the
+              identifier in this condition, then it MUST provide for the user to
+              visibly inspect and delete the identifier, like a cookie.</p>
               <p class="fingerprint">Since <code>deviceId</code> may persist
               across browsing sessions and to reduce its potential as a
               fingerprinting mechanism, <code>deviceId</code> is to be treated
```
martinthomson: s/is to be/MUST be/

getusermedia.html
```diff
@@ -3065,29 +3065,32 @@
...
-              origins to prevent the identifier being used to correlate the
-              same user across different origins.</p>
+              to the page's origin. This identifier MUST be un-guessable by
+              applications of other origins to prevent the identifier from being
+              used to correlate the same user across different origins.</p>
+              <p>If any local devices have been attached to a live
+              MediaStreamTrack in a page from this origin, or
+              <a href="#stored-permissions">stored permission</a> to access
+              local devices has been granted to this origin, then this
+              identifier MUST be persisted. Unique and stable identifiers let
```
martinthomson: MUST be persisted needs qualification, since it has a competing MUST below (well, if you accept my tweak to that language, that is).

martinthomson:
LGTM, modulo some minor comments.

guidou:
lgtm

stefhak:
Based on the comments so far it looks like we can merge this one once Martin's comments have been addressed (and Travis is happy).

getusermedia.html
jan-ivar: @martinthomson Hmm, "deviceId MUST be treated as other persistent storage mechanisms such as cookies" reads like a hard requirement of a lose definition, which seems odd, and I worry could be misconstrued as a hard requirement to implement exactly like cookies.

Rather this sentence seems like it was meant as a lead-in to the following sentence which has the specifics (two MUSTs), so how about instead tying them together with ", in that"?

Issue #375, opened 2016-07-22
Should devicechange fire once or twice when I insert a USB cam with mic?


jan-ivar:
The spec appears conflicted on this:

Under [devicechange](https://w3c.github.io/mediacapture-main/getusermedia.html#event-mediadevices-devicechange) it says: *"[Fired when] the set of media devices, available to the User Agent, has changed."*

i.e. when the ***set*** changes. This seems to allow an implementation to add a cam and a mic to the ***set*** and fire 1 event.

But [elsewhere](https://w3c.github.io/mediacapture-main/getusermedia.html#mediadevices) it says:

*"When a new media input or output device is made available, and if the permission state of the "list-devices" permission is "granted", the User Agent MUST queue a task that fires a simple event named devicechange at the MediaDevices object. These events are potentially triggered simultaneously across browsing contexts on different origins; user agents can add fuzzing on the timing of events to avoid cross-origin activity correlation."*

... which seems to normatively dictate two events when e.g. I insert my Logitech camera which has both a camera and a microphone on it.

Oddly, the same paragraph allows for time-shifting, so there's apparently no urgency in firing these events, so the sole purpose of firing two events would seem to satisfy someone who is counting device additions and removals (which seems like a poor reason, given that you won't know whether it was an addition or a removal without comparing the list against a previous pull of the list, at which point you'd be able to cope with multiples as well).

Was this our intent, and is this the best behavior? We seem to use the term "device" loosely, so my guess is no. I vote we allow implementations to optimize here and fire one event in this case.

jan-ivar:
Also the "can" should perhaps be a normative "MAY"?

burnburn:
+1 to allowing implementations to fire one or two events, their choice, in this situation.  I don't believe the intention was to be able to count device changes but rather to be able to deal with changes that have occurred.

Yes, I think the "can" should be a "MAY".

Can you please suggest some precise wording changes that also make clear to developers that such an event could mean more than one device has changed?

alvestrand:
Fixed by #376

PR #376, opened 2016-07-28
Allow firing devicechange just once when inserting a camera with mic.

jan-ivar:
Fix for https://github.com/w3c/mediacapture-main/issues/375.


jan-ivar:
The travis error seems unrelated, fyi.

Issue #372, opened 2016-07-08
List-devices permission seems wrong for ondevicechanged (after edits).


jan-ivar:
We're working on `ondevicechange` and trying to figure out *list-devices*, also used for labels. Back [in February](https://w3c.github.io/mediacapture-main/archives/20160222/getusermedia.html#dom-mediadevices-enumeratedevices) we had: **show labels = persistent gum permission OR active gum**.

Then an [editorial change](https://github.com/w3c/mediacapture-main/commit/aadf65a4c9e9dcaeb529f1fecd46a28fc35f37b7#diff-ea76d38900f79cfae8f60e5f7cf16dd1R2343) outsourced it:

  **list-devices = persistent gum permission OR active gum**
  **show label = list-devices**

Then, a [subsequent editorial change](https://github.com/w3c/mediacapture-main/commit/049fd00f87f9ed39a642826fb3ae521cf0e49bb2#diff-ea76d38900f79cfae8f60e5f7cf16dd1R2344) swapped that with:

  **list-devices = persistent gum permission**
  **show label = list-devices OR active gum**

Effectively, ***OR active gum*** was taken out of ***list-devices***, changing its (absent) definition.

Confusingly, we then met and [agreed to](https://github.com/w3c/mediacapture-main/commit/23a07add4f536ae595a88ccffa76f33757a77423) **ondevicechange = list-devices**.

But now **ondevicechange != show labels**, which is not OK, because during an active gum call in Firefox, we want JS to receive `ondevicechange`.

Also, why have a *list-devices* permission separate from *persistent gum permission*? We mostly talk about the latter, and never say how the former is set or acts. Unless one tracks the other, it's also a behavioral change compared to what we had before.

jan-ivar:
I've put up a PR to fix the immediate problem.

It's still odd we have *list-devices* as separate from *persistent permission* when they were the same thing before. Replacing one knob with two is not a neutral change, since the knobs may now be set differently.

alvestrand:
Fixed by #377

PR #377, opened 2016-07-29
JS should see devicechange events during active gum call.

jan-ivar:
Partial remedy for https://github.com/w3c/mediacapture-main/issues/372.


alvestrand:
Seems good to me. Seems to need a rebase.

PR #379, opened 2016-08-04
JS should see devicechange events during active gum call (rebased).

jan-ivar:
Rebase of https://github.com/w3c/mediacapture-main/pull/377.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2870,9 +2870,11 @@ interface NavigatorUserMedia {
       <p>The <code>MediaDevices</code> object is the entry point to the API
       used to examine and get access to media devices available to the User
       Agent.</p>
-      <p>When new media input and/or output devices are made available, and if the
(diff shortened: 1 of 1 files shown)


jan-ivar:
Partial remedy for https://github.com/w3c/mediacapture-main/issues/372.

alvestrand:
This only makes a difference if there is a condition where "list-devices" is not granted and there is an opened device. Previous thinking was that there was no such state, but if list-devices (or "device-info" as the permissions spec calls it) can be manipulated independently, then this makes sense.

jan-ivar:
@alvestrand It is perfectly normal in Firefox to have an active gUM call without persistent permission, so this language seems necessary today, so I'm a little confused. I have been working on the understanding that list-devices is not automatically persisted, are you working off a different understanding? The spec isn't clear.

PR #395, opened 2016-09-08
Mark the Implementation Suggestions as non-normative.

### LINKED ISSUE #390: Clearly mark the "Implementation Suggestions" section as normative or not
https://w3c.github.io/mediacapture-main/getusermedia.html#implementation-suggestions does not have a clear "This section is non-normative." indication, and the Conformance section says

> As well as sections marked as non-normative, all authoring guidelines, diagrams, examples, and notes in this specification are non-normative. Everything else in this specification is normative.

which says the Suggestions section is normative. However, the [change log for its addition](https://w3c.github.io/mediacapture-main/getusermedia.html#june-5-2012) says it's non-normative, and it'd be very strange for a section titled "suggestions" to actually hold requirements.

@jan-ivar's citing the section as normative in https://github.com/w3c/mediacapture-main/issues/387#issuecomment-245021280, so I think the decision either way will be controversial.

jyasskin:
Fixes #390 

@alvestrand Here's the simplest change that fixes the problem. Changing Conformance to say Best Practices are non-normative would work too, but takes more drafting.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3766,7 +3766,7 @@ interface MediaDeviceInfo {
         </div>
       </div>
     </section>
-    <section>
+    <section class="informative">
       <h2>Implementation Suggestions</h2>
       <div class="practice">
         <span class="practicelab" id="resource-reservation">Resource


alvestrand:
@jyasskin this works for me, but the w3c permissions bot doesn't like you. Can you click the "join this group" thing for webrtc at W3C? Google's already a member, so no IPR issue.

We might want to "class"-ify the other sections of the same type while we're at it.

jyasskin:
I've pressed the button to join. You should mention which group makes the bot happy in your CONTRIBUTING.md.

I'd rather separate no-op changes in how the "non-normative" marker is inserted, from this change which adds a marker.

burnburn:
See comment in #390 about reviewing the section before applying this PR.

Issue #388, opened 2016-08-31
Wording concern about the text describing the add track and remove track algorithms


a2sheppy:
I've been working on documentation updates for MDN around media stuff, and ran into an issue yesterday which I worry could throw off others.

First, the way the add a track and remove a track algorithms are described right after the note box in section 4.2 of the spec is confusing because the each operation (adding a track and removing a track) is described twice:

    "To add a track to a MediaStream , the User Agent MUST run the following steps..."
    "To remove a track from a MediaStream , the User Agent MUST run the following steps..."

    "When the User Agent initiates adding a track to a MediaStream , with the exception of initializing a newly created MediaStream with tracks, the User Agent MUST queue a task that runs the following steps..."
    "When the User Agent initiates removing a track from a MediaStream , the User Agent MUST queue a task that runs the following steps..."

This is a key source of confusion. It makes sense that there are two algorithms for each of these, since the operation differs depending on how the track addition or removal is initiated, but they're separated so that you see add then remove then add then remove. That means context is lost, and if you don't read on after the first remove, you may not catch the existence of the second algorithm.

On top of that, the wording is very tricky here. The first sentence means "For the user agent to add a track to a MediaStream, it must...". The third one means just about exactly the same thing, unless you understand that the first one is actually meant to mean "When a script adds a track to a MediaStream, the user agent must..." and the third one means "When the user agent initiates the addition of a track to a MediaStream, it must..."

That difference is not obvious when reading this text. You have to have some advance knowledge of how those phrases are meant to be interpreted, and I'm not sure it's wise to assume that everyone reading a specification will parse writing like that the way it's intended to be parsed. I certainly didn't. Took me several minutes of re-reading to catch on to this interpretation. :)

Finally, a direct quote from the description of the addtrack event:

    "A new MediaStreamTrack has been added to this stream. Note that this event is not fired when the script directly modifies the tracks of a MediaStream ."

The text "this event is not fired when the script directly modifies" suggests that the event isn't fired when a script adds a track by somehow bypassing of the addTrack() method, which isn't what's intended here. It actually means that the event isn't fired if you use addTrack(), which is exactly the opposite.

Anyway, I worry when the way specs are written leaves stuff open to interpretation so I try to point out things that I find unclear, like this one. It's really easy to interpret when the events should be fired exactly backward from what's intended, and that makes me nervous enough to point it out. :)

stefhak:
[...]

PR #399, opened 2016-09-19
Editorial: Unroll add and remove track steps

adam-be:
It might be easier to read the spec with less jumping around if we unroll these short algorithms.

Proposed fix for issue #388

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -289,7 +289,7 @@
           in <var>tracks</var>, run the following steps:</p>
           <ol>
             <li>
-              <p>If <var>track</var> in already in <var>stream</var>'s <a href=
+              <p>If <var>track</var> is already in <var>stream</var>'s <a href=
               "#track-set">track set</a>, continue with the next track, if
               any.</p>
             </li>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -285,8 +285,19 @@
           </ul>
         </li>
         <li>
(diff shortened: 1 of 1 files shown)


adam-be:
Rendered preview: http://rawgit.com/w3c/mediacapture-main/unroll-add-remove-track-steps/getusermedia.html#mediastream

getusermedia.html
```diff
@@ -285,8 +285,19 @@
...
         <li>
-          <p><a href="#stream-add-track">Add each track</a> in
-          <var>tracks</var> to <var>stream</var>.</p>
+          <p>For each <code><a>MediaStreamTrack</a></code>, <var>track</var>,
+          in <var>tracks</var>, run the following steps:</p>
+          <ol>
+            <li>
+              <p>If <var>track</var> is already in <var>stream</var>'s <a href=
+              "#track-set">track set</a>, continue with the next track, if
+              any.</p>
```
alvestrand: Is this read more easily if you write "skip this track" instead of "continue with the next track, if any"?

adam-be:
> Is there a corresponding PR for webrtc-pc, or isn't webrtc-pc using these methods?

I've looked through the webrtc-pc spec. In the 'dispatch a receiver' steps, we talk about adding a track to a stream, with a note saying that this will result in an 'addtrack' event as described in GUM. This is sort of a vague reference to the 'When the User Agent initiates' type of algorithms that we're keeping.

Issue #387, opened 2016-08-26
Reinstate strong language on permission ending when tracks stop, lost by editorial mistake


jan-ivar:
In February, [the spec said](https://w3c.github.io/mediacapture-main/archives/20160222/getusermedia.html#source-stopped):

"When all tracks using a source have been stopped or ended by some other means, the source is stopped. Unless there is a stored permission for the source in question, the given permission is revoked and the User Agent SHOULD also remove the "permission granted" indicator for the source."

This predates editorial attempts at integrating with the permissions spec, and says that while removing the indicator is optional, revoking permission is not, unless it's been "stored". 

Compare to [today](https://w3c.github.io/mediacapture-main/getusermedia.html#source-stopped):

"When all tracks using a source have been stopped or ended by some other means, the source is stopped. If there is no stored permission for the source, the User Agent SHOULD also remove the "permission granted" indicator for the source. If the data is being generated from a live source (e.g., a microphone or camera), then the User Agent SHOULD remove any active "on-air" indicator for that source."

This no longer says anything should happen to the permission, and makes it sounds like stopping all tracks SHOULD revoke permisson. We should clarify that this is not the case with stronger language closer to the February consensus.

jyasskin:
Dropping permission when all tracks are stopped means that a videochat site with a "mute video" button has to choose between

1. Leaving the hardware camera light on while video is "muted", which means users have to blindly trust that the site's no longer transmitting, or
2. Needing to re-prompt for permission when the user unmutes the video.

Neither of these is desirable. However, since:
   * the spec allows a "stored permission" for HTTPS sites,
   * such "storage" can live for just the lifetime of the realm if the UA wants, and 
   * all videochat applications should be HTTPS anyway,

this doesn't affect much in practice.

jan-ivar:
(I think you'll find that "stored permission" meant "persistent permission to origin across sessions" in those days. I think there's evidence of that in the spec. I don't think anyone considered realm).

I agree staring at the camera light on mute is undesirable, and simultaneously I think we'd ideally want `track.enabled` to be viable here (because that's the right API).

The [mediacapture spec](https://w3c.github.io/mediacapture-main/getusermedia.html#life-cycle-and-media-flow) actually says that *"When all tracks connected to a source are muted or disabled, the "on-air" or "recording" indicator for that source can be turned off; when the track is no longer muted or disabled, it MUST be turned back on."*

[...]

PR #401, opened 2016-09-26
Define indicator states.

jan-ivar:
Fix for https://github.com/w3c/mediacapture-main/issues/387 as discussed at TPAC.

I've updated the text from the TPAC slides based on feedback, including:
 - Use maps instead of dynamically composed internal slot names.
 - Reduce logic to boolean states and drop increment counters.
 - Make clear permissions apply only to sources exposed by getUserMedia.
 - Remove the option to set device accessible indicator differently depending on muting strategy.

@burnburn PTAL.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -5150,15 +5150,20 @@ below dictionary, but instead provide its own definition. See
           <a>MediaStream</a></code> containing a track of the new media type
           (see <a href="#mediastreams-in-media-elements"></a>),
         </li>
-        <li>updating <code><a>MediaDeviceKind</a></code>,</li>
+        <li>updating <code><a>MediaDeviceKind</a></code> if the new type has
+        enumerable devices,</li>
         <li>updating the <code>getCapabilities()</code> and
         <code>getUserMedia()</code> descriptions,</li>
         <li>adding the new type to the
-        <code><a>MediaStreamConstraints</a></code> dictionary, and</li>
+        <code><a>MediaStreamConstraints</a></code> dictionary,</li>
         <li>describing any new security and/or privacy considerations (see
         <a href="#privacy-and-security-considerations"></a>) introduced by the
-        new type.
-        </li>
+        new type, and</li>
+        <li>if the new type requires user authorization, defining new
+        permissions for it, and defining how they, along with access starting
+        and ending, as well as muting/disabling, affect any new and/or existing
+        "on-air" and "device accessible" indicator states (see
+        <a href="#mediadevices"></a>).</li>
       </ul>
       <p>Additionally, it should include updating</p>
       <ul>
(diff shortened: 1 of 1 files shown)


ShijunS:
At TPAC, we mentioned adding text for extensibility to cover screen capture and other capture sources with potential privacy concern.  It'd be nice to cover that with the same PR to keep the definition consistent.  One special aspect of the screen capture is that the deviceId is uniquely generated during runtime for each MST.

jan-ivar:
@ShijunS I've been careful to say this only covers sources exposed by getUserMedia. Since screensharing is exposed by getUserMedia, I considered it covered. If you could let me know of any additional language you were looking for specifically that would be great, otherwise I would expect the screensharing spec to menton how it ties in?

Note that temporary deviceIds obviously won't work with persistent permissions, but then in screensharing each screen or window is considered a "device", so per-device permissions doesn't make a lot of sense there anyway perhaps (I expect most browsers will skip the per-device stuff, frankly). Per-kind might suffice, if you're planning on adding any persistent permissions for screensharing at all (we're not due to the security implications). For the purposes of tracking indicators i think we're fine as long as ids are unique during runtime (I consider it implementation details to clean up unused unique ids).

ShijunS:
@jan-ivar.  Fully agree per-device permission persistence doesn't make sense for screen capture or **getDisplayMedia**.  

I noticed already that the PR only covers sources exposed by getUserMedia, so wanted to make sure explicitly whether the intention is to also cover **getDisplayMedia**.  It seems you are saying yes, but it's still a bit confusing to me when you say "screensharing is exposed by **getUserMedia**".  Please help clarify the intention, I can then help with the language if needed.

FWIW, there is still value to have a persistent permission for the getDisplayMedia API.  If it is "allowed", UA can show the thumbnail list for user to pick the display surface (so user is still in final control); in case "denied", UA won't need to enumerate the surfaces and expose the thumbnail UI.  Meanwhile, I agree we can treat that as UA implementation details mostly from UX perspective, so probably can be considered as a stored user _preference_ rather than as a stored user _permission_ per se.

jan-ivar:
@ShijunS My bad, I forgot about getDisplayMedia (Firefox still implements an unofficial and outdated draft where it was still under getUserMedia). How should we write it so that it is covered? Any ideas?

ShijunS:
@Jan-Ivar.  Would Section 14 be the right place for this, maybe as a short subsection on permission?  Further specifics could then be covered in other specs (incl. screen capture), based on the subsection.  I think that should be consistent with the intention from the group at TPAC.

jan-ivar:
@ShijunS I tried to add something, PTAL. Note however that the way the section I modified anticipated extension already does not fully match how `getDisplayMedia` extends things (e.g. screensharing produces video, not really a new media type, per se. e.g. there's no need to extend HTMLMediaElement).

ShijunS:
LGTM

PR #407, opened 2016-10-06
proposed change for #402

### LINKED ISSUE #402: Allow ondevicechange to be fired before permission is granted or any device is opened
As part of the MediaDevices interface, the ondevicechange event is to inform the app to enumerate capture devices.  Given device enumeration is independent from whether there is a device in use or not by the current UA, firing the event should not be restricted either.  

As one use case, assuming we have a desktop machine without any capture device to start with, app can decide not to call gUM and hide the related features.  Once a device is plugged in, app can respond to the ondevicechange event to expose related features, such as a profile photo button, or a voice recording button, etc.  The Edge implementation today supports the use case.  The spec allowed it as well prior to the 20160423 version.

ShijunS:
the first attempt to fix #402

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2966,7 +2966,8 @@ interface NavigatorUserMedia {
       the User Agent MUST run the following steps:</p>
       <ol>
         <li>
-          <p>If <var>storedDeviceList</var> exists, then delete <var>storedDeviceList</var>.</p>
+          <p>If <var>storedDeviceList</var> is not undefined, then delete <var>storedDeviceList</var>
+          so it becomes undefined.</p>
         </li>
         <li>
           <p>Queue a task that fires a simple event named <code><a href=
@@ -3029,7 +3030,7 @@ interface MediaDevices : EventTarget {
                   <p>Run the following steps in parallel:</p>
                   <ol>
                     <li>
-                      <p>If <var>storedDeviceList</var> exists, then let <var>resultList</var> 
+                      <p>If <var>storedDeviceList</var> is not undefined, then let <var>resultList</var> 
                         be a copy of <var>storedDeviceList</var>, and skip steps 2-5 below.</p>
                     </li>
                     <li>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2956,14 +2956,19 @@ interface NavigatorUserMedia {
         </li>
       </ol>
 
-      <p>When new media input and/or output devices are made available, and if either the
-      <a data-lt="retrieve the permission state">permission state</a> of the
-      "list-devices" permission is "granted" or any of the local devices are
-      attached to an active MediaStream in the current browsing context, then
-      the User Agent MUST queue a task
-      that fires a simple event named <code><a href=
-      "#event-mediadevices-devicechange">devicechange</a></code> at the
-      <code><a>MediaDevices</a></code> object. The User Agent MAY fire just a
+      <p>When new media input and/or output devices are made available, 
+      the User Agent MUST run the following steps:</p>
+      <ol>
+        <li>
+          <p>If <var>storedDeviceList</var> exists, then delete <var>storedDeviceList</var>.</p>
+        </li>
+        <li>
+          <p>Queue a task that fires a simple event named <code><a href=
+             "#event-mediadevices-devicechange">devicechange</a></code> at the
+             <code><a>MediaDevices</a></code> object.</p>
+        </li>
+      </ol>
+      <p>The User Agent MAY fire just a
       single event when several devices are added at the same time, e.g. a
       camera with a microphone. These events are potentially
       triggered simultaneously across browsing contexts on different origins;
@@ -3017,6 +3022,10 @@ interface MediaDevices : EventTarget {
                 <li>
                   <p>Run the following steps in parallel:</p>
                   <ol>
+                    <li>
+                      <p>If <var>storedDeviceList</var> exists, then let <var>resultList</var> 
+                        be a copy of <var>storedDeviceList</var>, and skip steps 2-5 below.</p>
+                    </li>
                     <li>
                       <p>Let <var>resultList</var> be an empty list.</p>
                     </li>
@@ -3075,6 +3084,10 @@ interface MediaDevices : EventTarget {
                           <var>resultList</var>.</p>
                         </li>
                       </ol>
+                    </li>
+                    <li>
+                      <p>Update <var>storedDeviceList</var> to be same as 
+                        <var>resultList</var>.</p>
                     </li>
                     <li>If any of the local devices are attached to a
                     <code>live</code> MediaStreamTrack in the current browsing
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2898,11 +2898,11 @@ interface NavigatorUserMedia {
           relevant global object</a>, run the following steps:</p>
           <ol>
             <li>
-              <p>create three internal slots: [[\devicesLiveMap]], [[\devicesAccessibleMap]], and
+              <p>Create three internal slots: [[\devicesLiveMap]], [[\devicesAccessibleMap]], and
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -2956,14 +2956,19 @@
...
-      attached to an active MediaStream in the current browsing context, then
-      the User Agent MUST queue a task
-      that fires a simple event named <code><a href=
-      "#event-mediadevices-devicechange">devicechange</a></code> at the
-      <code><a>MediaDevices</a></code> object. The User Agent MAY fire just a
+      <p>When new media input and/or output devices are made available, 
+      the User Agent MUST run the following steps:</p>
+      <ol>
+        <li>
+          <p>If <var>storedDeviceList</var> exists, then delete <var>storedDeviceList</var>.</p>
```
jan-ivar: Needs to be a [[**storedDeviceList**]] internal slot on the global object in the current realm I think, and set to an empty array.

getusermedia.html
```diff
@@ -3018,6 +3023,10 @@
                   <p>Run the following steps in parallel:</p>
                   <ol>
                     <li>
+                      <p>If <var>storedDeviceList</var> exists, then let <var>resultList</var>
```
jan-ivar: [[storedDeviceList]] is not empty,

getusermedia.html
```diff
@@ -3018,6 +3023,10 @@
                   <p>Run the following steps in parallel:</p>
                   <ol>
                     <li>
+                      <p>If <var>storedDeviceList</var> exists, then let <var>resultList</var> 
+                        be a copy of <var>storedDeviceList</var>, and skip steps 2-5 below.</p>
```
jan-ivar: Since step 5 sets storedDeviceList to resultList, I think we can invert this condition and say:

  *"If [[storedDeviceList]] is empty, then execute the following steps:"*

and move steps 2-5 in to be conditional sub-steps this step which culminate in setting [[storedDeviceList]].

getusermedia.html
ShijunS: @Jan-Ivar, thought about that, an "empty" list is still a valid list, i.e. before the event is fired.

getusermedia.html
ShijunS: @Jan-Ivar, if storedDeviceList exists, we still need to copy storedDeviceList to resultList.  We might need more changes if we want to move 2-5 to sub-steps.

getusermedia.html
```diff
@@ -2956,14 +2956,19 @@
         </li>
       </ol>
 
-      <p>When new media input and/or output devices are made available, and if either the
-      <a data-lt="retrieve the permission state">permission state</a> of the
-      "list-devices" permission is "granted" or any of the local devices are
-      attached to an active MediaStream in the current browsing context, then
```
ShijunS: @Jan-Ivar, I'll update this part based on your input.

getusermedia.html
jan-ivar: @ShijunS Good point. Let's initialize it to `undefined` then, as "existence" can be ambiguous.

getusermedia.html
jan-ivar: Or s/resultList/storedDeviceList/ in step 8.

getusermedia.html
```diff
@@ -2956,11 +2956,18 @@
...
+      "list-devices" permission is "granted", or any of the local devices are 
+      attached to an active MediaStream in the current browsing context, 
+      or the <a href="https://html.spec.whatwg.org/multipage/browsers.html#active-document">active document</a> 
+      in the current browsing context is <a href="https://html.spec.whatwg.org/multipage/browsers.html#fully-active">fully active</a>
+      and <a href="https://html.spec.whatwg.org/#gains-focus">has focus</a>, then
       the User Agent MUST run the following steps:</p>
       <ol>
         <li>
-          <p>If <var>storedDeviceList</var> exists, then delete <var>storedDeviceList</var>.</p>
+          <p>If <var>storedDeviceList</var> is not undefined, then delete <var>storedDeviceList</var>
```
ShijunS: @Jan-Ivar, not sure if that is absolutely needed here given _storedDeviceList_ is not tied directly to any JavaScript object.  And, not sure how to reset an internal slot to _undefined_ either.  Let's get comment from other experts, @adam-be?

jan-ivar: What is it tied to then? As I understand it, an internal slot can be set to `undefined` just like any other value.

jan-ivar: Yes, I put mine on page load [here](https://rawgit.com/w3c/mediacapture-main/fb47384fae1a58393c80bfbd2a798472772fe15f/getusermedia.html#mediadevices).

ShijunS: @jan-ivar, please take another look with the new commit.  I also replaced "undefined" with "null" to indicate the absence of a usable value.

getusermedia.html
```diff
@@ -2966,8 +2972,8 @@
       the User Agent MUST run the following steps:</p>
       <ol>
         <li>
-          <p>If <var>storedDeviceList</var> is not undefined, then delete <var>storedDeviceList</var>
-          so it becomes undefined.</p>
+          <p>If [[\storedDeviceList]] is not null, then set [[\storedDeviceList]]
+          to null.</p>
```
jan-ivar: Or maybe just: "Set [[storedDeviceList]] to null." ?

getusermedia.html
```diff
@@ -3093,7 +3099,7 @@
                       </ol>
                     </li>
                     <li>
-                      <p>Update <var>storedDeviceList</var> to be same as 
+                      <p>Update [[\storedDeviceList]]  to be same as
```
jan-ivar: Or maybe: "set [[storedDeviceList]] to resultList." ?

getusermedia.html
adam-be: I would prefer not to have explicit step numbers in the algorithms, since they need to be updated when a new step is added. You could either use a label and jump to it (see _Permission Failure:_ in the gUM() steps), or restructure as @jan-ivar suggests.

getusermedia.html
```diff
@@ -2956,14 +2962,25 @@
...
-      <a data-lt="retrieve the permission state">permission state</a> of the
-      "list-devices" permission is "granted" or any of the local devices are
-      attached to an active MediaStream in the current browsing context, then
-      the User Agent MUST queue a task
-      that fires a simple event named <code><a href=
-      "#event-mediadevices-devicechange">devicechange</a></code> at the
-      <code><a>MediaDevices</a></code> object. The User Agent MAY fire just a
+      <p>When new media input and/or output devices are made available, if either the 
+      <a data-lt="retrieve the permission state">permission state</a> of the 
+      "list-devices" permission is "granted", or any of the local devices are
```
adam-be: #410 replaces "list-devices" with "device-info". I don't know which will be merged first, but we should make sure it's not forgotten.

adam-be: #410 is merged now. Please update this PR accordingly.

getusermedia.html
ShijunS: Thanks @adam-be for the feedback. I made a few more edits, PTAL.

getusermedia.html
alvestrand: I still haven't figured out a situation where protecting the devicechange event gives any real protection (see previous discussion). Until I see such a scenario, I am in favor of removing all coupling between the devicechange event and the deviceinfo permission.

jan-ivar: I should add virtual machines to that list, if they have USB hooked up.

stefhak:
This LGTM now. Time to merge?

stefhak:
@ShijunS can you resolve the conflicts? We'll then merge.

ShijunS:
I had problem when resolving merge conflict in this PR patch, so replacing with PR #412.

PR #411, opened 2016-10-13
Add User Experience requirements section describing permissions display requirements

burnburn:
This is the second part of the permissions display update (see issue #387) that uses the indicator values defined in PR #401.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -5028,11 +5028,11 @@ below dictionary, but instead provide its own definition. See
   <section>
     <h1>User Experience Requirements</h1>
     <p>For each <var>device</var> for which there is a
-    [[\devicesAccessibleMap]]<var>[device]<\var> internal slot,
+    [[\devicesAccessibleMap]]<var>[device]</var> internal slot,
     define <var>&lt;device&gt;Accessible</var> to be On whenever the
     slot has a value of true and Off otherwise.</p>
     <p>For each <var>device</var> for which there is a
-    [[\devicesLiveMap]]<var>[device]<\var> internal slot,
+    [[\devicesLiveMap]]<var>[device]</var> internal slot,
     define <var>&lt;device&gt;Live</var> to be On whenever the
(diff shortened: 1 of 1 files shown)


burnburn:
@jan-ivar can you please review?

stefhak:
@jan-ivar please review.

getusermedia.html
```diff
@@ -5026,6 +5026,83 @@
...
+        [[\kindsAccessibleMap]]<var>[kind]</var> is true, and Off
+        otherwise.</li>
+        <li>Define <var>&lt;kind&gt;Live</var> to be On
+        whenever <var>&lt;device&gt;Live</var> is On for
+        any <var>device</var> of that <var>kind</var> and
+        Off otherwise.</li>
+      </ul>
+    <p>Define <var>Accessible</var> to be On
+    whenever <var>&lt;kind&gt;Accessible</var> is On for
+    any <var>kind</var> and Off otherwise.</p>
```
alvestrand: Could we rename Accessible as Any-Accessible, to separate it more visually from all the other Accessibles?

burnburn:
Still reviewing.  Will reply during IETF week.

burnburn:
@jan-ivar I agree with the inlining proposal you sent me by email.  Can you submit it as a replacement PR for this one and I'll close this one?  I'm fine if you want to update it with the change from Accessible to Any-Accessible (and same for Live), or if you leave it.  I think it's clear enough without that change, but I also don't think the addition of "Any" to the name makes things worse.

alvestrand:
Note - the reason why I think "Any-Accessible" is better than "Accessible" is that "Accessible" can be read by a casual reader as "All-Accessible", which is not what's intended.
A careful reader will understand what it is, so I don't care that much, but we do have casual readers.

jan-ivar:
@burnburn OK I'll do that.

burnburn:
Superceded by #421.  Will close this when that PR resolves.

PR #412, opened 2016-10-21
Rules for when devicechange events may fire

ShijunS:
This is for issue #402, includes all changes in PR #407.  I'll close #407 separately.


alvestrand:
This resolution says that the UA MUST fire an event at the current browsing context, but doesn't say explicitly what happens with other browsing contexts. Is that intended as an affordance to let UAs do what they want to do (MAY fire), or is it intended to say that events MUST NOT be fired on contexts that don't satisfy the criteria?

Editorial: I don't see language saying what happens when a device is removed. Is it unchanged, or was it missing before too?

alvestrand:
@guidou can you comment?

Issue #414, opened 2016-10-27
Devicechange events when not focus - permitted or forbidden?


alvestrand:
It's not clear from the text in #412 whether firing a devicechange event on tabs without the focus or permission is a MAY or a MUST NOT.

This needs to be clear.

alvestrand:
@jan-ivar can you please comment?

aboba:
@ShijunS Can yuu comment?

alvestrand:
@ShijunS we need your input here. Also @guidou

alvestrand:
@ShijunS I would like to resolve this with MAY, which permits current behavior the way I understand edge does it. I don't feel like I have a compelling security story for why it needs to be MUST NOT; we can lay out the security issues (which come down largely to fingerprinting, I believe), and stress the need for adequate safeguards.

ShijunS:
@Alvestrand, I'm fine with a MAY, which allows more time for all browser vendors to catch up with the specific security protection.

jan-ivar:
@alvestrand, @ShijunS  I read it as a MUST.

I think the confusion is that *"When new media input and/or output devices are made available"* is a global event, yet the rules that follow assume that they are being applied in the context of a browsing context. In other words, there's an implicit "for each browsing context in the system" here, that perhaps needs to be explicit?

Specifically, I do not interpret "[current browsing context](https://www.w3.org/TR/2015/WD-html51-20150506/browsers.html#browsing-context)" to mean "the active tab" only, rather it refers to the current browsing context in the implicit for-loop of all browsing contexts (as best as [I can decipher](https://www.w3.org/TR/2015/WD-html51-20150506/browsers.html#browsing-context-names) this is what "current" tends to refer to when talking about browsing contexts).

When read this way, the rules cover all browsing contexts, and effectively limit firing to contexts that have either `"granted"` permission, an active stream, or active focus.

jan-ivar:
Some clues to naming found in https://www.w3.org/TR/2015/WD-html51-20150506/browsers.html#browsing-context-names

jan-ivar:
I.e. "current" browsing context refers to the context the script is in, vs. "parent" (if in an iframe) or "top-level" browsing context.

jan-ivar:
Is anyone still confused by this? This is a MUST, not a MAY. Anything else is a mis-reading, as I tried to cover in a [previous comment here](https://github.com/w3c/mediacapture-main/issues/414#issuecomment-259831127).

[...]

PR #419, opened 2016-11-22
Make the devicechange consistent with enumerateDevices wrt permissions.

guidou:
EnumerateDevices reflects changes in the set of devices when permission has not been granted. Making devicechange consistent with this adds no additional privacy issues since device changes can be detected by polling enumerateDevices.

This matches what Microsoft Edge is currently shipping and what Chromium has implemented behind a flag and intends to ship.

This PR addresses issue #414

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -5243,7 +5243,7 @@ below dictionary, but instead provide its own definition. See
       <li>[#398] Editorial: Fix links to MediaStream's methods</li>
       <li>[#399] Unroll steps in 'add a track' and 'remove a track'</li>
       <li>[#401] Define states to be used in indicator requirements</li>
-      <li>[#410] Rename list-devices permission to device-info to</li>
+      <li>[#410] Rename list-devices permission to device-info to
       match permission spec</li>
       <li>[#412] Specify rules for when devicechange events may fire</li>
       <li>[#415] Remove unions in MediaTrackCapabilities</li>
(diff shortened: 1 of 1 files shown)


guidou:
@alvestrand @ShijunS @jan-ivar , PTAL

jan-ivar:
This PR removes the protections added and agreed to in https://github.com/w3c/mediacapture-main/issues/402. I see no consensus for this.

https://github.com/w3c/mediacapture-main/issues/414 requires either no action, or at most minor clarification, as [I claim](https://github.com/w3c/mediacapture-main/issues/414#issuecomment-259831127) it mis-reads of the spec.

jan-ivar:
> Making devicechange consistent with this adds no additional privacy issues since device changes can be detected by polling enumerateDevices.

@guidou `enumerateDevices` cannot be used to poll for new data ahead of a `devicechange` event, so this is not true. See [[storedDeviceList]] in step 2 in [enumerateDevices](https://rawgit.com/w3c/mediacapture-main/874e5356e0273dbcad1a60a0d24d071e9fba588e/getusermedia.html#dom-mediadevices-enumeratedevices).

jan-ivar:
I think this is another case of https://github.com/w3c/webrtc-pc/issues/942. Confusion because github.io lags behind tip.

guidou:
I was not aware of the spec change for enumerateDevices. Closing.

PR #420, opened 2016-12-02
Made removal of device(s) fire devicechange event.

### LINKED ISSUE #413: Need rules for devicechange event when devices are removed.
#412 seemed to have rules for raising events when devices are added, but not when they're removed.

This needs to be checked, and possibly more text added.

stefhak:
Close #413


alvestrand:
lgtm.
pinging @guidou for his review.

guidou:
lgtm

PR #421, opened 2016-12-12
Add Privacy Indicator Requirements

jan-ivar:
Fix for https://github.com/w3c/mediacapture-main/pull/411 based on @burnburn's pr. PTAL.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -5096,9 +5096,9 @@ below dictionary, but instead provide its own definition. See
       <li>Any false-to-true transition indicated MUST remain observable for a
       sufficient time that a reasonably-observant user could become
       aware of it.</li>
-      <li>Any of the above indications MAY be combined as long as
-      the combined indication cannot transition to off if any of its
-      component indications are still on.</li>
+      <li>Any of the above transition indications MAY be combined as long as
+      the combined indication cannot transition to false if any of its
+      component indications are still true.</li>
     </ul>
     <p>and the following are encouraged behaviors for the User Agent:</p>
     <ul>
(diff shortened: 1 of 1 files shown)


ekr:
I'm not convinced that this level of detail is appropriate to be imposing on an implementation at the specification level. I think it's reasonable to require implementations to provide some user-observable indication, but beyond that, we should let them figure it out.

jan-ivar:
@ekr If it helps any, no browser implements per-device permissions today, which is the bulk of the complexity in this section. The other half is allowing cellphone browsers to forgo always-visible state-indicators (it would have been less wordy to be more imposing). So in practice this is a lot simpler to implement than it sounds:

 * Keep track of "live" and "accessible" for the user using indicators (per kind if you wish).

Maybe the text is a bit paranoid in places, but better safe than sorry. FWIW the language ties in with the permissions spec which operates at this level of detail. Always looking for ways to simplify.

ekr:
As noted, I don't agree that having this level of prescriptiveness is helpful. You are of course free to feel differently, but I do not believe we should merge this PR>

jan-ivar:
@ekr This is the list of requirements Dan presented at TPAC (slides 26-28), that we [decided](https://lists.w3.org/Archives/Public/public-media-capture/2016Oct/0033.html) to move forward with. Which requirement do you want to loosen?

ekr:
As I said initially, I believe a simple statement of intent as indicated in my first comment is sufficient.

alvestrand:
Seems to me that implementations have a relatively large latitude even with this spec.
The usual mantra is "the observable behavior should be the same as what would be done if you executed this algorithm", and with the latitudes allowed here, that shouldn't be too prescriptive.

burnburn:
This looks good to me, and it properly matches what we agreed to at TPAC.

PR #422, opened 2016-12-16
Clarify when and where devicechange events fire.

jan-ivar:
Fix for https://github.com/w3c/mediacapture-main/issues/414.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2995,8 +2995,8 @@ interface NavigatorUserMedia {
       <a href="https://html.spec.whatwg.org/#gains-focus">gains focus</a>), the
       User Agent MUST execute the steps at that time.</p>
       <p>The User Agent MAY combine firing multiple events into firing one event
-      when several devices are added or removed at the same time,
-      e.g. a camera with a microphone.</p>
+      when several events are due or when multiple devices are added or removed
+      at the same time, e.g. a camera with a microphone.</p>
       <p>These events are potentially triggered simultaneously across browsing
       contexts on different origins; user agents MAY add fuzzing on the timing
       of events to avoid cross-origin activity correlation.</p>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2965,13 +2965,20 @@ interface NavigatorUserMedia {
       <p>When new media input and/or output devices are made available,
       or any available input and/or output device becomes unavailable, the User
       Agent MUST run the following steps only in browsing contexts where at
-      least one of the following criteria are met: The
-      <a data-lt="retrieve the permission state">permission state</a> of the
-      "device-info" permission is "granted", any of the input devices are
-      attached to an active MediaStream in the browsing context,
-      or the <a href="https://html.spec.whatwg.org/multipage/browsers.html#active-document">active document</a> 
-       is <a href="https://html.spec.whatwg.org/multipage/browsers.html#fully-active">fully active</a>
-      and <a href="https://html.spec.whatwg.org/#gains-focus">has focus</a>:</p>
+      least one of the following criteria are met:
+      <ul>
+        <li>The <a data-lt="retrieve the permission state">permission state</a>
+        of the "device-info" permission is "granted",</li>
+        <li>any of the input devices are attached to an active MediaStream in
+        the browsing context, or</li>
+        <li>the
+        <a href="https://html.spec.whatwg.org/multipage/browsers.html#active-document">
+        active document</a> is
+        <a href="https://html.spec.whatwg.org/multipage/browsers.html#fully-active">
+        fully active</a>
+        and <a href="https://html.spec.whatwg.org/#gains-focus">has focus.</a></li>
+      </ul>
+      The steps are:</p>
       <ol>
         <li>
           <p>Set [[\storedDeviceList]] to null.</p>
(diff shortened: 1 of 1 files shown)


jan-ivar:
[Preview link](https://rawgit.com/jan-ivar/mediacapture-main/e4f83eddd52a10eea69702a02268597b000449f9/getusermedia.html#mediadevices).

getusermedia.html
```diff
@@ -2963,14 +2963,15 @@
       </ol>
 
       <p>When new media input and/or output devices are made available,
-      or any available input and/or output device becomes unavailable, if either
-      the <a data-lt="retrieve the permission state">permission state</a> of the
-      "device-info" permission is "granted", or any of the local devices are
-      attached to an active MediaStream in the current browsing context,
+      or any available input and/or output device becomes unavailable, the User
+      Agent MUST run the following steps only in browsing contexts where at
+      least one of the following criteria are met: The
```
alvestrand: Can you make this into a bulleted list? I find it quite hard to read this as contiguous text.

getusermedia.html
```diff
@@ -2981,12 +2982,14 @@
...
         </li>
       </ol>
-      <p>The User Agent MAY fire just a
-      single event when several devices are added or removed at the same time,
-      e.g. a camera with a microphone. These events are potentially
-      triggered simultaneously across browsing contexts on different origins;
-      user agents MAY add fuzzing on the timing of events to avoid cross-origin
-      activity correlation.</p>
+      <p>The User Agent MUST NOT execute the above steps in browsing contexts
+      where none of the criteria are met.</p>
```
jan-ivar: Good point.

ShijunS: One comment on "polling devices" when the browsing context becomes "fully active" (and "has focus") - from the implementation perspective, the UA should make sure [[storedDeviceList]] is null so enumerateDevices() can get the latest list of devices based on the existing algorithm.  It is fine in case the UA simply reloads the tabs when they are resumed from a suspended state.

One question - is there any typical example where UA postpones (and accumulates) events until the browsing context is "fully active" (and "has focus")?

jan-ivar: @alvestrand @ShijunS  I've tried to address your feedback. PTAL. Shijun, I failed to find a precedent, though I didn't look too hard (I asked some folks).

getusermedia.html
```diff
@@ -2983,7 +2990,10 @@
         </li>
       </ol>
       <p>The User Agent MUST NOT execute the above steps in browsing contexts
-      where none of the criteria are met.</p>
+      where none of the criteria are met. If such a browsing context later comes
+      to meet the criteria (e.g.
+      <a href="https://html.spec.whatwg.org/#gains-focus">gains focus</a>), the
+      User Agent MUST execute the steps at that time.</p>
```
ShijunS: Thanks @Jan-Ivar.  Given the potential lack of precedent, I expect a "May" would be a better option for the delayed firing than a "Must".  It'd allow some flexibility in UA implementations.

jan-ivar: @ShijunS I had that initially, but I think it needs to be MUST because otherwise the [[storedDeviceList]] isn't cleared when getting focus, permission or gUM.

ShijunS: Good point - I'm thinking to just reset the [[storedDeviceList]] internally as a starting point, implementation wise.  Delaying the events might take some architecture change, if not done yet in the past.

A separate comment: it's not super clear whether the UA should fire a single event (in this case) when there are multiple device changes ahead of time.

jan-ivar: I've added this as a reason to the combining events language. It's a MAY but I think that should suffice.

stefhak:
LGTM.

jan-ivar:
Travis's "ReSpec error: Duplicate definition of 'getcapabilities'" is not caused by this PR.

getusermedia.html
```diff
@@ -2963,14 +2963,22 @@
...
-      or any available input and/or output device becomes unavailable, if either
-      the <a data-lt="retrieve the permission state">permission state</a> of the
-      "device-info" permission is "granted", or any of the local devices are
-      attached to an active MediaStream in the current browsing context,
-      or the <a href="https://html.spec.whatwg.org/multipage/browsers.html#active-document">active document</a> 
-      in the current browsing context is <a href="https://html.spec.whatwg.org/multipage/browsers.html#fully-active">fully active</a>
-      and <a href="https://html.spec.whatwg.org/#gains-focus">has focus</a>, then
-      the User Agent MUST run the following steps:</p>
+      or any available input and/or output device becomes unavailable, the User
+      Agent MUST run the following steps only in browsing contexts where at
```
alvestrand: Would it be clearer if we said "... MUST run the following steps in browsing contexts where at least one of the following criteria are met, but in no other contexts"?

getusermedia.html
```diff
@@ -2981,12 +2989,17 @@
...
       </ol>
-      <p>The User Agent MAY fire just a
-      single event when several devices are added or removed at the same time,
-      e.g. a camera with a microphone. These events are potentially
-      triggered simultaneously across browsing contexts on different origins;
-      user agents MAY add fuzzing on the timing of events to avoid cross-origin
-      activity correlation.</p>
+      <p>The User Agent MUST NOT execute the above steps in browsing contexts
+      where none of the criteria are met. If such a browsing context later comes
+      to meet the criteria (e.g.
```
alvestrand: If taking the previous suggestion, the "MUST NOT" is already stated, so this can be "If a browsing context later comes to meet the criteria".

jan-ivar:
@alvestrand Suggestions taken. Sorry for dropping the ball.

PR #427, opened 2017-01-12
Updating text on whether context object is allowed to request access.

stefhak:
A first attempt on #268.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -66,9 +66,17 @@
         a simple event</a></dfn> are defined in [[!HTML5]].</p>
         <p>The terms <dfn><a href=
         "http://dev.w3.org/html5/spec/webappapis.html#event-handlers">event
-        handlers</a></dfn> and <dfn><a href=
+        handlers</a></dfn>, <dfn><a href=
         "http://dev.w3.org/html5/spec/webappapis.html#event-handler-event-type">
-        event handler event types</a></dfn> are defined in [[!HTML5]].</p>
+        event handler event types</a></dfn> and <dfn><a href=
+        "https://www.w3.org/TR/html5/webappapis.html#responsible-document">
+        responsible document</a></dnf> are defined in [[!HTML5]].</p>
+        <p>The term <dnf><a href="https://www.w3.org/TR/html51/webappapis.html#current-settings-object">
+        current settings object</a></dnf> is defined in [[!HTML51]].</p>
+        <p>The term <dnf><a href="https://html.spec.whatwg.org/multipage/embedded-content.html#allowed-to-use">
+        allowed to use</a></dnf> is defined in WHATWG (note we don't have a proper
+        ref added, in addition it seems HTML5.1 will(?) add defintion
+        of this term).</p>
       </dd>
       <dt><code><dfn>DOMException</dfn></code></dt>
       <dd>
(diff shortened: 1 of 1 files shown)


alvestrand:
@annevk can you say if this text looks appropriate?

annevk:
As I said in the main issue, @zcorpan is changing how this works. If we make `allowusermedia` static. The check here might need to be different, not sure.

But overall this looks okay, except that "this context object" should just be "context object".

getusermedia.html
```diff
@@ -3578,8 +3576,14 @@
                   succeed.</p>
                 </li>
                 <li>
-                  <p>If the current [[!HTML51]] Document's <a>user media
-                  enabled flag</a> is unset, return a promise rejected with a
+                  <p>If this <a href="https://dom.spec.whatwg.org/#context-object">
```
zcorpan: This should say
> If the current settings object's responsible document is not allowed to use the feature indicated by attribute name allowusermedia, then throw a "SecurityError" DOMException.

See https://w3c.github.io/browser-payment-api/#constructor

(In particular, `MediaDevices` is not a node, so "node document" is undefined.)

The allowed to use check needs to be in navigator.getUserMedia() as well, right?

alvestrand: navigator.getUserMedia() is a shell around navigator.mediaDevices.getUserMedia(), so no.

zcorpan: OK

stefhak: @zcorpan I like your proposal to use "settings object" and "responsible document" rather than "context object" and "node document". 

One thing I was thinking about when doing this though was if we (some time) would allow a worker to use gUM. Would it still be clear what the "responsible document" is?

stefhak:
I updated as proposed by @zcorpan in https://github.com/w3c/mediacapture-main/pull/427/files/36c8de191cef114cf52964b3e35d2f2cb6140d47#r95865625 with the except for that I kept it at returning a rejected promise rather than throwing (we did discuss and agree that is the right way, right?).

zcorpan:
Oops. Yeah that was sloppy copy-paste on my part.

getusermedia.html
```diff
@@ -66,9 +66,17 @@
...
-        event handler event types</a></dfn> are defined in [[!HTML5]].</p>
+        event handler event types</a></dfn> and <dfn><a href=
+        "https://www.w3.org/TR/html5/webappapis.html#responsible-document">
+        responsible document</a></dfn> are defined in [[!HTML5]].</p>
+        <p>The term <dfn><a href="https://www.w3.org/TR/html51/webappapis.html#current-settings-object">
+        current settings object</a></dfn> is defined in [[!HTML51]].</p>
+        <p>The term <dfn><a href="https://html.spec.whatwg.org/multipage/embedded-content.html#allowed-to-use">
+        allowed to use</a></dfn> is defined in WHATWG (note we don't have a proper
+        ref added, in addition it seems HTML5.1 will(?) add defintion
+        of this term).</p>
```
alvestrand: Suggest removing parenthesis before merging.

getusermedia.html
```diff
@@ -74,9 +74,7 @@
         <p>The term <dfn><a href="https://www.w3.org/TR/html51/webappapis.html#current-settings-object">
         current settings object</a></dfn> is defined in [[!HTML51]].</p>
         <p>The term <dfn><a href="https://html.spec.whatwg.org/multipage/embedded-content.html#allowed-to-use">
-        allowed to use</a></dfn> is defined in WHATWG (note we don't have a proper
-        ref added, in addition it seems HTML5.1 will(?) add defintion
-        of this term).</p>
+        allowed to use</a></dfn> is defined in WHATWG.</p>
```
zcorpan: Should probably say WHATWG HTML

stefhak: Thanks, added that.

Issue #360, opened 2016-05-25
Specify relation between return from getConstraints and constraints argument


alvestrand:
The current constrainable-pattern getConstraints() text says:

"The getConstraints method returns the Constraints that were the argument to the most recent successful call of applyConstraints(), maintaining the order in which they were specified. Note that some of the optional ConstraintSets returned may not be currently satisfied. To check which ConstraintSets are currently in effect, the application should use getSettings."

I think we should specify that what's returned is the semantic equivalent of what was passed in, as seen from the browser. When implementing this, I found that it was onerous to preserve the exact form passed in. In particular:
- Unrecognized constraints are stripped away by the parser. We shouldn't have to return them.
- Since { width: { ideal: 570 }} and { width: 570 } are semantically identical, we should either be free to return either form, or we should require that the simpler form be returned.
- Same for { deviceId: [ "foo" ] } and { deviceId: "foo" }

Suggested modification:

"The getConstraints method returns a set of constraints that is semantically equivalent to the argument to the most  recent successful call of applyConstraints(), maintaining the order of "advanced" elements that was specified. Note that...."

alvestrand:
@burnburn might want to take a look

stefhak:
I see no issue in moving forward with Harald's suggestion. I see no point in having to return _exactly_ what was passed in rather than something that is semantically equivalent.
BTW, I assume that omitting an unrecognized constraint is considered being "semantically equivalent".

alvestrand:
@stefhak I think so too - the invariant should be that track.applyConstraints(track.getConstraints()) should lead to "no change in configuration".

jan-ivar:
> Unrecognized constraints are stripped away by the parser. We shouldn't have to return them.

This part is implicit in WebIDL and dictionaries, thus shouldn't need any language change IMHO.

As to your other points, I'm not so convinced, because that normalizing is not done by WebIDL.

Can't you simply store the dictionary before you normalize it? That's what [I'm doing](https://bugzil.la/1213517).

jan-ivar:
That said, due to our spec's novel use of dictionaries in unions, what I get back from

```js
navigator.mediaDevices.getUserMedia({video: true})
.then(s => console.log(s.getConstraints());
```
I think would surprise most people: 
```
{
  "deviceId":{},"echoCancellation":{},"facingMode":{},"frameRate":{},"height":{},
  "mediaSource":"camera","mozAutoGainControl":{},"mozNoiseSuppression":{},
[...]

PR #430, opened 2017-01-18
Allow getConstraints to return semantically equivalent constraint object

burnburn:
Fixes issue #360.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -4225,10 +4225,10 @@ interface ConstrainablePattern {
               they were specified. Note that some of the optional
               ConstraintSets returned may not be currently satisfied. To check
               which ConstraintSets are currently in effect, the application
-              should use <code>getSettings</code>.  The UA MAY return
-              normalized constraints with the identical effect in all
-              situations as the applied constraints, regardless of
-              source.</p>
+              should use <code>getSettings</code>.  Instead of
+              returning the exact constraints as described above, the
+              UA MAY return a constraint set that has the identical
+              effect in all situations as the applied constraints.</p>
               <div>
                 <em>No parameters.</em>
               </div>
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -4225,7 +4225,10 @@
               they were specified. Note that some of the optional
               ConstraintSets returned may not be currently satisfied. To check
               which ConstraintSets are currently in effect, the application
-              should use <code>getSettings</code>.</p>
+              should use <code>getSettings</code>.  The UA MAY return
+              normalized constraints with the identical effect in all
```
stefhak: is it clear what "normalized" means here? We only use that word at one other place in the document, and then it has to do with values of supported capabilities.

getusermedia.html
```diff
@@ -4225,7 +4225,10 @@
               they were specified. Note that some of the optional
               ConstraintSets returned may not be currently satisfied. To check
               which ConstraintSets are currently in effect, the application
-              should use <code>getSettings</code>.</p>
+              should use <code>getSettings</code>.  The UA MAY return
```
alvestrand: Slightly outside the PR, but "To check which ConstraintSets are currently in effect" -> "To check which settings have been chosen based on the ConstraintSets". getSettings returns values for constraints that haven't been constrained, too.

getusermedia.html
alvestrand: Suggest deleting the word. Perhaps "MAY return a constraint set that has the identical effect in all situations as the applied constraint, but does not need to return an identical constraint set."

getusermedia.html
burnburn: Yes, this is outside the changes in the PR.  Better to make a new issue or PR for this.

burnburn:
Please review again, @alvestrand and @stefhak .  I have made similar changes to what you suggested.

stefhak:
This LGTM!

Issue #394, opened 2016-09-08
Browser's ability to dynamically change settings mistakenly removed


burnburn:
In trying to write the PR for issue #360, I (re-) discovered that our long-ago change to use Select Settings only when applyConstraints() is called has effectively eliminated the possibility for the platform to change the settings on its own as long as the constraints continue to be satisfied, which does not match the original intent of constraints.  This needs to be fixed.

jan-ivar:
FWIW, when I implemented `getSettings()` in Firefox 50, I found the biggest omission in the spec in this area to be tabs competing for the camera/microphone (most browsers allow this).

In considering the settings of a singular hardware device, an implementation has to consider constraints from *all* active-gUM tabs, not just one (note there's no per-track downscaling in FF [yet](https://bugzil.la/1286945)).

I don't think the spec needs to lay this out in detail, but probably worth a mention (I devised a way to produce a set of net constraints (or `OverconstrainedError`), given a series of competing constraints).

alvestrand:
I think this belongs in the "life-cycle" section (presently in 4.3.1) - something like "During the lifetime of the track, events such as other requests or external envrionment changes may cause the device's operational mode to change. In such a case, the UA is free to run the "select settings" algorithm again; if the new conditions mean that the present constraints cannot be satisified, it MUST queue a task to fire an "overconstrained" event at the track."

jan-ivar:
@burnburn said "change the settings on its own as long as the constraints continue to be satisfied".

@alvestrand said "if the new conditions mean that the present constraints cannot be satisified, it MUST queue a task to fire an "overconstrained" event"

Which is it?

FWIW Firefox implements the first for competing tabs at least. That is, once you've managed to apply `{exact}` to a capability successfully, another tab can't change it on you, or cause `OverConstrainedError` on you, rather the other tab will fail to apply its competing constraints.

We don't (currently) detect drops in `frameRate` due to lighting or anything like that.

alvestrand:
These two statements are consistent.

jan-ivar:
Perhaps I misunderstood. I interpreted @burnburn's request as wanting to restore the UAs ability to change a source within the envelope of existing constraints on it. If so, this would never throw `OverconstrainedError`. @burnburn is that what you meant?

@alvestrand With "external environment changes" you mean e.g. lighting effects on `frameRate`?

[...]

PR #431, opened 2017-01-18
Require all constraints to be satisfied when settings are changed

burnburn:
Also moves the paragraph containing this change earlier in the doc so it's easier to find.

Fix for issue #394 .


stefhak:
LGTM.

alvestrand:
LGTM

PR #432, opened 2017-02-02
Ignore ideal values that don't belong on the device

### LINKED ISSUE #409: Clarify behavior of using unrelated constraints
The following programming error is not a `TypeError` in our spec:

``` js
if (navigator.mediaDevices.getSupportedConstraints().width) {
  let stream = await navigator.mediaDevices.getUserMedia({audio: {width: 640}}); // audio
}
```

My best reading of the current spec is that it should be silently ignored, which seems defensible given how dictionary members tend to work ("ignore unrelated members"), so that's not my question.

My question is, should the following reject with `OverconstrainedError`?

``` js
if (navigator.mediaDevices.getSupportedConstraints().width) {
  let stream = await navigator.mediaDevices.getUserMedia({audio: {width: {exact: 640}}});
}
```

It seems to hinge on reading this language:
- _"If the constraint is required ('min', 'max', or 'exact'), and the [settings dictionary](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-settings-dictionary)'s value for the constraint does not satisfy the constraint, the fitness distance is positive infinity."_

"Satisfy" does not appear defined anywhere. Specifically, it doesn't say how to handle when a settings-dictionary does not have a value for the constraint, i.e. the constraint is irrelevant.

alvestrand:
Ideal values (like height for an audio device) should be ignored
when calculating fitness distance. They still matter when exact.

Closes #409

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -4291,6 +4291,13 @@ interface ConstrainablePattern {
                   does not satisfy the constraint, the fitness distance is
                   positive infinity.</p>
                 </li>
+                <li>
+                  <p>
+                    If the constraint is not required, and does not apply for
+                    this type of device, the fitness distance is 0 (that is,
+                    the constraint does not influence the fitness distance).
+                  </p>
+                </li>
                 <li>If no ideal value is specified, the fitness distance is
                 0.</li>
                 <li>For all positive numeric non-required constraints (such as


stefhak:
LGTM.

jan-ivar:
lgtm too (catching up).

PR #433, opened 2017-02-03
Specify the "applyConstraints algorithm" more clearly

### LINKED ISSUE #386: What does "getSettings" return?
In implementing getSettings(width, height), I have been faced with a quandary.
There are 3 width/height pairs to consider:
1. The device native width/height. Given that we do cropping, this is not what we want.
2. The target width/height. This is a bit tricky to calculate, since it's not necessarily a fixed value - it is influenced by the size of the image from the source and the applied constraints, including aspect ratio, and considering that some directions may be unbounded; it's not clear that the answer will make sense if we just look at the target independent of the image from the source.
3. The frame size actually delivered (after cropping). Given the previous point, this may be the easiest value to return consistently - but it leaves us at 0x0 until the first frame arrives.

What do people think "getSettings" should return?

alvestrand:
Defines the "applyConstraints algorithm" separately from the
applyConstraints function, and links to algorithm in getUserMedia calls.

Fixes #386

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -4246,7 +4246,6 @@ interface ConstrainablePattern {
             </dd>
             <dt><code>getConstraints</code></dt>
             <dd>
-              <p>The <dfn>getConstraints()</dfn> method returns the Constraints
               <p>The <dfn>getConstraints</dfn> method returns the Constraints
                 that were the argument to the most recent successful invocation of
                 the <a>applyConstraints algorithm<a> on the object,
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -4249,7 +4249,7 @@ interface ConstrainablePattern {
               <p>The <dfn>getConstraints()</dfn> method returns the Constraints
               <p>The <dfn>getConstraints</dfn> method returns the Constraints
                 that were the argument to the most recent successful invocation of
-                the <code>applyConstraints algorithm</code> on the object,
+                the <a>applyConstraints algorithm<a> on the object,
                 maintaining the order in which
                 they were specified. Note that some of the advanced
               ConstraintSets returned may not be currently satisfied. To check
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -4236,9 +4236,9 @@
             <dt><code>getConstraints</code></dt>
             <dd>
               <p>The <dfn>getConstraints</dfn> method returns the Constraints
-              that were the argument to the most recent successful call of
-              <code>applyConstraints()</code>, maintaining the order in which
-              they were specified. Note that some of the optional
+              that were the argument to the most recent successful invocation of
+              the <code>applyConstraints algorithm</code>, maintaining the order in which
```
jan-ivar: This seems wrong, unless the sole invocation of the applyConstraints algorithm is from the applyConstraints method, in which case I think it was clearer before. Also both the old and new text here fail to be providing the appropriate context, which is *on this track*.

alvestrand: "On this track" did not leave a link.
I don't see what's wrong - could you be clearer?
If the last call was getUserMedia, the getuserMedia call will now invoke the applyConstraints algorithm - that's the other part of this change. In that case, the argument from there will be returned from getConstraints; if there weren't any constraints, an empty object will be returned.

jan-ivar: We want to cache the last thing this caller successfully set on this track, not the last thing successfully set by anyone who calls this algorithm from anyplace or on any object, whether it be on this track or other tracks, or even something else implementing the Constrainable pattern in the future. JS can call getUserMedia multiple times even in the same script, so the cache needs to be stored on the track object itself. We should say that.

Maybe something like: *"that were the argument to the most recent successful invocation of the applyConstraints method on this track."*

jan-ivar: s/track/constrainable object/

alvestrand: Thanks for bearing with my density. Pushed a fix.

getusermedia.html
```diff
@@ -4247,9 +4247,11 @@
             <dt><code>getConstraints</code></dt>
             <dd>
               <p>The <dfn>getConstraints()</dfn> method returns the Constraints
-              that were the argument to the most recent successful call of
-              <code>applyConstraints()</code>, maintaining the order in which
-              they were specified. Note that some of the optional
+              <p>The <dfn>getConstraints</dfn> method returns the Constraints
+                that were the argument to the most recent successful invocation of
```
jan-ivar: Indentation inconsistent with surrounding text.

stefhak:
LGTM.

getusermedia.html
```diff
@@ -4247,9 +4247,11 @@
...
+                the <a>applyConstraints algorithm<a> on the object,
```
stefhak: Travis complains about the second <a> (should be a </a>).

getusermedia.html
burnburn: Also a duplication of the first line.

burnburn:
I will fix the typos and merge.

PR #440, opened 2017-03-30
Replacing 'allowusermedia' with policy 'feature's 'camera' and 'microphone'.

### LINKED ISSUE #434: Disable user media by default in cross-origin iframes
We would like to explore disabling user media (camera/mic/speakers) by default for cross-origin iframes. The idea is that it would be possible for the embedder to re-enable user media using the proposed [Feature Policy](https://docs.google.com/document/d/1k0Ua-ZWlM_PsFCFdLMa8kaVTo32PeNZ4G7FFHqpFx4E/edit#heading=h.4yubgixv5l6b) mechanism.

getUserMedia already has a failure mode that occurs as a result of the user denying permission for hardware. This same failure mode can be reused but we probably still want to alter the spec to include the additional check to see if the feature is allowed by Feature Policy.

I think we probably want to add 3 features: camera, microphone and speakers, align with the permissions that exist in the [permissions spec](https://w3c.github.io/permissions/#permission-registry).

Note that this issue is mainly just intended to start the discussion about this change. We would need to nail down the Feature Policy spec before moving forward. Guidance is appreciated :).

The motivations for this change and a discussion of compatibility risk can be found here: https://docs.google.com/document/d/13dp9xWVyGM8THAQohDOT2mMOTSGLxEhSZEvgpmVLrxU/edit

@clelland @alvestrand

stefhak:
Fixes #434 

Very first draft - do not merge!!

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -197,8 +197,10 @@
       </dd>
       <dt>Features</dt>
       <dd>
-        <p>The terms <dfn>feature</dfn>, <dfn>feature</dfn> and 
-        <dfn>default allowlist</dfn> are defined in [[!featurepolicy]].</p>
+        <p>The terms <dfn><a href="https://wicg.github.io/feature-policy/#feature">feature</a></dfn>,
+        <dfn><a href="https://wicg.github.io/feature-policy/#feature-name">feature name</a></dfn> and 
+        <dfn><a href="https://wicg.github.io/feature-policy/#default-allowlist">default allowlist</a></dfn>
+        are defined in [[!featurepolicy]].</p>
       </dd>
 
     </dl>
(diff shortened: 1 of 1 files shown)


raymeskhoury:
@stefhak thanks so much for looking at this. Sorry for the delay I've been OOO.

@clelland would be best to take a look at the details of the spec integration. I'm definitely not well versed with this spec, but at a high level though I'm wondering if these checks should be done at the same point that we check to see whether the user agent has permission. These feature policy features align exactly with permissions defined in the permissions API (for camera, mic, speakers) and so ideally we want the behavior of this API to be the same whether camera was disabled by feature policy or by the user agent. What do you think?

stefhak:
@raymeskhoury I think you are right. I need to check a bit more, but my gut feeling is that we should (as you say) talk about whether the app has permission (and refer to where that is defined) rather than what I've done i the PR.

getusermedia.html
```diff
@@ -195,6 +195,14 @@
         permissions <a href=
         "https://github.com/w3c/permissions/issues/62">issue 62</a>.</p>
       </dd>
+      <dt>Features</dt>
+      <dd>
+        <p>The terms <dfn><a href="https://wicg.github.io/feature-policy/#feature">feature</a></dfn>,
```
clelland: This has been changed in the most recent draft to "policy-controlled feature", so that we don't camp on such a generic term.

getusermedia.html
```diff
@@ -3543,6 +3551,17 @@
       <p>The <code>getSupportedConstraints</code> method is provided to allow
       the application to determine which constraints the User Agent
       recognizes.</p>
+      <h4>Feature Definitions</h4>
+        <p>The <dfn data-lt-nodefault="" data-lt="camera" id=
+        "camera">camera</dfn>  <a>feature</a> controls access to
```
clelland: Same as above -- I think you'd need to word it something like

    <dfn data-lt="camera">camera</dfn> is a <a>policy-controlled feature</a>,
    which controls access to video input devices.

aboba:
TPAC feedback indicates that Chrome, Firefox and Safari are considering implementing this.

raymeskhoury:
We've implemented this in Chrome, it's going out in M64. I have a PR up on the Permissions spec:
https://github.com/w3c/permissions/pull/163

If we can merge that, then we will get basic integration without needing a change to the mediacapture spec, because it already queries the permissions spec for access. We would only need to do something if we wanted to a custom error message but it may be OK as a first cut.

alvestrand:
Let's close this one and wait for w3c/permissions#163 to merge.

Issue #444, opened 2017-04-11
Add autoGainControl and noiseSuppression constraints.


jan-ivar:
Chrome has `googAutoGainControl` and `googNoiseSuppression`,
& Firefox has `mozAutoGainControl` and `mozNoiseSuppression`.

With two working implementations, we should standardize `autoGainControl` and `noiseSuppression` to avoid [prefix pollution](https://hsivonen.fi/vendor-prefixes/).

With this in the spec, my hope is Google would rethink [moving](https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/mba5ziDLFU4/ZlODIY3CDgAJ) `googAutoGainControl` and `googNoiseSuppression` into `navigator.mediaDevices.getSupportedConstraints`, and we would remove our `mozAutoGainControl` and `mozNoiseSuppression` mistakes.

foolip:
That would be great! @guidou, have you done any editing in this spec?

guidou:
No, I haven't edited this spec. 
I have reviewed a couple of changes related to things I an familiar with in Chromium.

stefhak:
As noted in #445 we should take this to the list.

ShijunS:
Could someone share any developer docs related to implementations in Chrome and Firefox? Maybe @jan-ivar, @guidou?  

I hope they are not implemented specific for WebRTC - the control knobs should be in WebRTC otherwise.  In addition, if you expect it will benefit WebAudio, it'd be helpful to get some feedback from some Audio WG experts.

jan-ivar:
@ShijunS Not sure what you mean with docs. It's implemented solely in gUM. Try it in Firefox [here](https://jsfiddle.net/jib1/c1pLd061/).

ShijunS:
@jan-ivar, I was hoping you'd have MDN pages to inform web developers how to use the constraints. For Edge, both AGC and NS are integrated internally with the RTC stack - it won't make sense for developers to turn them off. From FF perspective, what are the typical scenarios other than RTC where you would enable the functionality.  It'd be nice to get a couple examples of actual major websites doing that.  

Btw, I didn't see any signal from the jsfiddle test page while using Nightly, not sure what I have missed or should expect.

jan-ivar:
[...]

PR #445, opened 2017-04-11
Add autoGainControl and noiseSuppression.

jan-ivar:
Fix for https://github.com/w3c/mediacapture-main/issues/444.


stefhak:
This seems reasonable, but I think we should discuss on the list first to make sure no one has problems with adding these. @jan-ivar can you ask the list?

stefhak:
The list has been consulted, but no responses. I take that as at least people are not against adding this.

PR #446, opened 2017-04-12
Specify how applyConstraints() should behave without the argument

### LINKED ISSUE #438: What does it mean that the argument to applyConstraints() is optional
Constrainable Pattern method `applyConstraints()` (defined [here](https://w3c.github.io/mediacapture-main/#media-stream-track-interface-definition) and [here](https://w3c.github.io/mediacapture-main/#interface-definition) (*)) takes an `optional {MediaTrack}Constraints constraints`, but what does it mean to call `applyConstraints()` with no argument? 
Suggest removing the `optional`.

FTR [`Gecko implements it like that`](https://github.com/mozilla/gecko-dev/blob/72823b1e95bed9d3edaff79ebd64c391a33e6d7b/dom/webidl/MediaStreamTrack.webidl#L91) whereas Chrome doesn't have it yet (I'm landing it).

(*) links are broken: I recommend migrating to bikeshed.

adam-be:
Fixes #438

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -4432,7 +4432,7 @@ interface ConstrainablePattern {
                   <p>Let <var>p</var> be a new promise.</p>
                 </li>
                 <li>
-                  <p>Let <var>newContraints</var> be the argument to this
+                  <p>Let <var>newConstraints</var> be the argument to this
                   function, or an empty dictionary if the argument was
                   omitted.</p>
                 </li>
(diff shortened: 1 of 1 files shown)


aboba:
Should "newContraints" be "newConstraints"?

jan-ivar:
This should be reverted. This language is redundant, as WebIDL bindings ensure that an empty dictionary is present. It's invariant by design. I've filed https://github.com/w3c/mediacapture-main/issues/458.

PR #450, opened 2017-05-11
MediaStream: allow construction with given id

fippo:
allows MediaStreams to be created with a specific id, similar
to how it can be done for a MediaStreamTrack.

fixes https://github.com/w3c/webrtc-pc/issues/1196

Text is taken from https://w3c.github.io/mediacapture-main/#dom-mediastreamtrack-id with "MediaStreamTrack" replaced by "MediaStream" and "track id" by "stream id"

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -484,12 +484,13 @@ interface MediaStream : EventTarget {
             <dt><code>id</code> of type <span class=
             "idlAttrType"><a>DOMString</a></span>, readonly</dt>
             <dd>
-              <p>Unless a <code><a>MediaStream</a></code> object is created as 
-              a part of a special purpose algorithm that specifies how the 
-              stream id must be initialized, the User Agent MUST generate an 
-              identifier string, and MUST initialize the object's <code><a href=
-              "#dom-mediastream-id">id</a></code> attribute to that string. A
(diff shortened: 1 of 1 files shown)


stefhak:
LGTM.

getusermedia.html
```diff
@@ -484,9 +484,10 @@
...
             "idlAttrType"><a>DOMString</a></span>, readonly</dt>
             <dd>
-              <p>When a <code><a>MediaStream</a></code> object is created, the
-              User Agent MUST generate an identifier string, and MUST
-              initialize the object's <code><a href=
+              <p>Unless a <code><a>MediaStream</a></code> object is created as 
+              a part of a special purpose algorithm that specifies how the 
+              stream id must be initialized, the User Agent MUST generate an 
+              identifier string, and MUST initialize the object's <code><a href=
               "#dom-mediastream-id">id</a></code> attribute to that string. A
```
jan-ivar: Could we add back "...when it is created" somehow? This is in the getter after all.

fippo: will do (in another PR)

burnburn:
We will merge when we see the new PR from @fippo .

fippo:
thanks for the reminder -- pushed as separate commit in this PR to allow squashing before merge

PR #456, opened 2017-05-18
Add text about alternative ways to obtain local media

### LINKED ISSUE #448: Way to generate synthetic data for automated testing
Developers I were talking to want the API to support a way to generate synthetic traffic for testing on machines that do not have a microphone or camera.  (Firefox for example has the fake:true constraint).

adam-be:
Fixes #448 

Adds informative reference to "Media Capture from DOM Elements" spec


fluffy:
Thanks - looks good.

Issue #454, opened 2017-05-17
Should we be clearer about what's legacy?


jan-ivar:
The WebRTC spec has a section called [4.3.3 Legacy Interface Extensions](http://w3c.github.io/webrtc-pc/#legacy-interface-extensions). This seems informative.

Would the mediacapture spec benefit from similar language?

Right now, the word "legacy" only appears in a note in [10.1 NavigatorUserMedia Interface Extensions](https://w3c.github.io/mediacapture-main/getusermedia.html#navigatorusermedia-interface-extensions).

Should we perhaps rename it to "10.1 Legacy Interface Extensions"?

stefhak:
This proposal makes sense to me.

stefhak:
@jan-ivar can you do a PR?

jan-ivar:
@stefhak Done.

PR #461, opened 2017-05-26
Rename 10.1. to Legacy Interface Extensions.

jan-ivar:
Fix for https://github.com/w3c/mediacapture-main/issues/454.


stefhak:
LGTM.

fluffy:
seems like this bug should be closed given it was fixed.

Issue #472, opened 2017-08-04
How to implement web-compatible camera downscaling?


jan-ivar:
The spec leaves it to UAs whether to offer downscaled resolutions and decimated frame-rates.

This leaves web developers with a problem. This:
```js
await navigator.mediaDevices.getUserMedia({video: {height: {exact: 360}}});
```

...means, depending on browser¹, either:
 1. Do you have a 360 native mode? If not, `OverConstrainedError` please, so I can do som'else, or
 2. Downscale the camera output to 360p, please.

Assuming both groups of users exist, how do we solve their respective problems in all browsers?

---
<sub>1. Firefox and Edge implement 1, while Chrome (even Canary w/new algorithm flag) implements 2. Safari didn't take constraints yet.</sub>

jan-ivar:
Observation: The 2nd group has no need for `min`, `max`, `exact`, `ideal`, or `advanced`:
```js
await navigator.mediaDevices.getUserMedia({video: {height: 360}});
```
...is all they need. An inherently simple rescale-me-this API.

### Proposal
Give them that API, and force the 1st group to use `min`, `max`, `exact`, `ideal`, or `advanced`.

That is, have the presence of those keywords be a hint to ignore downscaled [settings dictionaries](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-settings-dictionary).

This would mean:
 * Only the group that needs them need learn the algorithm-related keywords.
 * `360` would downscale as needed (even in Firefox/Edge).
 * `{ideal: 360}` would find the closest native resolution (even in Chrome).
 * `{exact: 360}` would produce 360p native or `OverconstrainedError` (even in Chrome).
 * No downscaling in `advanced` (even in Chrome).
 * Should work for `frameRate` as well.

Note that this wouldn't otherwise change the meaning of plain values.

#### Examples
```js
await track.applyConstraints({height: 360}); // 360p downscaled if necessary
await track.applyConstraints({height: {ideal: 360}}); // closest native to 360p
await track.applyConstraints({height: {exact: 360}}); // native 360p or throw
await track.applyConstraints({height: {min: 320, max: 1920}}); // native in range or throw
await track.applyConstraints({height: {min: 320, max: 1920, ideal: 320}}); // lowest native
await track.applyConstraints({height: {max: 1920, ideal: 1920}}); // highest native
await track.applyConstraints({height: 9999}); // highest native available
```

[...]

PR #473, opened 2017-08-06
Omit downscaling/decimating for non-bare values.

jan-ivar:
Example fix for https://github.com/w3c/mediacapture-main/issues/472.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -4394,6 +4394,12 @@ interface ConstrainablePattern {
                   fitness distance is 0.</p>
                 </li>
                 <li>
+                  <p>If the constraint is not a bare value (rather 'min', 'max',
+                  'exact' or 'ideal'), and the <a>settings dictionary</a> relies
+                  on the User Agent downscaling or decimating the device output,
+                  the fitness distance SHOULD be positive infinity.</p>
+                </li>
+                <li>
                   <p>If the constraint is required ('min', 'max', or 'exact'),
                   and the <a>settings dictionary</a>'s value for the constraint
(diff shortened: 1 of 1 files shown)


alvestrand:
If accepted, this would mean that it's not possible to use constraints to invoke the downscaling function.
At the moment, that's something that works (and, I believe, is heavily used) in Chrome.

jan-ivar:
@alvestrand With this idea, you'd still be able to invoke downscaling with bare value constraints:
```js
await track.applyConstraints({width: 640, height: 360}); // would downscale
await track.applyConstraints({width: 640, height: {max: 360}}); // would NOT downscale
```
...on the idea that `min`,`max` etc. are redundant when you know what you want.

In any case, this was just an idea I had about moving Chrome and Firefox behaviors closer together in https://github.com/w3c/mediacapture-main/issues/472#issuecomment-320484268. @guidou and I did not reach any consensus on that, so let me close it. We can re-open if there's renewed interest. The idea that bare values turn on downscaling is arbitrary perhaps.

The [current idea we're working on](https://bugzilla.mozilla.org/show_bug.cgi?id=1388667#c10)  in Firefox is to downscale only where we'd throw `OverconstrainedError` otherwise, keeping everything else the same. This seemed like a reasonably intuitive way to not disrupt discovery of modes (which is our interpretation of spec intent), e.g. people can use `exact` or `max` to force downscaling.

However, while that seems like a reasonable interpretation of the spec to us, it does little to bridge differences between Chrome and Firefox, since bare values like
```js
await track.applyConstraints({width: 640, height: 360});
```
...would continue to downscale in Chrome but not in Firefox.

PR #474, opened 2017-08-07
Replace serializer with toJSON

### LINKED ISSUE #469: WebIDL serializer has been deprecated in favor of toJSON operation
Hi!

We recently [deprecated WebIDL serializers][0]. You can now directly specify [toJSON operations][1] instead, which you previously weren't allowed to do.

To deal with common cases, we added a new [\[Default\] extended attribute][2] which triggers the [default toJSON operation][3] that behaves similarly to how `serializers={attributes}` or `serializers={attributes, inherit}` used to. That is, it serializes all attributes that are of a [JSON type][4] into a vanilla JSON object.

It seems only the following interface in this spec is impacted by this change:

- [ ] [`MediaDeviceInfo`](https://w3c.github.io/mediacapture-main/getusermedia.html#idl-def-mediadeviceinfo)

It's a good candidate for the [default toJSON operation][3], so the below should be all you need:

```webidl
[Exposed=Window]
interface MediaDeviceInfo {
    readonly attribute DOMString       deviceId;
    readonly attribute MediaDeviceKind kind;
    readonly attribute DOMString       label;
    readonly attribute DOMString       groupId;
    [Default] object toJSON();
};
```

I'm sorry for the inconvenience this causes, but our hope is that this ultimately makes things a lot simpler and clearer for everybody.

Please feel free to reach out if you have any questions.

Thanks!

[0]: https://github.com/heycam/webidl/commit/a505f33
[1]: https://heycam.github.io/webidl/#idl-tojson-operation
[2]: https://heycam.github.io/webidl/#Default
[3]: https://heycam.github.io/webidl/#es-default-tojson
[4]: https://heycam.github.io/webidl/#dfn-json-types

dontcallmedom:
close #469


aboba:
@dontcallmedom When we applied a similar change in WebRTC-PC and it now generates a warning (see: https://rawgit.com/w3c/webrtc-pc/master/webrtc.html).  Is that issue likely to come up with this PR?

dontcallmedom:
yes (unless we upgrade respec at the same time)

adam-be:
Before the last webrtc-pc release, I manually added a description for the default toJSON method to get rid of the warning. I borrowed the of text from [1]. We could do the same thing here and remove the manually added text when respec generates it automatically.

[1] https://github.com/w3c/respec/issues/1309

PR #479, opened 2017-08-25
Upgrade our references to HTML5.2

dontcallmedom:
Use new data-cite mechanism from respec to avoid outdated links
HTML5.2 now has 'allowed to use' definition, so using it instead of HTML LS
Remove note about referencing HTML5.1 for  as no longer relevant
Remove note about HTML Media Controllers (removed from HTML spec)


adam-be:
Tested locally. WFM

PR #485, opened 2017-09-15
Add promise terminology

adam-be:


adam-be:
We might want to reference the promise writing guide as well. It defines the actions like "resolve" and "reject" in the Shorthand Phrases section. There's a note saying that that section should go into WebIDL eventually.

[1] https://www.w3.org/2001/tag/doc/promises-guide#shorthand-phrases

Issue #488, opened 2017-10-07
Touching a browsing context in an "in parallel" section is not OK.


jan-ivar:
Some feedback from our DOM team on the [getUserMedia](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-getusermedia()) algorithm's handling of `originIdentifier`:

1)  A browsing context does not have a concept of "origin".
2)  Touching a browsing context in an "in parallel" section is not OK.
3)  The top-level browsing context may not have anything to do with the API being invoked.

Basically, this algorithm needs to capture the origin from a sane place before going parallel.  Right now it's not doing any of those things.

PR #490, opened 2017-10-07
Move originIdentifier out of in-parallel steps.

jan-ivar:
Fix for #488.


jan-ivar:
@bzbarsky PTAL.

bzbarsky:
Looks reasonable, assuming you want current settings, not relevant settings.

jan-ivar:
@bzbarsky The algorithm was already using current settings, so I stuck with that. If we want to change to use relevant settings, we should probably do it consistently, and in a separate issue.

Issue #489, opened 2017-10-07
Disallow getUserMedia in non-fully-active documents.


jan-ivar:
To avoid some sharp edges in the the [getUserMedia](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-getusermedia()) algorithm, we should check that the relevant document is fully active. See https://github.com/w3c/permissions/issues/162.

PR #491, opened 2017-10-07
Throw on getUserMedia in non-fully active documents.

jan-ivar:
Fix for #489.


jan-ivar:
@bzbarsky, @adam-be PTAL.

bzbarsky:
Looks sane.

adam-be:
:+1:

Issue #487, opened 2017-09-22
Link is to wrong permission algorithm (editorial)


jan-ivar:
See https://github.com/w3c/permissions/pull/159#discussion_r140592752. The correct algorithm to link to is [request permission to use](https://w3c.github.io/permissions/#request-permission-to-use), which is the one returning `"granted"` or `"denied"`.

The other one is going away.

jan-ivar:
We should probably give UAs a choice here to use either [request permission to use](https://w3c.github.io/permissions/#request-permission-to-use) or [prompt the user to choose](https://w3c.github.io/permissions/#prompt-the-user-to-choose), to allow Firefox's permission prompt. However, while trying to write a PR for this I'm running into trouble using the latter. Will open an issue on permissions.

PR #492, opened 2017-10-07
Fix permission-to-use link and language.

jan-ivar:
Fix for https://github.com/w3c/mediacapture-main/issues/487.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3477,8 +3477,8 @@ interface MediaDeviceInfo {
                       <p>For the origin identified by
                       <var>originIdentifier</var>,
                       <a>request permission to use</a> a PermissionDescriptor
-                      with its name member set to either "camera" or
-                      "microphone" depending on <var>kind</var>, and,
+                      with its name member set to the permission name associated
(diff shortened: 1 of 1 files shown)


alvestrand:
Note (from Dan): Since camera & microphone are listed by name, this item needs to be extended if new types are added.

getusermedia.html
```diff
@@ -3477,8 +3475,12 @@
                     </li>
                     <li>
                       <p>For the origin identified by
-                      <var>originIdentifier</var>, <a>request permission</a>
-                      for use of the devices, while considering all devices
+                      <var>originIdentifier</var>,
+                      <a>request permission to use</a> a PermissionDescriptor
+                      with its name member set to either "camera" or
+                      "microphone" depending on <var>kind</var>, and,
+                      optionally, its deviceId member set to the device's
```
burnburn: Since you are specifically listing camera and microphone here, please update the extension suggestions as well so extenders of the spec will know to extend this as well.

jan-ivar:
@burnburn Done. I also tweaked the sentence to be consistent with the other places we reference `"camera"` and `"microphone"` in this spec.

PR #493, opened 2017-10-19
Add media constraint property disableInputProcessing (issue #457)

### LINKED ISSUE #457: Specification to capture unmodified audio
Current versions of Firefox (53.0.3) and Chrome (58.0.3029.110) preprocess audio capture streams for WebRTC even if the echoCancellation constraint property is set to false. I do not know what the preprocessing exactly does, but it adds some artifacts to the stream. And if we capture stereo the input streams are mixed together to mono,  preprocessed and the processed stream is duplicated. The recorded channels are always identical.  I could not find a way to capture unmodified audio using Firefox or Chrome, but this is required for some kind of applications. In our case we record speech for phonetical analysis. With this audio artifacts the recordings are not appropriate for our needs. The preprocessing is also problematic for music recording apps.
Surprisingly Microsoft Edge is the only browser I've tested, which does no audio preprocessing.

I think that Media Streams API should specify that implementations should not preprocess audio if echoCancellation is set to false.

klausj:
Fixes #457 .

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -446,10 +446,10 @@ interface MediaStream : EventTarget {
             "idlAttrType"><a>DOMString</a></span>, readonly</dt>
             <dd>
               <p>When a <code><a>MediaStream</a></code> is created, the
-              User Agent MUST generate an identifier string, and MUST 
+              User Agent MUST generate an identifier string, and MUST
               initialize the object's <code><a href="#dom-mediastream-id">
-              id</a></code> attribute to that string, unless the object is 
-              created as part of a special purpose algorithm that specifies how the 
+              id</a></code> attribute to that string, unless the object is
+              created as part of a special purpose algorithm that specifies how the
               stream id must be initialized.
               A good practice is to use a UUID [[rfc4122]], which is 36
               characters long in its canonical form. To avoid fingerprinting,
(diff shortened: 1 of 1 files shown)


dontcallmedom:
thanks @klausj for the contribution, it is much appreciated!

Could you reach out by email to me (dom@w3.org) and my colleague @vivienlacourba (vivien@w3.org) to discuss the IPR context of  your submission?

getusermedia.html
```diff
@@ -1820,6 +1849,16 @@
...
             </tr>
+            <tr id="def-constraint-disableInputProcessing">
+              <td>disableInputProcessing</td>
+              <td><code><a>ConstrainBoolean</a></code></td>
+              <td>For some applications it is desirable to capture the original
+              signal from the input source and therefore to disable all input
+              processing. The user agent should not alter the signal. Echo
+              cancellation, auto gain control, noise suppression and maybe
+              additional input processing features must be turned off
+              or bypassed.</td>
```
alvestrand: This is not well defined. For audio, the "original" is air pressure variations; for video, the "original" is photons on a sensor.
Many devices also do things like preliminary compression (for instance HD cameras use Motion JPEG in order to not overflow USB 2 bandwidth); some of this can be turned off, some can't.
A term like "minimizeInputProcessing" may be more accurate - asking for the minimally changed signal from the appropriate level of the media stack.

burnburn:
Yes, thank you @klausj !  In addition to Harald's comments, I have a minor nit to ask -- can you please remove the spurious whitespace changes?  If you look at the 'Files changed' tab here in GitHub you can see some lines where the only difference is whitespace.  Undoing those changes would make merging easier.  If for some reason you can't do that let us know and we can handle it as editors, but it's usually better if the requester can do that.

klausj:
Thanks @burnburn. I think I unintentionally used an editor which removed the trailing spaces automatically. I've commited the changes to revert the whitespace differences.

getusermedia.html
klausj: I understand the term input source as the source provided by the operating system, which is already a digital representation. However, since the user agent may not be able to control preliminary processings, "minimizeInputProcessing" is a good suggestion.

alvestrand:
@klausj can you clear the PR situation?

klausj:
My organization is currently discussing a membership. But this may take a while. Maybe it is better to close the pull request for now and create a new one if the situation is cleared.

alvestrand:
Closing for now. Please revisit when IPR is clear.
(I may argue about the name at that time, but let's paint that bikeshed when we get to it.)

Issue #466, opened 2017-06-16
Questions about setting belonging to source in Section 3.


guidou:
I would like clarification for the following case:

Let's say the system has a single camera that can provide frames at 60Hz.
There are two getUserMedia calls. One requests an exact frame rate of 30Hz, and the other one requests an exact frame rate of 60Hz.

Chrome currently allows both calls to succeed. For the first track, Chrome drops every other frame, so the track has an effective frame rate of 30 Hz. For the second track no frames are dropped, so it has the requested 60Hz.

Section 3 says _"User Agent processing, such as downsampling, may be used to ensure that all tracks have appropriate media"_, so processing to get a 30Hz track seems compliant.

The question is what should getSettings return?
Chrome currently returns 30Hz for the first track and 60 Hz for the second track, which I think is what a user would expect.

However, the spec says that settings don't belong to the tracks, but to the source. More specifically, it says _"Sources have constrainable properties which have capabilities and settings. The constrainable properties are "owned" by the source and are common to any (multiple) tracks that happen to be using the same source (e.g., if two different track objects bound to the same source ask for the same capability or setting information, they will get back the same answer)."_

This part suggests that returning 30Hz and 60Hz could be a violation of the spec, unless one considers the source to be not the camera, but the processed 30Hz stream.

Interestingly, the definition of getSettings() says _"The getSettings() method returns the current settings of all the constrainable properties of the object, whether they are platform defaults or have been set by the applyConstraints algorithm."_, where _object_ is the track, so this part of the spec suggests that settings might belong to tracks after all.

So, should Chrome be "fixed" to return 60Hz for the track with 30Hz?
Note that this is relevant to other properties for which Chrome currently provides processing.

stefhak:
Assigning to Dan.

jan-ivar:
@guidou Good find, and in the terminology section! The contradicting statements again:

> "User Agent processing, such as downsampling, may be used to ensure that all tracks have appropriate media",

> "(e.g., if two different track objects bound to the same source ask for the same capability or setting information, they will get back the same answer)."

The second statement seems to assume device settings are hardware settings, and/or that hardware devices cannot do two things at once. That's artificially limiting for audio at least:
 * aec, noise-reduction, auto-gaincontrol are often not implemented at the device level.
 * low-latency, sample-rate, sample-size, channel-count are, but still trivial to decimate in software.

AFAIK there's no quality loss either. Different values per track makes sense to me here. @burnburn?

[...]

PR #511, opened 2018-03-08
Make different settings values per track possible.

jan-ivar:
Fix for https://github.com/w3c/mediacapture-main/issues/466.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -111,7 +111,8 @@
         happen to be using the same source, whereas settings MAY differ per
         track (e.g., if two different track objects bound to the same source
         query capability and settings information, they will get back the same
-        capabilities, but may get different settings that are tailored to them).
+        capabilities, but may get different settings that are tailored to
+        satisfy their individual constraints).
         </p>
       </dd>
       <dt>Setting (Source Setting)</dt>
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -99,17 +99,17 @@ <h2>Terminology</h2>
...
-        two different track objects bound to the same source ask for the same
-        capability or setting information, they will get back the same
-        answer).</p>
+        media that conforms to the constraints present on that track, to that
+        track. Multiple tracks can be attached to the same source. User Agent
+        processing, such as downsampling, MAY be used to ensure that all tracks
+        have appropriate media.</p>
+        <p>Sources have <code><a>capabilities</a></code>, which are manipulated
+        through constrainable properties on the track, and observed by
+        <code><a>settings</a></code> on the track. The constrainable properties
```
burnburn: Not sure I agree with changing the semantics here to say that sources have capabilities, and that those capabilities are manipulated via constrainable properties.  I still think the fundamental unit here is the property, not the capability.

alvestrand:
Needs more review.

jan-ivar:
OK this should hopefully be better. @burnburn PTAL.

burnburn:
@alvestrand editors agree you can merge if you are happy with this.

PR #515, opened 2018-03-15
Change types of long constraints to be unsigned and clamped

### LINKED ISSUE #384: Should we use [EnforceRange] on min/max in constraints?
I was trying to debug a test with this constraint:

navigator.getUserMedia({video: {width: {min:Infinity}}} ....)

It failed in an interesting way (so there is probably a bug), but I tried to figure out what SHOULD happen.

WebIDL: https://www.w3.org/TR/WebIDL/#es-long - short version: ToInt32(ToNumber(string))

Assuming that "Infinity" is spelled correctly, this is ToInt32(+inf). Follow link to ECMAScript.

http://es5.github.io/#x9.5 - If number is NaN, +0, −0, +∞, or −∞, return +0. (WAT)

So if it's just a very large number? "Let int32bit be posInt modulo 2^32"

Result: constraining to +inf is the same as constraining to 0. Constraining to 2^32+100 is the same as constraining to 100.

Would the behavior specified at https://www.w3.org/TR/WebIDL/#es-long for [EnforceRange] (which is "throw a TypeError") be a less surprising behavior for users?

### LINKED ISSUE #385: ConstrainLong only concern positive values
As far as I can tell, all the usage of the `ConstrainLong` types refers to values that can only be positive (`width`, `height`, `sampleSize`, `sampleRate`, `channelCount`).

Should we be using `unsigned long` rather than `long` then? (and probably rename it to `ConstrainUnsignedLong`

dontcallmedom:
close #384
close #385 

DO NOT MERGE for now since this needs a bug fix in respec

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1251,20 +1251,20 @@ interface MediaStreamTrack : EventTarget {
         appropriate type.</p>
         <div>
           <pre class="idl">dictionary MediaTrackCapabilities {
-             LongRange           width;
-             LongRange           height;
+             ULongRange           width;
+             ULongRange           height;
              DoubleRange         aspectRatio;
              DoubleRange         frameRate;
              sequence&lt;DOMString&gt; facingMode;
              sequence&lt;DOMString&gt; resizeMode;
              DoubleRange         volume;
-             LongRange           sampleRate;
-             LongRange           sampleSize;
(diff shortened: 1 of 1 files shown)


dontcallmedom:
Now with the respec bug fix in.

Issue #520, opened 2018-05-31
audioElement should ignore video tracks in a stream wrt ended.


jan-ivar:
Currently, our prose on HTMLMediaElement impact relies on the media stream [active](https://w3c.github.io/mediacapture-main/getusermedia.html#stream-active) attribute, which only really makes sense for video elements, which end when the total number of tracks go to zero.

For audio elements, this seems wrong. Audio elements cannot play video, and should arguably end when the total number of *audio* tracks goes to zero.

jan-ivar:
@burnburn can you take a look? Looks like I don't have access to assign or set flags like I do in webrtc.

jan-ivar:
FWIW here's current behavior: https://jsfiddle.net/jib1/vbf2Lqhc/

Firefox: false, false, false, true, ended, true, true (only ended once both tracks go away).
Chrome: false, false, false, false, false, false

eric-carlson:
FWIW, in Safari the current behavior is: false, false, false, false, ended, true, true

alvestrand:
Can this be closed based on #521 being merged?

Issue #518, opened 2018-05-25
Clarification needed on HTMLMediaElement with MediaStream ending playback


Pehrsons:
Spec says this on ending media elements:
> When the MediaStream state moves from the active to the inactive state, the User Agent MUST raise an ended event on the HTMLMediaElement and set its ended attribute to true.

This seems too simple as an HTMLMediaElement whose source reaches the end goes through [a lot more steps](https://html.spec.whatwg.org/multipage/media.html#reaches-the-end):

> When the current playback position reaches the end of the media resource when the direction of playback is forwards, then the user agent must follow these steps:
> 
> 1. If the media element has a loop attribute specified, then seek to the earliest possible position of the media resource and return.
> 2. As defined above, the ended IDL attribute starts returning true once the event loop returns to step 1.
> 3. Queue a task to run these steps:
>     1. Fire an event named timeupdate at the media element.
>     2. If the media element has ended playback, the direction of playback is forwards, and paused is false, then:
>         1. Set the paused attribute to true.
>         2. Fire an event named pause at the media element.
>         3. Take pending play promises and reject pending play promises with the result and an "AbortError" DOMException.
>     3. Fire an event named ended at the media element.

It seems preferable to me to refer to this algorithm instead of rolling our own.

Note that the HTMLMediaElement spec also lists [the conditions](https://html.spec.whatwg.org/multipage/media.html#ended-playback) for when a media element is said to have ended playback:

> A media element is said to have ended playback when:
> - The element's readyState attribute is HAVE_METADATA or greater, and
> - Either:
>   - The current playback position is the end of the media resource, and
>   - The direction of playback is forwards, and
>   - The media element does not have a loop attribute specified. 
> - Or:
>   - The current playback position is the earliest possible position, and
>   - The direction of playback is backwards. 
> 
> The ended attribute must return true if, the last time the event loop reached step 1, the media element had ended playback and the direction of playback was forwards, and false otherwise.

It seems to me that if we override the condition for when the media element has ended playback to support MediaStreams, we'd get the algorithm with events and state for free.

alvestrand:
Media elements fed from a MediaStream can't loop, can't go backwards, and can't (usefully) pause, so all the other conditions for "something different" to happen aren't really relevant.
But not overriding other specs is usually a Good Thing, so I wouldn't mind changing this.
The added actions if we change this are to fire timeupdate, set the paused attribute and to fire paused.

alvestrand:
[...]

PR #521, opened 2018-05-31
New audible corollary to active stream property, to end audio elements

jan-ivar:
Fix for https://github.com/w3c/mediacapture-main/issues/520.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -347,6 +347,14 @@
       "#track-ended">ended</a>. A <code><a>MediaStream</a></code> that does not
       have any tracks or only has tracks that are <a href=
       "#track-ended">ended</a> is <dfn id="stream-inactive">inactive</dfn>.</p>
+      <p>A <code><a>MediaStream</a></code> object is said to be <dfn id=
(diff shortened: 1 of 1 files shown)


jan-ivar:
Much of this language may end up different from https://github.com/w3c/mediacapture-main/issues/518 but I wanted to get this change in first.

jan-ivar:
@burnburn The Extensibility section on [adding a new media type](https://w3c.github.io/mediacapture-main/getusermedia.html#defining-a-new-media-type-beyond-the-existing-audio-and-video-types) already calls out:

 * *updating how the HTMLMediaElement works with a MediaStream containing a track of the new media type (see 6. [MediaStreams in Media Elements](https://w3c.github.io/mediacapture-main/getusermedia.html#mediastreams-in-media-elements)),*

... which points here. Is that sufficient?

alvestrand:
The difference in behavior is that if you feed a video track (only) to an <audio> element, it should be considered "ended".
Do we have a Web Platform Test that will exercise this behavior?

Issue #525, opened 2018-07-09
Clarify constraints processing model wrt dictionary member presence


jan-ivar:
As part of new WebIDL considerations, I got alerted that the mediacapture spec is a bit unclear in its dealings with dictionary members and their [presence](https://heycam.github.io/webidl/#dfn-present) in its constraints processing, particularly its [fitness distance](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-fitness-distance) algorithm.

MediaTrackConstraintSet dictionary defines a bunch of members that are unions including dictionaries; these used to get defaulted to "empty dictionary" but with https://github.com/heycam/webidl/issues/76 they may no longer not get defaulted at all.

Specifically:
 1. With _"for each constraint provided for a constraint name in CS"_, this likely implies presence, but isn't clear. Moreover, all constraints use WebIDL unions, which today means they're always present.
 2. The sentence _"If an empty object or list has been given as the value for a constraint"_, misses the fact that we're operating on WebIDL invariants here, not JS objects. There is no "object" by the time you get to this code. There's whatever the output of IDL argument conversion was; in this case a dictionary or one of the other types constraints can have.

The good news is that AFAICT the fitness distance algorithm is entirely presence agnostic, caring only about values and keywords it finds. So I'm pushing a PR to clarify this.

jan-ivar:
Hit submit by accident. I'm cleaning up by editing initial comment now.

PR #526, opened 2018-07-09
Clarify+make fitness distance algorithm agnostic to dictionary member presence.

jan-ivar:
Fix for https://github.com/w3c/mediacapture-main/issues/525. @bzbarsky PTAL.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -4114,15 +4114,18 @@ interface ConstrainablePattern {
               value being compared against.</p>
               <p>We define the <dfn>fitness distance</dfn> between a
               <a>settings dictionary</a> and a constraint set <var>CS</var> as
-              the sum, for each constraint provided for a constraint name in
+              the sum, for each member (represented by a
+              <var>constraintName</var> and <var>constraintValue</var> pair) in
               <var>CS</var>, of the following values:</p>
               <ol>
                 <li>
-                  <p>If the constraint is not supported by the browser, the
-                  fitness distance is 0.</p>
+                  <p>If <var>constraintName</var> is not supported by the
+                  browser, the fitness distance is 0.</p>
                 </li>
                 <li>
-                  <p>If the constraint is required ('min', 'max', or 'exact'),
+                  <p>If the constraint is required (<var>constraintValue</var> either
+                  contains one or more members named 'min', 'max', or 'exact', or is
+                  itself a bare value and bare values are to be treated as 'exact'),
                   and the <a>settings dictionary</a>'s value for the constraint
                   does not satisfy the constraint, the fitness distance is
                   positive infinity.</p>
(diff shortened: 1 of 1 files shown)


jan-ivar:
Ugh, rebase accident. One sec.

getusermedia.html
```diff
@@ -4132,20 +4135,20 @@ <h2>Methods</h2>
                   this type of device, the fitness distance is 0 (that is, the
                   constraint does not influence the fitness distance).</p>
                 </li>
-                <li>If no ideal value is specified, the fitness distance is
-                0.</li>
+                <li>If no ideal value is specified (<var>constraintValue</var>
+                  either contains no member named 'ideal', or is itself a bare
+                  value and bare values are to be treated as 'ideal'), the
```
alvestrand: grammar error here - the second part is in the positive tense and should have been in the negative tense - may be better said as "either contains no member named 'ideal' or, if bare values are to be treated as 'ideal', isn't a bare value".

jan-ivar:
@alvestrand Had to add more. Can you take another look? This is to allow https://github.com/heycam/webidl/issues/76 removing the invariant that member dictionary members are always present.

This means we have to test for `undefined` (using [present](https://heycam.github.io/webidl/#dfn-present)) and `null` because unions, as I understand it. @bzbarsky ptal.

bzbarsky:
> This means we have to test for undefined (using present) 

Correct.

> and null because unions

What?  Why?

jan-ivar:
@bzbarsky You're right, you explained to me that null is  a valid value for an IDL dictionary type and gets converted to empty dictionary. Will fix.

getusermedia.html
```diff
@@ -4114,15 +4114,19 @@ <h2>Methods</h2>
               value being compared against.</p>
               <p>We define the <dfn>fitness distance</dfn> between a
               <a>settings dictionary</a> and a constraint set <var>CS</var> as
```
bzbarsky: So a constraint set is a webidl dictionary, right?  That's not really specified anywhere.

Also, https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-selectsettings should make it clear that it has a constraint set argument, not an ambient constraint set.

getusermedia.html
```diff
@@ -4114,15 +4114,19 @@ <h2>Methods</h2>
...
-              the sum, for each constraint provided for a constraint name in
+              the sum, for each member (represented by a
+              <var>constraintName</var> and <var>constraintValue</var> pair)
+              <a href="https://heycam.github.io/webidl/#dfn-present">present</a> in
               <var>CS</var>, of the following values:</p>
               <ol>
                 <li>
-                  <p>If the constraint is not supported by the browser, the
-                  fitness distance is 0.</p>
+                  <p>If <var>constraintName</var> is not supported by the
```
bzbarsky: So this is a little odd, actually.

The input to this thing is a WebIDL dictionary, right?  If a browser doesn't support some random spec that defines some constraint, it wouldn't know to include that name in the dictionary to start with, so it would not trigger this clause.

Is this clause meant to just apply to constraints that the browser knows about in the sense of having them in its IDL but doesn't actually implement matching for?  Or is it meant to apply to constraints that the browser really doesn't know about?  In the latter case, the input here obviously can't be a dictionary, because those presuppose a fixed set of known members and ignore everything else.

What do UAs actually implement here?

getusermedia.html
```diff
@@ -4132,20 +4136,20 @@ <h2>Methods</h2>
                   this type of device, the fitness distance is 0 (that is, the
                   constraint does not influence the fitness distance).</p>
                 </li>
-                <li>If no ideal value is specified, the fitness distance is
-                0.</li>
+                <li>If no ideal value is specified (<var>constraintValue</var>
+                  either contains no member named 'ideal', or, if bare values
+                  are to be treated as 'ideal', isn't a bare value), the
```
bzbarsky: This looks wrong to me.  It's saying that if bare values are to be treated as ideal and constraintValue is not a bare value then the ideal value is not specified.  That's clearly not the intent, but that's what it says.

I would strongly urge clarity of processing model over brevity here.  If we have an "ideal value is specified" boolean with a complicated way to compute it, we should compute it in some substeps instead of trying to jam the computation into a parenthetical.  Code that does that  is not considered a good idea; neither are specs that do it.

getusermedia.html
```diff
@@ -4132,20 +4136,20 @@ <h2>Methods</h2>
...
+                  fitness distance is 0.</li>
                 <li>For all positive numeric non-required constraints (such as
```
bzbarsky: What does "numeric" mean here?  Is that a property a constraint can have as part of its definition (in which case please link to the concept's definition) or something else?

getusermedia.html
```diff
@@ -4185,7 +4189,7 @@ <h2>Methods</h2>
                 <li>Each <a data-lt="constraints">constraint</a> specifies one
                 or more values (or a range of values) for its property. A
                 property MAY appear more than once in the list of 'advanced'
-                ConstraintSets. If an empty object or list has been given as
+                ConstraintSets. If an empty list has been given as
```
bzbarsky: It's really not clear to me what cases this is supposed to be handling. Is this dealing with constraints that are sequence-valued?

PR #527, opened 2018-07-09
Fix links to SelectSettings and fitness distance algorithms (editorial).

jan-ivar:
Found while fixing https://github.com/w3c/mediacapture-main/issues/525.


getusermedia.html
```diff
@@ -4204,23 +4204,23 @@ <h2>Methods</h2>
                 removed.)</li>
                 <li>
                   <p>For every possible <a>settings dictionary</a> of
-                  <var>copy</var> compute its <a>fitness distance, treating
+                  <var>copy</var> compute its <a>fitness distance</a>, treating
                   bare values of properties as ideal values</a>. Let
```
alvestrand: Need to remove the </a> here.

Issue #534, opened 2018-08-30
Isolated Media Streams require forethought in gUM permission algorithm


jan-ivar:
From https://github.com/w3c/webrtc-pc/issues/1646#issuecomment-344984088:
> We also need to exclude isolated tracks in the following sentence (**e.g.**):
> 
> *"..., request permission for use of the devices, while considering all devices attached to a live* **non-isolated** *MediaStreamTrack in the current browsing context to have permission status "granted""*

PR #535, opened 2018-08-30
Limit gUM permission exemption to when there are non-isolated live tracks.

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/534.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3600,11 +3600,16 @@ interface MediaDeviceInfo {
                       (e.g. "camera", "microphone"), and, optionally, its
                       deviceId member set to the device's <var>deviceId</var>,
                       while considering all devices attached to a
-                      <code>live</code> MediaStreamTrack in the current
+                      <code>live</code> and <a>same-permission</a>
+                      MediaStreamTrack in the current
                       <a data-cite=
                       "!HTML52/browsers.html#browsing-context">browsing
                       context</a> to have permission status "granted",
-                      resulting in a set of provided media.</p>
+                      resulting in a set of provided media.
+                      <dfn>Same-permission</dfn> in this context means a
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -3600,11 +3600,16 @@ <h2>Methods</h2>
...
+                      MediaStreamTrack in the current
                       <a data-cite=
                       "!HTML52/browsers.html#browsing-context">browsing
                       context</a> to have permission status "granted",
-                      resulting in a set of provided media.</p>
+                      resulting in a set of provided media.
+                      <dfn>Same-permission</dfn> in this context means a
+                      MediaStreamTrack that required the same level of
+                      permission to obtain as what is being requested (e.g. not
+                      isolated).</p>
```
martinthomson: This is tricky and while this is a good start, we need some better definition here if this is the direction we want to go (but see below).  If you want to do this, you need to say condition this on the site having permission to access that source, but only if existing access is based on a permission grant that is as permissive or *more* permissive than the request.  Permissions are granted based on the type of source, the specific instance of a source, and any isolation properties.  For example, a request for isolated media from a source that has a live, non-isolated track would be granted automatically, but a request for isolated media from a source that has a live, isolated track with a different principal would not be granted.

We could punt on this entirely and let the browsers sort this out.  Maybe with some suggestions along the above lines.  That is, browsers would need to build their own heuristics about automatic grants might look like.  I don't think that's a good idea though; having reliable and predictable behaviour around the use and reuse of existing sources is a good thing.

jan-ivar: I think it's an improvement as well. I wanted to protect existing gUM permissions from new kinds of permissions. I hope to leave to webrtc-identity to define permissions for isolated streams.

PR #542, opened 2018-10-03
Clarify what happens to non-applicable constraints.

### LINKED ISSUE #533: What happens to constraints that are not applicable? Ignored or OverconstrainedError?
WebIDL automatically "filters out" any constraints (members in a dictionary) passed in to getUserMedia() that are not defined in the WebIDL, which means that typos or unrecognized constraints by an implementation would silently be ignored.

However, any constraints that are defined are passed in, regardless if they are applicable or not. I don't know if it's clear what to do in this case, for example if passing in audio-only constraints to a video-only GUM-request. Would the fitness distance for the audio constraint be infinity making getUserMedia() throw an OverconstrainedError, or should it be ignored?

If fitness distance is infinity, does that mean that the following will throw an exception...
```
getUserMedia({video:{volume:{exact:0.5}}})
```
... but the following would pass?
```
getUserMedia({video:{volume:{ideal:0.5}}})
```
Also since this would cause the same "damage" to the fitness distance to ALL sources, are the sources prioritized correctly despite of this or are all of them "infinity" and handled as if they all have the same fitness distance (meaning other constraints would in practice be ignored?)

@guidou Does the GUM algorithm need to be clarified?

henbos:
Fixes #533.

Some of the constrainable properties are said to be "defined to apply only to [video/audio] MediaStreamTrack objects". It is not clear whether specifying a constraint that doesn't apply for the type (e.g. {audio:{exact:640}}) would cause an OverconstrainedError or simply be ignored. Both are valid interpretations.

This PR says to ignore such constraints, because...
- This is what current implementations do.
- This requires less code to be updated when new constraints are added.
- If a constraint is not recognized by a User Agent it is ignored (because constraints are passed as dictionaries, anything not in the IDL would be ignored); this is consistent with "constraint is not implemented for this type".
- This would allow editorial changes such as splitting audio and video constrainable properties into separate dictionaries without implicitly causing change in behavior.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3535,11 +3535,22 @@ interface MediaDeviceInfo {
                           <code><a>name</a></code> attribute has the value
                           <code>NotFoundError</code> and abort these steps.</p>
                         </li>
-                        <li>If the value of the <var>T</var> entry of
-                        <var>constraints</var> is "true", set CS to the empty
-                        constraint set (no constraint). Otherwise, continue
-                        with <var>CS</var> set to the value of the <var>T</var>
-                        entry of <var>constraints</var>.</li>
+                        <li>
+                          If the value of the <var>T</var> entry of
+                          <var>constraints</var> is "true", set <var>CS</var> to
+                          the empty constraint set (no constraint). Otherwise,
+                          continue with <var>CS</var> set to the value of the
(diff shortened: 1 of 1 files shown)


alvestrand:
Tests needed.

henbos:
Filed #13363

PR #558, opened 2019-01-15
Add clarification for overconstrained constraints evaluation.

armax00:
The spec is currently unclear as of which constraints should be
evaluated when a new configuration is retrieved. We should consider
relevant for OverconstrainedErrorEvent only those changes that do not
comply with the basic constraints provided.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2754,6 +2754,9 @@ interface OverconstrainedErrorEvent : Event {
             the current constraints against a given source and is not able to
             configure the source within the limitations established by the
             intersection of imposed constraints.</p>
+            <p class="note">The aforementioned constraints SHOULD be interpreted
+            as basic constraints. This effectively means that the configuration
+            SHOULD be evaluated only against the provided basic constraints.</p>
             <p>Due to being over-constrained, the User Agent must mute each
             affected track.</p>
             <p>The affected track(s) will remain <a href=


alvestrand:
@jan-ivar has volunteered to suggest wordsmithing.

getusermedia.html
```diff
@@ -2754,6 +2754,9 @@ <h2>Event summary</h2>
             the current constraints against a given source and is not able to
             configure the source within the limitations established by the
             intersection of imposed constraints.</p>
+            <p class="note">The aforementioned constraints SHOULD be interpreted
+            as basic constraints. This effectively means that the configuration
+            SHOULD be evaluated only against the provided basic constraints.</p>
```
jan-ivar: `SHOULD` has specific normative meaning, which we want to avoid in notes. And not all basic constraints are required.

This is already poorly explained in two places, so rather than a new note, I vote we fix those:
1. We could fix the missing definition of *requiredConstraint* under [11.1](https://w3c.github.io/mediacapture-main/getusermedia.html#interface-definition).
2. The Constrainable section has terminology flaws under **Setting (Source Setting)**: *"A source's settings must always conform to the current set of basic (mandatory) constraints on all attached tracks. ... A User Agent attempts to ensure that sources adhere to advanced (optional) constraints as closely as possible"*

In short, not all basic constraints are "mandatory" (aka "required" which is the newer name for "mandatory"). 

Instead:

*"A source's settings must always conform to the current set of required basic constraints on all attached tracks. ... A User Agent attempts to ensure that sources adhere to optional basic constraints, as well as advanced constraints, as closely as possible"*

armax00: Sounds good, and thanks for the clarification of the normative language in NOTICE sections.

With respect to the mandatory basic constraints, I intentionally used the language that way. While writing the code for the feature, I figured that if a change of parameters contradicts an ideal constraint, we can consider that as getUserMedia decided to skip the constraint and hence ignore it because of its ideal nature. This means, that in the context of OverconstrainedErrorEvent we should care only on exact, min, and max constraints, for what I can understand at least.

jan-ivar: @armax00 I'm also curious what constraints you plan on firing this event for, and what the use-case is. So far we've been ignoring it in Firefox.

armax00: > > we can consider to ... skip the constraint and hence ignore it because of its ideal nature
> 
> That's implementation talk.

Good point :) though I still do not quite get the separation, as I would have imagined that for compatibility across browsers there should be some prescriptions that guarantees consistency of behavior.

> In code written solely for the event, yes. More broadly, I'd expect this event to come from the browser code that reacts to device changes. In that context, I'd say it's an implementation choice whether to ignore optional constraints outright. The spec doesn't mandate what a user agent does _"within the constraints established by the application"_.
> 
> A different choice would be to rerun the applyConstraints algorithm. For most constraints we support today, there's probably little impact on fitness distance from runtime changes, and no difference.
> 

Right, but that would impact the whole set of settings applied to a device, as of how the values are computed and how each group of constraints is then accepted or rejected.

> @armax00 Yes that's the missing definition of _requiredConstraints_ under [11.1](https://w3c.github.io/mediacapture-main/getusermedia.html#interface-definition). The closest definition of "required" I found was _"all 'min', 'max', and 'exact' constraints in the basic Constraint structure are together treated as the 'required' set"_.
> 
> I can try a PR if you want. Let me know.

Sure, if you think that might make the process quicker go for it. I believe it might take quite some back and forth before I manage to come out with something acceptable :)

jan-ivar:
Closing in favor of https://github.com/w3c/mediacapture-main/pull/569.

Issue #562, opened 2019-01-24
What constraint name should be exposed in case of a getUserMedia query with multiple failing constraints


youennf:
Crafting a getUserMedia request that always fails allows guessing the capabilities of a system, in particular if the request has one constraint that is known to always fail and a second constraint that may or may not fail.

Some potential countermeasures:
- Not return the failing constraint name if getUserMedia is not already granted
- Return the constraint name that can never be met

jan-ivar:
> this requires to process all constraints

As does success. We're happy to implement this FWIW, and don't anticipate difficulty, as optimizing a failure path is of low value to us.

Compared to what's already missing for WPT, I think this change pales. If anything, it decreases implementation variance, simplifying tests.

Most importantly, I believe this PR is something editors can merge now without controversy.

> I would prefer the spec to allow returning an empty constraint name (in pre-grant case at least)

I think that's going to be harder to get through at this point (this spec is in CR). *OverconstrainedError* and the whole constraints language was designed for the pre-grant case. If you think we can get rid of it, I would personally support it, but also bet against it succeeding (have you seen [section 7.2](https://w3c.github.io/mediacapture-main/getusermedia.html#overconstrainederror-object)? someone must really want it) 😉

In any case, I don't see this PR as an obstacle to that, so in the interest of progress, I vote we merge it, and open a new issue for that.

youennf:
It makes sense to me for the WG to first decide whether filtering constraintName is feasible or not, and if so, as a MAY or a MUST.

> but also bet against it succeeding (have you seen [section 7.2](https://w3c.github.io/mediacapture-main/getusermedia.html#overconstrainederror-object)? someone must really want it) 😉

Let's try to see whether we can make it succeed!

It would be nice if the people supporting this error reporting could chime in and express whether such change would break their use cases and why it is important to have it pre-grant info compared to post-grant.

FWIW, I believe constraintName might not be sufficient as is.
It does not provide, on its own, sufficient information for widely used cases like getUserMedia({audio: true, video: true}) or  getUserMedia with both audio/video deviceId constraints.

I would think that in many cases, it is the combination of constraints that make the query fails.
Say I want a 4K/120fps camera, but my system only has 4K/60fps or HD/120fps. Whether we return frameRate or width does not provide enough information for the web app to reason about it.

> I think that's going to be harder to get through at this point (this spec is in CR).

Well, this spec is expected to go into another CR soon which means substantive changes have been made since the first CR.
[...]

Issue #561, opened 2019-01-24
enumerateDevices can be used to track user devices in background pages


youennf:
enumerateDevices currently use [storedDeviceList] to return the same list of devices as long as there is no devicechange event. devicechange events only fire when page has focus.
enumerateDevices can still be used on an iframe being reloaded every 1 second to gather the exact number of devices since [storedDeviceList] is initially null.

As per the spec, if the origin was granted once device access (say for a funny hat cool game), any page of that origin will be able to monitor in near realtime the exact list of devices including labels.

jan-ivar:
This circumvents the intent behind the [[storedDeviceList]], so I think we need to fix this.

jan-ivar:
I ran a couple ideas in my head:

If, [on pageload](https://w3c.github.io/mediacapture-main/getusermedia.html#mediadevices), the document is not fully active and in focus, then instead of `null`:
 1. Initialize [[storedDeviceList]] to an empty list.
 2. Initialize [[storedDeviceList]] to a previous in-memory cached storedDeviceList for this origin, or `null` if not found.
 3. Initialize [[storedDeviceList]] to an empty list if origin is found in in-memory cached list, or null.
 4. Initialize [[storedDeviceList]] from a global in-memory storedDeviceList cache that holds back updates until tabs receive focus.
 5. Defer resolving enumerateDevices() promise until document becomes fully active and in focus.

(1) might break legitimate WebRTC sites, e.g. restoring Firefox with pinned tabs after a restart.
(2) would mitigate (1), but requires some work, and is vulnerable to `1.evil.com`, `2.evil.com`.
(3) is a cheaper (2) that's meaner to background tabs that fall in the net.
(4) probably is the full airtight solution, but again probably requires some work, and hard to spec.
(5) would be a simple solution that side-steps the problem, but might depend on https://github.com/w3c/mediacapture-main/issues/560.

youennf:
There is the case of a page being loaded (so it has focus) that can instantiate a 1000 frames which will have focus at that time (hence with a [[storedDeviceList]] set as null). Each frame can then be used to call enumerateDevices at different times.

Options 1 to 4 can probably be specced so that at the first enumerateDevices call, in case [[storedDeviceList]] is null and page has not focus, do a specific handling like setting [[storedDeviceList]] to the empty list.

With that change, option 1 (which should also include triggering the devicechange event when getting focus) might be ok, and would be the occasion to have enumerateDevices return an empty array :)

[...]

Issue #559, opened 2019-01-18
Spec does no handle fingerprinting related to exposing non default capture devices


youennf:
The spec calls out that the number of devices is a potential fingerprint.
Most devices have a standard number of cameras and microphones that web sites can infer from information like navigator.userAgent.
Devices with additional cameras/microphones (say USB webcams) can be fingerprinted if enumerateDevices is exposing these additional cameras/microphones.

I wonder whether this specific fingerprint should be called out in the specification.
Countermeasures should be implementable/could be proposed.
In particular, implementations should be allowed to remove any device deemed appropriate when copying resultList in filteredList (step 2.6.2 of enumerateDevices).

jan-ivar:
As you mention, the spec already calls out the number of devices is a potential fingerprint. I believe the significance of outliers (fonts, dead pixels, uncommon configurations) is well known in the fingerprinting space, and can be assumed to have been understood the several times the fingerprinting surface of this API was discussed in the past, so I'm not sure this qualifies as new information.

FWIW, Firefox implements a `privacy.resistFingerprinting` preference used by the Tor browser (see [blog](https://www.ghacks.net/2018/03/01/a-history-of-fingerprinting-protection-in-firefox/) and [enumerateDevices specifics](https://bugzilla.mozilla.org/show_bug.cgi?id=1372073)), and was accomplished without changing or breaking the API.

youennf:
> As you mention, the spec already calls out the number of devices is a potential fingerprint. I believe the significance of outliers (fonts, dead pixels, uncommon configurations) is well known in the fingerprinting space

There are some evidence on the web that enumerateDevices is currently in use for fingerprinting purpose. If it seems useful for fingerprints, what is the benefit to allow this new fingerprinting vector and what do we forbid if we try removing this fingerprinting vector?

Two additional points:
1. Compared to other fingerprinting, devices might appear/disappear frequently and might be related to user activities.
2. It is not only about the number of devices, it is also about the three level of fingerprinting the current spec is defining:
a. never granted capture access: number of devices are exposed. can change over time.
b. granted capture access once: number of devices are exposed and individual device presence can be tracked.
c. capture is granted: everything is exposed.
The privacy section of the spec does not really points out the fingerprinting issues related to 'b', it only takes about 'a' and 'c'.

[...]

PR #564, opened 2019-01-24
Mitigate fingerprinting from OverconstrainedError in gUM().

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/562.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3556,7 +3556,7 @@ interface MediaDeviceInfo {
                           whose fitness distance was infinity for
                           all settings dictionaries examined while executing
                           the <a>SelectSettings</a> algorithm, and jump to the
-                          step labeled <em>Constraint Failure</em> below.</p>
+                          step labeled <em>Constraint Failure</em> below.
                           The user agent SHOULD choose the constraint that
                           minimizes the amount of fingerprinting information
                           given.</p>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3557,6 +3557,9 @@ interface MediaDeviceInfo {
                           all settings dictionaries examined while executing
                           the <a>SelectSettings</a> algorithm, and jump to the
                           step labeled <em>Constraint Failure</em> below.</p>
+                          The user agent SHOULD choose the constraint that
+                          minimizes the amount of fingerprinting information
+                          given.</p>
                           <p class="fingerprint">This error gives information
                           about what the underlying device is not capable of
                           producing, before the user has given any


youennf:
This PR tries to mitigate information leakage in the case enumerateDevices does not stop leaking information.

Even with this PR though, one could silently learn some information about the device setup that cannot be gathered from enumerateDevices.
For instance, by mixing width constraint with deviceId constraint, a script could potentially distinguish cases like 720p vs 1080p cameras with no prompt triggered:
- If constraintName is width (length of failedConstraints is 2), capture at that resolution is not supported.
- If constraintName is deviceId (length of failedConstraints is 1), capture at that resolution is supported.

The fingerprinting statement is good to have.
I wonder whether this is the error that silently gives fingerprint information or just failureConstraint. If it is the latter, we could be more explicit in the statement.

It would be nice to add some mitigation to this fingerprinting, for instance:
"User agents MAY tamper the value of failedConstraint as a fingerprinting mitigation".

jan-ivar:
> For instance, by mixing width constraint with deviceId constraint, a script could potentially distinguish cases like 720p vs 1080p cameras with no prompt triggered:

Great point! The "never return deviceId"-part of the PR is flawed. (link to [demo](https://jsfiddle.net/jib1/n0scjq18/) again).

> "User agents MAY tamper the value of failedConstraint as a fingerprinting mitigation".

I agree we should shoot for a more general statement like that. Not leaking information here may require UA smarts beyond what we want to write in stone in a spec.

However, we should probably be clear whether "tamper" extends to returning no name at all, which may be controversial (in today's spec,  `error.constraint == ""` means *"you're getting close"*, i.e. none of the required constraints used would fail by themselves), it's solely a "combination failure".

alvestrand:
@youennf, "failedConstraint" is a string, not an array. You can't take its length. (comment #2)

alvestrand:
I *still* don't get the point of Jan-Ivar's demo.
Yes, you can tell what devices are present or not.
That's what getMediaDevices() is intended to accomplish. We gave people an API to do this; there is no additional privacy concern with giving them information we've already given them.

What's the problem we're trying to address here?

jan-ivar:
@alvestrand Yes, in fact I originally wrote that demo to show @youennf *getUserMedia* leaks the same info as *enumerateDevices()* as [he wants to make enumerateDevices() more fingerprint-safe](https://github.com/w3c/mediacapture-main/issues/559#issuecomment-456598662) than today.

This [equivalence is a problem](https://github.com/w3c/mediacapture-main/issues/559#issuecomment-456990813) if we want to move the needle on fingerprinting. It may not seem like a problem if we want the spec to stay where it is... *except*:

It sometimes leaks more info than *enumerateDevices()*:

 1. My demo bypasses *enumerateDevices()'s* [[storedDeviceList]] and time-fuzzing mitigations in browsers that allow *getUserMedia* calls to fulfill in background-tabs without focus. I.e. a background tab can detect fairly accurately when a device is inserted or removed, and time-correlate this across domains, even between regular and private browsing, to identify someone on multiple sites as the same individual.

 2. As @youennf showed [above](https://github.com/w3c/mediacapture-main/pull/564#issuecomment-463712683), the same trick can be used to bypass the "empty dictionary" mitigation of [deviceInfo.getCapabilities()](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-inputdeviceinfo-getcapabilities), to leak detailed device information, like width, height and frame-rate, pre-initial-grant.

So I claim we need to address this and https://github.com/w3c/mediacapture-main/issues/561, to even keep the needle where we thought it was.

alvestrand:
Resolution: Change the text to say "browser SHOULD choose the constraint that minimizes the amount of fingerprinting information given, such as..."

jan-ivar:
Had to rebase.

youennf:
I wonder how this PR translates in terms of implementation.
If there is a query and one has two constraints, width and sampleRate, or width and audio deviceId, how am I supposed to know which one to pick?
Should the spec define the rules of which constraint is triggering higher fingerprinting?
How can we verify that implementation strategies cannot be abused?

WebKit is returning an empty string constraint when getUserMedia is not granted or site has no persistent access. This is a simple strategy which is AFAIK the most robust against fingerprinting.
If we cannot define a better strategy that is as robust and enables more use cases, I think the spec should allow implementations to pick this strategy.

I would tend to add to the PR the following sentence:
"Implementations MAY select no constraint to further limit fingerprinting risks".

> However, we should probably be clear whether "tamper" extends to returning no name at all, which may be controversial (in today's spec, `error.constraint == ""` means _"you're getting close"_, i.e. none of the required constraints used would fail by themselves), it's solely a "combination failure".

Is it controversial though?
I am not aware of any real usage to "you're getting close" or to failure stats.
I am not aware of queries used by websites for which the constraint information would provide really useful information.

FWIW, this constraint mechanism is not providing sufficient information to real cases like "I want to get the same mic and camera as last time": the returned constraint would be deviceId but the important information for an app would be to know whether the failure is for mic, camera or both.

jan-ivar:
Good points. It seems hard to implement a worthwhile strategy that couldn't be exploited somehow, let alone spec one in stone.

alvestrand:
This direction is OBE, so closing.

PR #569, opened 2019-02-14
Define the three collective forms of constraints, to clarify prose (Editorial).

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/pull/558. @alvestrand @henbos PTAL.

[Preview](https://rawgit.com/w3c/mediacapture-main/5fa79d737369b892e981e5ffddac7e0ab20e9dbd/getusermedia.html#dfn-required-constraints)

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -182,16 +182,18 @@
         <p>For each constrainable property, a constraint exists whose name
         corresponds with the relevant source setting name and capability
         name.</p>
-        <p>A constraint takes one of three forms, depending on the syntax used.
-        Collectively, we refer to all constraints that apply on a track for a
-        type of device, of each form, in three groups as follows:
-        <dfn>required constraints</dfn> are all <a data-lt=
+        <p>A constraint falls into one of three groups, depending on its place
+        in the constraints structure. The groups are:
+        <ul>
+          <li><dfn>required constraints</dfn> are all <a data-lt=
           "advanced constraints">non-advanced</a> constraints that are
-        <a>required</a>; <dfn>optional basic constraints</dfn> are the remaining
-        <a data-lt="advanced constraints">non-advanced</a> constraints; and
-        <dfn>advanced constraints</dfn> are all constraints specified
+        <a>required</a>.</li>
+          <li><dfn>optional basic constraints</dfn> are the remaining
+        <a data-lt="advanced constraints">non-advanced</a> constraints.</li>
+          <li><dfn>advanced constraints</dfn> are all constraints specified
         using the <code><a href="#dom-constraints-advanced">advanced</a></code>
-        keyword.</p>
+        keyword.</li>
+        </ul></p>
         <p>In general, User Agents will have more flexibility to optimize the
         media streaming experience the fewer constraints are applied, so
         application authors are strongly encouraged to use <a>required
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -121,12 +121,12 @@
         read-only.</p>
         <p>A source's settings can change dynamically over time due to
         environmental conditions, sink configurations, or constraint changes. A
-        source's settings must always conform to the current set of basic
-        (mandatory) constraints on all attached tracks. A source that cannot
+        source's settings must always conform to the current set of
+        <a>required constraints</a> on all attached tracks. A source that cannot
         conform to these constraints causes affected tracks to become
         <a>overconstrained</a> and therefore <a href="#track-muted">muted</a>.
         A <a>User Agent</a> attempts to ensure that sources adhere to advanced
-        (optional) constraints as closely as possible, see <a href=
+        constraints as closely as possible, see <a href=
         "#constrainable-interface"></a>.</p>
         <p>Although settings are a property of the source, they are only
         exposed to the application through the tracks attached to the source.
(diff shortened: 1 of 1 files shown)


henbos:
@armax can you take a look? This is @jan-ivar attempt to tackle the same issue you did

getusermedia.html
```diff
@@ -179,13 +179,23 @@ <h2>Terminology</h2>
...
-        constraints sparingly.</p>
         <p>For each constrainable property, a constraint exists whose name
         corresponds with the relevant source setting name and capability
         name.</p>
+        <p>A constraint takes one of three forms, depending on the syntax used.
+        Collectively, we refer to all constraints that apply on a track for a
+        type of device, of each form, in three groups as follows:
+        <dfn>required constraints</dfn> are all <a data-lt=
+          "advanced constraints">non-advanced</a> constraints that are
+        <a>required</a>; <dfn>optional basic constraints</dfn> are the remaining
```
armax00: Here you explicitely specify "optional basic" while before the constraints are just identified as 'required'. I would suggest to keep the description consistent and use 'basic' in both cases, namely
- required basic constraints, and
- optional basic constraints.

getusermedia.html
```diff
@@ -179,13 +179,23 @@ <h2>Terminology</h2>
...
-        application authors are strongly encouraged to use mandatory
-        constraints sparingly.</p>
         <p>For each constrainable property, a constraint exists whose name
         corresponds with the relevant source setting name and capability
         name.</p>
+        <p>A constraint takes one of three forms, depending on the syntax used.
+        Collectively, we refer to all constraints that apply on a track for a
+        type of device, of each form, in three groups as follows:
+        <dfn>required constraints</dfn> are all <a data-lt=
+          "advanced constraints">non-advanced</a> constraints that are
```
armax00: shouldn't this point to something like "basic constraints" instead? If not, wouldn't it be better to linkify only the "advanced" word rather than the whole non-advanced? To me it seems confusing.

getusermedia.html
```diff
@@ -4033,15 +4047,15 @@ <h2>Interface Definition</h2>
       "#media-stream-track-interface-definition">MediaStreamTrack Interface
       Definition</a> for an example of this.</p>
       <p>When the User Agent is no longer able to satisfy the
-      <var>requiredConstraints</var> from the currently valid Constraints, the
+      <a>required constraints</a> from the currently valid Constraints, the
```
armax00: lower case?

getusermedia.html
```diff
@@ -4184,29 +4198,33 @@ <h2>Methods</h2>
...
                   either contains one or more members named 'min', 'max', or
                   'exact', or is itself a bare value and bare values are to be
                   treated as 'exact'), and the <a>settings dictionary</a>'s
                   value for the constraint does not satisfy the constraint, the
                   fitness distance is positive infinity.</p>
                 </li>
                 <li>
-                  <p>If the constraint is not required, and does not apply for
+                  <p>If the constraint is
+                  <a data-lt="required">not required</a>, and does not apply for
```
armax00: same comment about not including 'not' as above.

getusermedia.html
jan-ivar: I had that originally, but then all required constraints are basic, so it seemed redundant, and we're already using "required constraints" elsewhere. This isn't symmetrical.

The only reason I could see to spell out "required basic constraints" everywhere, is to avoid confusion with the "required" definition (when followed by the word "constraint") inside the fitness distance, which is different since it is also used inside the "advanced" algorithm (and "advanced" constraints as a whole are closer to optional). @alvestrand @henbos WDYT?

getusermedia.html
jan-ivar: "Basic" is another word for "not advanced". We seem to be using the term in various places, and it seems to come from https://w3c.github.io/mediacapture-main/getusermedia.html#constrainable-interface

I was trying to limit the terminology definitions to those needed (there are only 3 groups), but I'm open to alternatives. If we go with the "required basic constraints", is that perhaps sufficient?

As to negated links, there's precedence in https://heycam.github.io/webidl/#dfn-present

getusermedia.html
jan-ivar: What do you mean?

armax00: sorry, I thought selecting the word would have highlighted what I was pointing at. the word on the same line "Constraints" should be lower case now, right?

jan-ivar:
> the problem in OverconstrainedErrorEvent is that constraints are mentioned as "current constraints"

Ah, my bad. I'd forgotten to update the manual Preview link I added (now fixed)! It didn't include all my changes. See [here](https://rawgit.com/w3c/mediacapture-main/593f75d7f815839d0ec669719ebe2f97341b4d86/getusermedia.html#event-summary) for the text you're missing.

We really should figure out how to turn on that github feature, so I don't have to hack it (it's turrned on in webrtc-pc).

armax00:
> > the problem in OverconstrainedErrorEvent is that constraints are mentioned as "current constraints"
> 
> Ah, my bad. I'd forgotten to update the manual Preview link I added (now fixed)! It didn't include all my changes. See [here](https://rawgit.com/w3c/mediacapture-main/593f75d7f815839d0ec669719ebe2f97341b4d86/getusermedia.html#event-summary) for the text you're missing.
> 
> We really should figure out how to turn on that github feature, so I don't have to hack it (it's turrned on in webrtc-pc).

Thanks for that, yep this looks like is clarifying the text in the right way with respect to OverconstrainedErrorEvent.

getusermedia.html
alvestrand: Personally, I feel like it's overkill to use links that resolve inside the same sentence.

But the paragraph is troublesome for other reasons:

- It's not forms, it's meanings that are different. I'd suggest "A constraint falls into one of three groups, depending on its place in the constraints structure".

- The whole "Collectively .... as follows" sentence adds absolutely no additional information. Suggest to replace it with "The group are:" and then use a bulleted list with three bullets rather than flowing prose.

getusermedia.html
alvestrand: Agree with jan-ivar here.

getusermedia.html
jan-ivar: I find links helpful when implementing, especially here with the myriad of different definitions and terms. I find it especially helpful to link to terms from outside the spec, like in conversations. There's also precedent in https://heycam.github.io/webidl/#dfn-present so I vote to keep it.

> * It's not forms, it's meanings that are different. I'd suggest "A constraint falls into one of three groups, depending on its place in the constraints structure".

Will incorporate. Thanks!

> * The whole "Collectively .... as follows" sentence adds absolutely no additional information

That wasn't my intent. With *"Collectively, we refer to all constraints that apply on a track for a type of device"*, I was aiming for a definition that excluded
 1. constraints from other specs and kinds
 2. constraints not applied on the track in question.

We all know what we mean, but I was concerned the spec wasn't clear. I'll update it to *"The groups are:"*. I guess we can file new issues if we find references where this scope isn't clear from context.

PR #570, opened 2019-02-15
exclude ConstrainablePattern from reffy.

jan-ivar:
See https://github.com/heycam/webidl/pull/609#issuecomment-463831882. @foolip

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -4049,7 +4049,7 @@ const constraints = {
       manipulate a camera in a way that makes it impossible to provide a
       resolution or frameRate that satisfies the constraints. The User Agent
       MAY take other actions as a result of the overconstrained situation.</p>
-      <div>
+      <div class="exclude">
         <pre class="idl">[NoInterfaceObject]
 interface ConstrainablePattern {
     Capabilities  getCapabilities ();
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -4049,7 +4049,7 @@ <h2>Interface Definition</h2>
       manipulate a camera in a way that makes it impossible to provide a
       resolution or frameRate that satisfies the constraints. The User Agent
       MAY take other actions as a result of the overconstrained situation.</p>
-      <div>
+      <div class="exclude">
```
foolip: You'll need to put the class name on pre for this to work, if I'm reading the reffy code right.

alvestrand:
So what's the point of this?
@foolip can you explain what reffy is, and why excluding this from it is a good idea?

foolip:
IDL in specs end up in WPT via a process described in https://github.com/web-platform-tests/wpt/blob/master/interfaces/README.md which involves https://github.com/tidoust/reffy.

If interfaces which shouldn't be implemented verbatim are included, it can lead to failing idlharness.js tests generated from that IDL.

jan-ivar:
[Preview](https://rawgit.com/w3c/mediacapture-main/81af9767c09335d3826991facc2aff22a6a13025/getusermedia.html#dom-constrainablepattern)

foolip:
Looks good, thanks!

jan-ivar:
[Travis](https://travis-ci.org/w3c/mediacapture-main/builds/498229532?utm_source=github_status&utm_medium=notification) warns of a missing \<p\> but at a line number nowhere near this PR, so it seems unrelated. Merging.

PR #572, opened 2019-02-25
Add video latency constrainable property

kuddai:
We would like to extended existing [latency](https://www.w3.org/TR/mediacapture-streams/#def-constraint-latency) constrainable property API to the video case as well. The justification is essentially the same as for audio. There might be apps which are not sensitive for the higher latency on the video track which can open the same optimization benefits as for audio.

Another problem it might solve is audio video desynchronization. Currently you have option to only increase latency for the audio track and if it has associated video track then they would be desynced.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1698,6 +1698,18 @@ interface MediaStreamTrack : EventTarget {
                 <a>ConstraintSet</a>.
               </td>
             </tr>
+            <tr id="def-constraint-latency">
+              <td>latency</td>
+              <td><code><a>ConstrainDouble</a></code></td>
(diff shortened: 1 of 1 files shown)


henbos:
@guidou and @armax00 does this make sense?

henbos:
This would also require a webrtc-pc change here: https://w3c.github.io/webrtc-pc/#mediatracksupportedconstraints-mediatrackcapabilities-mediatrackconstraints-and-mediatracksettings

In fact, since what you want to change is specific to remote tracks sourced by RTCPeerConnection, you could add the definition for video there and leave mediacapture-main alone.

PR #574, opened 2019-03-21
Mitigate getUserMedia/enumerateDevices fingerprinting issues based on device info permission

youennf:
Introduce a device information exposure check routine that does the checks previously defined to fire the devicechange event.
Use that routine for devicechange event.
Use that routine to postpone getUserMedia and enumerateDevices steps until this routine returns true.
Sanitize constraint error if device info permission is granted.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2919,7 +2919,7 @@ interface OverconstrainedErrorEvent : Event {
       <p>When new media input and/or output devices are made available, or any
       available input and/or output device becomes unavailable, the User Agent
       MUST run the following steps in browsing contexts for which the
-      <a>device information exposure check</a> returns <code>true</code>,
+      <a>device information can be exposed check</a> returns <code>true</code>,
       but in no other contexts:</p>
       <ol>
         <li>
@@ -2932,7 +2932,7 @@ interface OverconstrainedErrorEvent : Event {
         </li>
       </ol>
       <p>If a browsing context later comes to meet the
-      <a>device information exposure check</a> criteria (e.g.
+      <a>device information can be exposed check</a> criteria (e.g.
       <a data-cite="!HTML52/editing.html#gain-focus">gains focus</a>), the User
       Agent MUST execute the steps at that time.</p>
       <p>The User Agent MAY combine firing multiple events into firing one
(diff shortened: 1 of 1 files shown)


jan-ivar:
I should clarify "us" there was Firefox. We should hear from others before merging.

getusermedia.html
```diff
@@ -3128,6 +3128,26 @@ <h2>Access control model</h2>
         <dfn>MediaDeviceInfo</dfn> dictionary will contain the deviceId, kind,
         label, and groupId.</p>
       </section>
+      <section>
+        <h2>Device information exposure check</h2>
+        <p>To perform a <dfn data-lt="device-information-exposure-check" id=
+        "device-information-exposure-check">device information exposure check</dfn>,
```
alvestrand: Suggest to call this check "Device information can be exposed", to give obvious meaning to true and false.

alvestrand:
@guidou can you check if this (waiting with resolution until tab is foreground) is reasonable to implement in Chrome?

guidou:
> @guidou can you check if this (waiting with resolution until tab is foreground) is reasonable to implement in Chrome?

Yes, this is reasonable to implement in Chrome.

youennf:
> Yes, this is reasonable to implement in Chrome.

@guidou, are you saying that Chrome could always wait to reject/resolve until tab is foreground?
Or that the PR as it is is ok.

I am asking this as currently, if device permission is granted (which is I believe the case in Chrome if user granted once camera/mic permission for the origin), the PR does not require to wait.

I believe Safari (and probably Firefox) will probably wait no matter the device permission state.
If it is fine for Chrome to always wait until tab is foreground, we could simplify the PR further.

guidou:
@youennf I meant that Chrome can be modified to wait until the tab is in the foreground. If we implement this in Chrome, it will probably also always wait until the tab is in the foreground. Thus, the PR can be simplified further.

youennf:
Updated PR based on @guidou response.

getusermedia.html
youennf: Done

youennf:
Travis error is: 
getusermedia.html:196:14: Warning: inserting implicit <p>

Given this PR is not changing anything near this line, I think this is a pre-existing issue.

PR #576, opened 2019-04-04
Remove "onoverconstrained"

### LINKED ISSUE #573: Remove overconstrained event (Why do we have overconstrained event?)
In practice it seems that constraints would never be violated for getUserMedia()-supplied tracks. You get what your device is configured for. In theory, the device might not give you what you ask for, but in practice you should. At least the way Chrome is implemented, you would never reconfigure a camera such that existing tracks become overconstrained.

According to Chrome's GUM implementer @guidou, becoming overconstrained therefor doesn't seem applicable other than in the event of non-device supplied sources e.g. network media sources such as RTCPeerConnection's remote tracks.

In recent discussions (https://github.com/w3c/webrtc-pc/issues/2121) it is pointed out that constraints are under-specified, and it wasn't clear what constraints should be applicable to remote tracks.

So question: _Why do we have overconstrained if not for remote tracks?_

@jan-ivar @alvestrand

armax00:
Fixes #573 

Considering the thread in #573, it seems to me that there is a
relatively large consensus as what the faith of onoverconstrained should
be. This commit attempts to remove onoverconstrained from the
mediastream spec.


alvestrand:
Always use informative titles.

getusermedia.html
```diff
@@ -119,15 +119,6 @@ <h2>Terminology</h2>
         <p>A <a href="#settings">setting</a> refers to the immediate, current
         value of the source's constrainable properties. Settings are always
         read-only.</p>
-        <p>A source's settings can change dynamically over time due to
-        environmental conditions, sink configurations, or constraint changes. A
-        source's settings must always conform to the current set of
-        <a>required constraints</a> on all attached tracks. A source that cannot
-        conform to these constraints causes affected tracks to become
```
alvestrand: Don't delete this - it's just reality. Instead change "A source that cannot conform ...." to say "A source may temporarily stray outside the bounds set by constraints, such as when a camera switches to a slower frame rate due to low light conditions. The platform SHOULD try to minimize such excursions as far as possible, but will continue to deliver media even when a temporary condition exists that prevents satisfying the constraint."

armax00: Thanks, I added the text you suggested (please double-check if I did it right).

guidou:
The section added back talks about sources straying outside constraints, but constraints belong to tracks, not sources.
Maybe say that tracks may temporarily stray outside constraints?

armax00:
> The section added back talks about sources straying outside constraints, but constraints belong to tracks, not sources.
> Maybe say that tracks may temporarily stray outside constraints?

You are right with respect to where the constraints reside, but if I understand correctly the concept of source in the spec is loose and might indicate the devices, for example, that are used to record audio. In that case it would be the devices' conditions that change so that the track cannot satisfy the constraints anymore, hence to me it sounds like we do not need the change. Possibly, a way to rephrase it could be this:

"sources conditions may dynamically change such that the related tracks might not satisfy the constraints any longer"

What do you think?

guidou:
Your proposal sgtm

armax00:
> Your proposal sgtm

Done, a slightly rephrased to match the existing text. PTAL.

getusermedia.html
```diff
@@ -119,15 +119,13 @@ <h2>Terminology</h2>
...
-        <a>overconstrained</a> and therefore <a href="#track-muted">muted</a>.
-        A <a>User Agent</a> attempts to ensure that sources adhere to advanced
-        constraints as closely as possible, see <a href=
-        "#constrainable-interface"></a>.</p>
+        <p>A source conditions may dynamically change, such as when a camera
+        switches to a slower frame rate due to low light conditions. In these
```
stefhak: s/slower/lower/ ?

getusermedia.html
```diff
@@ -119,15 +119,13 @@ <h2>Terminology</h2>
...
+        cases the tracks related to the affected source might not satisfy the
+        set constraints any longer. The platform SHOULD try to minimize such
+        excursions as far as possible, but will continue to deliver media even
+        when a temporary condition exists that prevents satisfying the
```
stefhak: I think `temporary`should be removed - there is no way to know whether it's temporary or not usually.

armax00: I agree with you but I would go to the extent of explicitly writing "temporary or permanently".

armax00:
Please, take a look :)

armax00:
Friendly ping.

alvestrand:
Seems ready to merge, but let's check with @dontcallmedom for completeness before merging.

dontcallmedom:
LGTM as well except indeed for the  stray `</p>`

armax00:
> Lgtm modulo stray </p>

What I do not understand is why that </p> should be spurious. There is a <p> open above that would otherwise never get closed. Maybe it should be placed before <ul> ?

armax00:
After two tests we did with hbos, it seems that Travis has something wrong in how it is creating the checkout because is always failing with the same hash (which makes me believe it is not the specific pull request but rather something wrong from Travis' side).

henbos:
Travis is broken, it's not this PR. I got it here too #580:

> Unhandled rejection Error: Command failed: /usr/bin/git checkout 7b144b3b83ce296d5b29c5e4fc2ea2a65e9d3f32
> fatal: reference is not a tree: 7b144b3b83ce296d5b29c5e4fc2ea2a65e9d3f32

Exact same problem

Issue #577, opened 2019-04-23
echo cancellation scope


jgupta:
In reference to [echo cancellation definition](https://w3c.github.io/mediacapture-main/#def-constraint-echoCancellation), it should be removing the sound being played. Now, [some browsers](https://bugs.chromium.org/p/chromium/issues/detail?id=687574) considers only sound being played from Webrtc remote peers for echo cancellation and not from the sound being played from other sources such as Web Audio API.

I think it is desired that all sounds being played should be considered for echo cancellation. If this is not the general case, then should there be an option in specifications for users to choose scope?

henbos:
This spec does not limit echo cancellation to WebRTC-generated audio, it says "it is often desirable to attempt to remove the sound being played from the input signals recorded by the microphones". It sounds like what you are experiencing are browser implementation bugs, not spec bugs.

Alternatively, if we want echo cancellation only to apply to specific use cases such as WebRTC, then perhaps the definition of the constraint should be moved to the WebRTC spec or clarification needed.

alvestrand:
Current text is "When one or more audio streams is being played in the processes of various microphones, it is often desirable to attempt to remove the sound being played from the input signals recorded by the microphones."

It might be clearer if we express it as "When audio is being played out at the same time as recording from a microphone".

jgupta:
We have tested this with various browsers. It appears that Firefox, Safari, Edge considers all audio being played from the browser for echo cancellation. However, Chrome only considers audio being played from the webrtc remote peer connection.

Google Chrome also [does not consider this a spec bug](https://bugs.chromium.org/p/chromium/issues/detail?id=687574#c32) as they say "it's not specified which audio should be cancelled. It specifies that an echo cancellation can be controlled on/off."

I feel that all audio being played should be removed in echo cancellation and there is no requirement to introduce scope but in that case, the spec should not have ambiguity. If there are some cases where it is desired to limit audio sources being played for echo cancellation then there should be a way to control this behaviour.

henbos:
OK the spec should be more explicit about it.

PR #582, opened 2019-04-29
#577 make echo cancellation scope more explicit

jgupta:


dontcallmedom:
Marked as non substantive for IPR from ash-nazg.

PR #588, opened 2019-05-16
Remove the 'volume' constraint.

### LINKED ISSUE #585: Remove or rename 'volume'
The 'volume' constrainable property is defined as being a multiplier applied to the linear audio samples. This means that this property does not control the volume of the audio device capture (which can be adjusted by the user and determined by the OS/HW) but rather the gain. More precisely, since the range allowed is between 0 and 1, this constrainable property controls the loss (gain between 0 and 1).

I propose that this property should be renamed from volume to either gain or loss.

Further, the description of volume in https://www.w3.org/TR/mediacapture-streams/#def-constraint-volume should be clearer in identifying it as gain/loss.

armax00:
Fixes #585

In #585 the WG has expressed interest in removing the volume constraint
from the specifications. This commit addresses the issue by removing the
definition of 'volume' together with the related references.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1140,7 +1140,6 @@ interface MediaStreamTrack : EventTarget {
              boolean frameRate = true;
              boolean facingMode = true;
              boolean resizeMode = true;
-             boolean volume = true;
              boolean sampleRate = true;
              boolean sampleSize = true;
              boolean echoCancellation = true;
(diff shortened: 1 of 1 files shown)


henbos:
If it passes travis we can merge

armax00:
I rebased and it looks like Travis is happy

henbos:
Then let's merge!

PR #589, opened 2019-05-17
Clarify groupId persistency

youennf:
Make it clear groupId is per browsing context.
Refactor to have the groupId constraint refer to MediaDeviceInfo.groupId definition.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1664,17 +1664,17 @@ interface MediaStreamTrack : EventTarget {
               <td><dfn>groupId</dfn></td>
               <td>DOMString</td>
               <td>
-                The browsing session-unique group identifier for the source of
-                the <a>MediaStreamTrack</a>. Two devices have the same group
-                identifier if they belong to the same physical device; for
-                example, the audio input and output devices representing the
-                speaker and microphone of the same headset would have the same
-                groupId. Note that the setting of this property is uniquely
+                The browsing context-unique group identifier for the device
+                generating the content of the <a>MediaStreamTrack</a>.
+                It conforms with the definition of
+                <code><a>MediaDeviceInfo</a></code>.<code><a
+                href="#def-mediadeviceinfo-groupId">groupId</a></code>.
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -1664,17 +1664,17 @@ <h2>Constrainable Properties</h2>
               <td><dfn>groupId</dfn></td>
               <td>DOMString</td>
               <td>
-                The browsing session-unique group identifier for the source of
-                the <a>MediaStreamTrack</a>. Two devices have the same group
-                identifier if they belong to the same physical device; for
-                example, the audio input and output devices representing the
-                speaker and microphone of the same headset would have the same
-                groupId. Note that the setting of this property is uniquely
+                The browsing context-unique group identifier for the device
```
alvestrand: @jan-ivar will suggest what word to use here. "on the relevant global object"?

henbos:
Update to say "document-unique"

getusermedia.html
```diff
@@ -3098,13 +3098,16 @@ <h2>Attributes</h2>
...
-              devices have the same group identifier if they belong to the same
-              physical device; for example a monitor with a built-in camera and
-              microphone.</p>
+              <p>The group identifier of the represented device. Two devices
+              have the same group identifier if they belong to the same
+              physical device. For example, the audio input and output devices
+              representing the speaker and microphone of the same headset
+              have the same groupId.</p>
+              <p>The group identifier MUST be uniquely generated for each
+              browsing context.</p>
```
henbos: for each document

getusermedia.html
henbos: The document-unique

youennf: Done

getusermedia.html
youennf: Done

youennf:
There is a side discussion that we might want to have as well.
To prevent tracking, device ids are currently partitioned by the document origin and top level browsing context origin in WebKit. I'll file a separate issue.

PR #590, opened 2019-05-17
Unify device id description

youennf:
Refactor to have the deviceId constraint description refer to MediaDeviceInfo.deviceId description.
Update the description so that deviceId constraint refers to a device and not a source of the content.
Web audio track and canvas track definitions should probably make it clear that deviceId is not meaningful there.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1626,11 +1626,11 @@ interface MediaStreamTrack : EventTarget {
               <td><dfn>deviceId</dfn></td>
               <td>DOMString</td>
               <td>
-                The origin-unique identifier for the source of the
-                <a>MediaStreamTrack</a>. The same identifier MUST be valid
-                between browsing sessions of this origin, but MUST also be
-                different for other origins. Some sort of GUID is recommended
-                for the identifier. Note that the setting of this property is
+                The identifier of the device generating the content of the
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -1643,11 +1643,11 @@ <h2>Constrainable Properties</h2>
...
               <td>
-                The origin-unique identifier for the source of the
-                <a>MediaStreamTrack</a>. The same identifier MUST be valid
-                between browsing sessions of this origin, but MUST also be
-                different for other origins. Some sort of GUID is recommended
-                for the identifier. Note that the setting of this property is
+                The identifier of the device generating the content of the
+                <a>MediaStreamTrack</a>. It conforms with the definition of
+                <code><a>MediaDeviceInfo</a></code>.<code><a
+                href="#def-mediadeviceinfo-deviceId">deviceId</a></code>.
```
alvestrand: I don't think we should take out the language about "valid between browsing sessions", and the recommendation for GUID.
Either leave them in, or block this one on the PR that restores it.

youennf: I moved these lines in this PR to make progress.
We will see how to update them in other related PRs.

aboba:
@youennf Can you rebase?

PR #591, opened 2019-05-17
Recommend UUID for deviceId and groupId

youennf:


getusermedia.html
```diff
@@ -2944,6 +2944,14 @@ <h2>Methods</h2>
                           "MediaDeviceInfo">groupId</a></code> member be a
                           newly generated unique identifier.</p>
                         </li>
+                        <p>A good practice for generating <code><a
+                          data-link-for="MediaDeviceInfo">deviceId</a></code>
+                          and <code><a data-link-for="MediaDeviceInfo">groupId</a></code>
+                          values is to use a UUID [[rfc4122]], which is 36
+                          characters long in its canonical form. To avoid
+                          fingerprinting, implementations SHOULD use the forms in
+                          section 4.4 or 4.5 of RFC 4122.
```
alvestrand: This is the CL that should take out the corresponding language in the "id" attribute of 4.2 "MediaStream".

youennf: I am not sure to understand what you suggest.
If we do not want this text in two different places, we can move it to a single place and link to it from MediaStream.id, MediaDeviceInfo.deviceId and MediaDeviceInfo.groupId

getusermedia.html
```diff
@@ -2944,6 +2944,14 @@ <h2>Methods</h2>
...
+                        </p>
```
alvestrand: Missing ```</li> ```

alvestrand:
Merge once Travis goes green.

youennf:
Rebased and moved the p section inside groupId li section.

PR #595, opened 2019-05-21
Interpret the empty string as if the constraint is not specified

youennf:


alvestrand:
For backwards compatibility reasons, it seems that we only want to accept "empty string" as a device ID, not for every constraint - other constraints have "MUST be a member of the following values..."

youennf:
I am fine with special casing empty string to deviceId only.

> other constraints have "MUST be a member of the following values..."

Can you point me to the part of the spec requiring this?

getusermedia.html
```diff
@@ -1642,6 +1642,9 @@ <h2>Constrainable Properties</h2>
                 "MediaStreamTrack">applyConstraints()</a>, since any attempt to
                 set a different value will result in an unsatisfiable
                 <a>ConstraintSet</a>.
+                If a string of length 0 is used as a deviceID value constraint
```
henbos: deviceId

youennf:
Updated typo from deviceID to deviceId.

Issue #599, opened 2019-05-28
Clarification needed on HTMLMediaElement attributes that carry over


Pehrsons:
Imagine the following code:

```
let v = document.createElement("video");

v.playbackRate = 0.5;
console.log(v.playbackRate);

v.srcObject = new MediaStream();
console.log(v.playbackRate);

v.srcObject = null;
console.log(v.playbackRate);
```

Per the spec the first two logged lines will be `0.5` and `1`, respectively. But what should be returned the third time? Does it keep `1`, or is it reset to `0.5` which it had before the `MediaStream` was assigned?

The same applies to the `defaultPlaybackRate` and `preload` attributes.

The spec also says to not fire `ratechange` on setting `playbackRate` or `defaultPlaybackRate`. That makes sense since the value doesn't change. But shouldn't `ratechange` reasonably be fired if the attribute was changed to `1` when the `MediaStream` was assigned, and (possibly) when it changes back after being unassigned?

Pehrsons:
I eventually found this in the HTMLMediaElement spec:

> The media element load algorithm consists of the following steps.
> 1.   Abort any already-running instance of the resource selection algorithm for this element.
> 2.   Let pending tasks be a list of all tasks from the media element's media element event task source in one of the task queues.
> 3.   For each task in pending tasks that would resolve pending play promises or reject pending play promises, immediately resolve or reject those promises in the order the corresponding tasks were queued.
> 4.   Remove each task in pending tasks from its task queue
> 
> Basically, pending events and callbacks are discarded and promises in-flight to be resolved/rejected are resolved/rejected immediately when the media element starts loading a new resource.

That takes care of my "ratechange" concern. But the question remains of what values `playbackRate`, `defaultPlaybackRate` and `preload` should have when changing to a non-MediaStream source after having had a MediaStream source..

jan-ivar:
> But what should be returned the third time?

`0.5`, because [html](https://html.spec.whatwg.org/multipage/media.html#dom-media-playbackrate) says *"on getting it must return the last value it was set to, ..."*, which was `0.5`.

[Mediacapture](https://w3c.github.io/mediacapture-main/getusermedia.html#mediastreams-in-media-elements) says: *"On setting: ignored. On getting: return 1.0"*. It doesn't say *"set [state] to 1"*.

This might have been clearer had the HTML spec used internal slots, but I think the principle still holds, that mediacapture does not say to alter any (internal and implicit) state.

[...]

PR #600, opened 2019-05-28
Clarify language around element attributes after stream removal.

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/599.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2265,7 +2265,7 @@ interface MediaStreamTrackEvent : Event {
           <td><code>boolean</code></td>
           <td><code>false</code></td>
           <td>A MediaStream is not seekable. Therefore, this attribute MUST
-          always return the value <code>false</code>.</td>
+          always have the value <code>false</code>.</td>
         </tr>
         <tr>
           <td>
@@ -2276,7 +2276,7 @@ interface MediaStreamTrackEvent : Event {
           <td><code>double</code></td>
           <td>On getting: <code>1.0</code>. On setting: ignored.</td>
           <td>A MediaStream is not seekable. Therefore, this attribute MUST
-          always return the value <code>1.0</code> and any attempt to alter it
+          always have the value <code>1.0</code> and any attempt to alter it
           MUST be ignored. Note that this also means that the
           <code><a data-cite=
           "!HTML52/semantics-embedded-content.html#eventdef-media-ratechange">ratechange</a></code>
@@ -2291,7 +2291,7 @@ interface MediaStreamTrackEvent : Event {
           <td><code>double</code></td>
           <td>On getting: <code>1.0</code>. On setting: ignored.</td>
           <td>A MediaStream is not seekable. Therefore, this attribute MUST
-          always return the value <code>1.0</code> and any attempt to alter it
+          always have the value <code>1.0</code> and any attempt to alter it
           MUST be ignored. Note that this also means that the
           <code><a data-cite=
           "!HTML52/semantics-embedded-content.html#eventdef-media-ratechange">ratechange</a></code>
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -2337,13 +2338,29 @@ <h2>MediaStreams in Media Elements</h2>
...
           <td>Setting the <code>loop</code> attribute has no effect since a
           <code><a>MediaStream</a></code> has no defined end and therefore
           cannot be looped.</td>
         </tr>
       </tbody>
     </table>
+    <p>When an <code>HTMLMediaElement</code>'s <code><a data-cite=
+      "!HTML52/semantics-embedded-content.html#dom-htmlmediaelement-srcobject">srcObject</a></code>
+    is assigned a <code>MediaStream</code>, and its previous value was
+    <code>null</code>, the user agent MUST change the element's attributes
```
youennf: s/change/set/
https://html.spec.whatwg.org/multipage/media.html#media-element-load-algorithm step 7 is talking about setting the playbackRate attribute. This 'set' action will trigger a rate change event if needed.
Ideally the media.html spec would have a set playbackRate algorithm that we could refer and detail how rate change event is fired.
I would reuse the media.html wording though and talk about setting the attribute(s). This might allow to not mention ratechange event, or mention it in a non normative way only.

That also raises the question of when the 'set' action actually happens.
If we look at https://html.spec.whatwg.org/multipage/media.html#dom-media-srcobject, the 'set' action should either happen right before or right after the media element load algorithm.

jan-ivar: I decided to revert. See https://github.com/w3c/mediacapture-main/issues/599#issuecomment-498406770

jan-ivar:
@Pehrsons can you write a wpt test?

Pehrsons:
I have a [WIP one](https://hg.mozilla.org/try/rev/883ff03849df96f67ab7418102132658f9502162) going through our test suite now. Expect it on [Firefox bug 1554699](https://bugzilla.mozilla.org/show_bug.cgi?id=1554699) soon.

alvestrand:
Tagging @guidou in case he wants to comment on Chrome's implementation.

getusermedia.html
```diff
@@ -2246,8 +2246,9 @@ <h2>MediaStreams in Media Elements</h2>
             <code>currentTime</code></a>
           </td>
           <td><code>double</code></td>
-          <td>Any non-negative integer. The initial value is 0 and the values
-          increments linearly in real time whenever the stream is playing.</td>
+          <td>Any non-negative integer. The initial value is <code>0</code> and
+          the values increments linearly in real time whenever the stream is
```
youennf: s/values/value/

getusermedia.html
```diff
@@ -2337,13 +2338,36 @@ <h2>MediaStreams in Media Elements</h2>
...
+      >srcObject</a></code> is assigned <code>null</code> or a non-stream object,
+      just ahead of the <a data-link-type="dfn"
+      href="https://html.spec.whatwg.org/multipage/media.html#media-element-load-algorithm"
+      >media element load algorithm</a>. As a result, the <code><a data-cite=
+      "!HTML52/semantics-embedded-content.html#eventdef-media-ratechange">ratechange</a></code>
+      event may fire (from step 7) if <a class="externalDFN" data-cite=
+      "!HTML52/semantics-embedded-content.html#dom-htmlmediaelement-playbackrate">
+      <code>playbackRate</code></a> and <a class="externalDFN" data-cite=
+      "!HTML52/semantics-embedded-content.html#dom-htmlmediaelement-defaultplaybackrate">
+      <code>defaultPlaybackRate</code></a> were different from before a
```
youennf: "were different from before" is unclear to me.
My initial interpretation was that playbackRate and defaultPlaybackRate values were different before/after the MediaStream was assigned.
I think the intent is to say that the "playbackRate and defaultPlayBackRate values were different from one another"

guidou:
The clarifications in this change LGTM.

Issue #549, opened 2018-10-25
Clarify that enumerateDevices must not expose devices that a given context cannot use through getUserMedia


youennf:
getUserMedia access to camera/mics/speakers can be controlled by FeaturePolicy.
In case, feature policy disables camera access, enumerateDevices should not enumerate camera devices. Ditto for other types of devices.

From my reading of the spec, this is the required behavior.
I haven't found a WPT test though and I am unsure whether browsers are actually implementing that.

A WPT test would be good if this is actually not covered.
Maybe a note in the spec if we see browsers not implementing this (although a test might be more effective than a note).

jan-ivar:
I think this is a good idea. It'd mean *enumerateDevices* would resolve empty from cross-origin iframes unless either `allow="camera"`, `allow="microphone"`, `allow="speaker"`, or any combination of those were specified, and even then you'd only get the relevant info.

> From my reading of the spec, this is the required behavior.

It seems to hinge on the definition of "available". I think this deserves more specific mention. I'll work on a PR.

PR #603, opened 2019-06-06
Filter enumerateDevices by feature policy.

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/549.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2875,11 +2875,14 @@ interface MediaDevices : EventTarget {
                   <p>Run the following steps in parallel:</p>
                   <ol>
                     <li>
+                      <p>Let <var>document</var> be the <a>current settings object</a>'s
+                      <a>responsible document</a>.</p>
+                    </li>
+                    <li>
                       <p>The User Agent MUST wait to proceed to the next step until the
                       <a>device information can be exposed check</a> returns <code>true</code>.</p>
                       <p>The User Agent MAY wait to proceed to the next step
-                      until the <a>current settings object</a>'s <a>responsible
-                      document</a> is <a data-cite=
+                      until <var>document</var> is <a data-cite=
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -2905,6 +2908,22 @@ <h2>Methods</h2>
...
+                          with the next device (if any).</p>
+                        </li>
+                        <li>
+                          <p>If <var>device</var> is a microphone, and
+                          <var>document</var> is not <a data-link-type="dfn"
+                            href="https://html.spec.whatwg.org/multipage/iframe-embed-object.html#allowed-to-use"
+                            >allowed to use</a> the feature identified by
+                          <code>"microphone"</code>, abort these steps and
+                          continue with the next device (if any).</p>
+                        </li>
```
youennf: I like that!

It would also make sense to filter output speakers according https://w3c.github.io/permissions/#dom-permissionname-speaker feature policy.

There might be some potential side effects of this filtering with oldList, depending on how oldList actually works. Maybe we should get rid of oldList.

jan-ivar: @youennf `oldList` by definition is stuff that's already been exposed, so I don't think there's a privacy issue there. Can feature-policy change over the lifetime of an iframe anyway?

Agree with speaker. Should I update the PR or merge this and create a new for speaker?

Malvoz: @jan-ivar Going through the [FP features list](https://github.com/w3c/webappsec-feature-policy/blob/master/features.md#proposed-features) to potentially update the spec link to the `speaker` feature, I thought it'd be listed here: https://w3c.github.io/mediacapture-main/#feature-policy-integration. I suppose this is TBD per this comment?

Also https://github.com/w3c/mediacapture-main/issues/434#issuecomment-371389896

Thanks :)

Issue #604, opened 2019-06-06
oldList definition in enumerateDevices is unclear


youennf:
oldList in enumerateDevices seems to be per browsing session.
I guess it should be scoped by document, as groupId is per-document and we also want deviceIds to be partitioned per origin, which oldList might mess up with.

It is also not clear when oldList is set/built/updated. Is it set the first time enumerateDevices is called? Or is oldList the result of the last time enumerateDevices is called? Or is it a consolidation of all devices exposed during the various enumerateDevices calls?

I am also not very clear about the purpose of oldList as well. It seems that this is either an optimization (unnecessary then), a way to make sure deviceId/groupId are consistent call after call (but this is insufficient for deviceIds which are living longer than the browsing session and somehow redundant with MediaDeviceInfo.deviceId/groupId definitions), or a way for device === when enumerateDevices is called more than once (testing suggests this is not implemented this way in browsers currently).

I would tend to try removing oldList.

youennf:
> testing suggests this is not implemented this way in browsers currently

https://jsfiddle.net/63Lorgkq/

youennf:
Above jsfiddle.net tends to show that browsers do not implement step 3 of https://w3c.github.io/mediacapture-main/#dom-mediadevices-enumeratedevices.

Also, step 8.2 is not very clear in whether filteredList is a deep clone or not of resultList.
If not, the same object would be returned but the label getter would return the label conditionally on the 'device-info' permission. This is probably worth a separate issue.

youennf:
Closing since remaining items were fixed in https://github.com/w3c/mediacapture-main/issues/620

PR #617, opened 2019-08-30
Remove oldList and make it clear that groupId scope is the document

youennf:
PR for https://github.com/w3c/mediacapture-main/issues/604.
This hopefully makes it clear that a groupId is stable for the lifetime of a document.
Previously, if a device groupId was not in oldList/resultList, it would need to be regenerated.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2914,13 +2914,6 @@ interface MediaDevices : EventTarget {
                       <p>Let <var>resultList</var> be an empty list.</p>
                     </li>
                     <li>
-                      <p>If this method has been called previously within this
-                      browsing session, let <var>oldList</var> be the list of
-                      <code><a>MediaDeviceInfo</a></code> objects that was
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -2913,13 +2913,6 @@ <h2>Methods</h2>
                     <li>
                       <p>Let <var>resultList</var> be an empty list.</p>
```
henbos: Can we make "If [[storedDeviceList]] is not null, then let resultList be a copy of [[storedDeviceList]], and jump to the step labeled Complete Enumeration." say to do a deep copy instead of a copy?

youennf: We could but this is not related to this particular issue.
This would partially fix https://github.com/w3c/mediacapture-main/issues/620.

alvestrand:
This means that enumerateDevices always returns new objects, so object-equals is never true between two calls to enumerateDevices.

youennf:
WPT tests: web-platform-tests/wpt#19488

PR #622, opened 2019-09-10
Upgrade to Respec 24.32.1, fix constructor and optional dictionary syntax

### LINKED ISSUE #621: getUserMedia() and getSupportedConstraints() don't have useful `lt` values
So, this *looks* like a problem in ReSpec itself, not the spec document, but for whatever reason, the `dfn` for getUserMedia() has busted/incorrect `lt` values.

In particular, in the rendered source, it has `data-lt="mediadevices.getusermedia()|mediadevices.getusermedia|getusermedia()|getusermedia"`.

Every single one of these is wrong, because all of the casing has been lost. Thus, when Bikeshed reads this spec and extracts all the definitions, someone correctly attempting to autolink to the definition with `{{getUserMedia()}}` will get a linking failure.

Since the HTML source doesn't appear to have any `lt` at all, I'm pretty sure this is something broken in ReSpec attempting to generate Bikeshed-compatible `lt` values. I dunno how this work - can y'all override this or something? Maybe just add `data-lt="getUserMedia()|getUserMedia(constraints)"` to the element?

You also need to add a type to the definition, which I guess is done by manually adding `data-dfn-type=method`? Maybe that'll fix it the first problem, in fact.

getSupportedConstraints() has the exact same problem, so you're gonna want to apply the fix there as well. And maybe other methods in the spec, I haven't checked.

dontcallmedom:
this also should fix #621


dontcallmedom:
The remaining Travis error is due to a likely bug in ReSpec https://github.com/w3c/respec/issues/2513 - I would be tempted to ignore it, if only temporarily

PR #626, opened 2019-09-27
Update to ReSpec 24.33

saschanaz:
Couldn't see a reason to keep the comments inside `<script>`, does anyone have one?


saschanaz:
Marked as non substantive for IPR from ash-nazg.

PR #633, opened 2019-10-24
Update applyConstraints to be a no-op in case track is ended

youennf:


getusermedia.html
```diff
@@ -1055,32 +1055,53 @@ <h2>Methods</h2>
...
+                    <code><a>MediaStreamTrack</a></code> object.</p>
+                  </li>
+                  <li>
+                    <p>If <var>track</var>'s <code><a href=
+                    "#dom-mediastreamtrack-readystate">readyState</a></code>
+                    attribute is ended, run the following sub steps:</p>
+                    <ol>
+                      <li><p>Let <var>p</var> be a new promise.</p></li>
+                      <li><p><a>resolve</a> <var>p</var> with <code>undefined</code>.</p></li>
+                      <li><p>Return <var>p</var>.</p></li>
```
henbos: Wouldn't it make more sense to reject the promise if the track has ended? (What does it mean to constrain an ended track?)

youennf: I do not have a strong opinion there. I see some value in not rejecting:
- All browsers tend to not reject in case of calling applyConstraints({}) on an ended track.
- Backward-compatibility risk might be lower.
- applyConstraints is basically a no-op in case of an ended track.

PR #637, opened 2019-10-29
Use HTMLMediaElement's ended playback definition.

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/518.

[Preview](https://rawgit.com/w3c/mediacapture-main/98588c04c0c76aeabbd0cd9a690ae594a3ba7c57/getusermedia.html#mediastreams-in-media-elements)

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2188,37 +2188,74 @@ interface MediaStreamTrackEvent : Event {
         is ordered</p>
       </li>
       <li>
-        <p>If the element is an <code>HTMLVideoElement</code>, then when the
-        <code><a>MediaStream</a></code> state moves from the active to the
-        <a href="#stream-inactive">inactive</a> state, the User Agent MUST
-        raise an <a data-cite=
-        "!HTML/#event-media-ended">ended</a>
-        event on the <code>HTMLVideoElement</code> and set its <a data-cite=
-        "!HTML/#dom-media-ended">ended</a>
-        attribute to <code>true</code>. Note that once <a data-cite=
-        "!HTML/#dom-media-ended">ended</a>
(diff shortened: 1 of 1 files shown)


jan-ivar:
@Pehrsons PTAL

Pehrsons:
The text for having ended playback when the stream is active/audible got tricky, but works. I don't see an immediately simpler way of phrasing it.

LGTM, thanks!

getusermedia.html
```diff
@@ -2188,37 +2188,76 @@ <h2>MediaStreams in Media Elements</h2>
...
+            or greater, and</p>
+            <ol>
+              <li>
+                <p>The <code><a>MediaStream</a></code> state is
+                <a href="#stream-inaudible">inaudible</a> after having been
+                <a href="#stream-audible">audible</a>, or</p>
+              </li>
+              <li>
+                <p>The <code><a>MediaStream</a></code> state is
+                <a href="#stream-audible">audible</a> after having been
```
henbos: "inaudible after having been audible", not "audible after having been inaudible after having been audible"

henbos: Same with "active" above for video, is this on purpose?

Pehrsons: If you reach HAVE_METADATA or greater, then playback ends, then play() is called;
- Either play() was called when the stream was audible: ending is handled by the former inaudible-after-audible case, or
- play() was called when the stream was inaudible: ending is handled by this corner case
  - inaudible-after-audible doesn't work here since for us to reach HAVE_METADATA it must have been audible, so playback would end immediately after this case of play().

Or so I interpret it. Does it make more sense now?

Issue #612, opened 2019-08-21
Move enumerateDevices behind permission


snyderp:
Currently websites can enumerate hardware capabilities with enumerateDevices().  Querying device capabilities is the kind of pseudo-identifier that allow users to be passively identified / fingerprinted.  To prevent this, websites should, to the greatest degree possible, need user permission to query device capabilities.  Websites should therefor need permission to access enumerateDevices().

Related: the Moz efforts discussed at the bottom of this comment : https://github.com/w3c/mediacapture-main/issues/607#issuecomment-514460424

vr000m:
@pallab-gain @karthikbr82 Please take a look

youennf:
enumerateDevices has a "device-info" permission (https://w3c.github.io/permissions/#dom-permissionname-device-info). This permission is granted when getUserMedia is granted. This permission controls exposure of device labels currently, not exposure of devices.

Safari goes a bit further by obfuscating most information about devices if 'device-info' is not granted: it obfuscates most devices. This is probably somehow similar to what Mozilla is experimenting and @guido is mentioning (empty device IDs).

Another approach would be to prompt in case enumerateDevices is called and "device-info" permission is not granted. This seems harder to deploy given how enumerateDevices is used currently. The prompt user message might be also more difficult to understand than the getUserMedia message.

The additional use case to consider is for a webpage to monitor whether there is any camera/microphone at all, and update its UI accordingly.
A website can get this information through getUserMedia which seems more appropriate than enumerateDevices from a privacy perspective. ondevicechange is currently also available for that purpose.

youennf:
One possibility discussed on the mailing list is:
- Remove "device-info"permission
- Allow unfiltered enumerateDevices results (and devicechange events) ONLY if getUserMedia was previously called successfully for the document.
This would essentially put enumerateDevices behind the getUserMedia permission prompt.

youennf:
Adding some more filtering in https://github.com/w3c/mediacapture-main/pull/632

youennf:
Uploaded https://github.com/w3c/mediacapture-main/pull/641 to remove "device-info" permission.
Another possibility would be to keep "device-info" permission but also add this boolean so that a given browsing context can only get device info after at least one getUserMedia successful call.

youennf:
Now that https://github.com/w3c/mediacapture-main/pull/641 is merged, enumerateDevices is effectively behind a camera/microphone permission.
[...]

PR #641, opened 2019-11-13
Replace device-info permission by a browsing context boolean flag

youennf:


alvestrand:
I disagree with this approach.
I think it has no advantages over the permission-based approach, and requires special-casing situations where we want to have device-enumeration permission separate from device-access permission, rather than just saying "and you can grant this permission in that case too".

Please file a bug describing the problem you want to solve.

youennf:
The bug is https://github.com/w3c/mediacapture-main/issues/612.
It is not great that because I give access to a camera once to take a picture of me, the website will be able to get all my device information for my future visits.

Also, we agreed that the usual good approach is to do getUserMedia-then-enumerateDevices.
This PR is consistent with that approach.

Additionally, I am not sure the device info permission concept is good.
Other permissions are things like "can I use a camera", which is easy to describe to a user.
This specific permission is hard to explain to a user and AFAIK, is not exposed by any UI, contrary to camera/mic access.

jan-ivar:
This PR removes an arguably redundant permission, which I approve of.

But it still says user agents may revoke this ability (permission) _whenever they want_. This could be:
1. Never
2. During actual camera or mic capture, sabotaging post-gum in-content selection 

I think we need to define this in more detail to protect web interop. 

Original intent: when users revoke both camera and microphone permission, they revoke device-info as well (not before). I.e.
```
device_info = camera_permission || microphone_permission
```
This is true in all browsers I tested, so it must have been the original intent. Other parts like `groupId` support this notion. I vote we tighten the spec to conform with implementations here.

This was also meant to be extensible, which is where I share @alvestrand's concern. E.g.
```
device_info = camera_permission || microphone_permission || speakers
```
If we end up not needing this because of https://github.com/w3c/mediacapture-output/pull/86, great! But it would still be nice if this spec didn't forbid extension, I think—Forbidding user agents from doing whatever they want is not the same as forbidding other specs from extending things.

Is there a way to write it like this?

jan-ivar:
I understand permissions may be per-device, so if this needs to be per-device that would WFM.

getusermedia.html
```diff
@@ -3107,18 +3109,18 @@ <h2>Device information can be exposed check</h2>
...
-          <li>If any of the input devices are attached to an
-            <a href="#stream-active">active</a> <code><a>MediaStream</a></code>
-            in the browsing context, return <code>true</code>.</li>
           <li>If the <a>current settings object</a>'s <a>responsible
           document</a> is <a data-cite="!HTML/#fully-active">fully
           active</a> and <a data-cite="!HTML/#gains-focus">has
           focus</a>, return <code>true</code>.
           </li>
           <li>Return <code>false</code>.</li>
+          <div class="note">
```
dontcallmedom: this would need to be moved in the `<li>`

alvestrand:
Needs a PR to remove device-info from feature-policy. @youennf has promised to do this.

Issue #628, opened 2019-10-07
applying constraints on an ended track


youennf:
I did https://jsfiddle.net/2kpdbnyz/ to test applyConstraints and getSettings on an ended track.
Testing Firefox, Chrome and Safari, all browsers behave differently:

- Firefox is resolving applyConstraint promise for ended track, getSettings returns an empty dictionary for ended tracks
- Safari tis resolving applyConstraint promise for ended track, getSettings returns some parameters that applyConstraints can change
- Chrome is rejecting applyConstraint promise for ended tracks (at least if some constraints are set), getSettings returns a reduced set of the settings after the track is ended (deviceId, groupId)

Looking at the spec, all implementations could potentially qualify as being compliant.
It would be better though to get some alignment there.
We could special case applyConstraints/getSettings for ended tracks, something like:
- Resolve applyConstraints promise early if track is ended
- Return deviceId/groupId settings if track is ended

alvestrand:
After a track is ended, it shouldn't matter what the constraints are - it won't affect the output (there is none).

deviceId / groupId would seem to make sense to return, because it tells you what the track is attached to - this is info that could let you try to reopen the device, if so desired. I could argue that the other settings should be frozen in time as they were when the track ended, but if we currently don't return them, I don't see a strong reason to bother.

Thinking abstractly, I'd say that attempting to set constraints on a dead track is a silly thing and should be rejected with WrongStateError (or whatever similar error exists). But we're normally claiming that an applyConstraints() that doesn't cause any setting to change should always succeed. I could go either way on that one.

guest271314:
> Testing Firefox, Chrome and Safari, all browsers behave differently:

FWIW the implementations of `MediaStreamTrack` at Chrome, Chromium and Firefox, Nightly are _vastly_ different in several observable ways. 

Have not yet compiled a discrete list of all of the differences that have observed through own experimentation and testing, though can compile such a list if that would be helpful for interoperability and if the gist of this issue is 

> It would be better though to get some alignment there.

which agree with.

Some differences, particularly with regard to capturing a `MediaStream` from an element

- [Issue 957340: MediaStreamTrack ended event not dispatched when the source is changed by setting the src or srcObject attributes of the media element or when the source of the track is disconnected or exhausted](https://bugs.chromium.org/p/chromium/issues/detail?id=957340); https://github.com/w3c/mediacapture-fromelement/issues/78
[...]

PR #657, opened 2020-01-22
In case track is ended, make sure deviceId/groupId are exposed if they were exposed when the track was not ended

youennf:


youennf:
Fixes https://github.com/w3c/mediacapture-main/issues/628 for getSettings

getusermedia.html
```diff
@@ -1044,8 +1044,32 @@ <h2>Methods</h2>
...
+                  <li>
+                    <p>If <var>track</var>'s <code><a href=
+                    "#dom-mediastreamtrack-readystate">readyState</a></code>
+                    attribute is ended, run the following sub steps:</p>
+                    <ol>
+                      <li><p>Let <var>settings</var> be a new <a><code>MediaTrackSettings</code></a> dictionary.</p></li>
+                      <li><p>Set <var>settings</var>.<code><a href="#def-mediadeviceinfo-deviceId">deviceId</a></code>
+                      to the identifier of the device that was generating content, if any.</p></li>
+                      <li><p>Set <var>settings</var>.<code><a href="#def-mediadeviceinfo-groupId">groupId</a></code>
+                      to the group identifier of the device that was generating content, if any.</p></li>
```
jan-ivar: I think we need to make this extensible. I suggest breaking out a **list of inherent constrainable track properties** definition, add `deviceId`, `groupId` and `facingMode` to that list, and refer to it here.

This would let us add to the list in the future. See https://github.com/w3c/mediacapture-main/issues/628#issuecomment-577766841

youennf: Added the list of properties

Issue #642, opened 2019-11-18
Only Firefox turns off device on disabled track. Stronger language needed?


jan-ivar:
[Live](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediastreamtrackstate-live) camera/mic tracks can be turned on/off with the [enabled](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediastreamtrack-enabled) attribute.

The obvious (and only?) use-case is the [mute button](https://jsfiddle.net/jib1/ru16j2y8/):
```js
muted.onclick = () => { video.srcObject.getVideoTracks()[0].enabled = !muted.checked; }
```

Unfortunately, only Firefox [turns off the device](https://blog.mozilla.org/webrtc/better-privacy-on-camera-mute-in-firefox-60/) (and hardware light) in this case.

This has led to web compat issues:

Sites like Hangouts use a hack to get this desired property on Chrome: relinquish camera track on mute, and reacquire it w/`getUserMedia` on unmute (the microphone track is left alone, presumably because few mics have a light?)

The hack backfires in Firefox, where Hangouts users get a needless extra prompt on unmute. Frustratingly, it would have worked beautifully in Firefox without the hack.

We worked hard to get this right in https://github.com/w3c/mediacapture-main/issues/389, but do we need stronger language to fix this web compat issue?

Right now [we have](https://www.w3.org/TR/mediacapture-streams/#life-cycle-and-media-flow): _"when a track becomes either muted or disabled, and this brings all tracks connected to the device to be either muted, disabled, or stopped, then the UA MAY, using the device's deviceId, deviceId, set [[devicesLiveMap]][deviceId] to false"_

Should this be a MUST? And do we need to say something about actually relinquishing the device?

If we can't get consensus on compatible behavior, should we scrap `enabled`?

henbos:
Is the desired behavior from Hangouts' perspective "mute the track but don't turn the device off"? It would seem like this behavior is already supported with replaceTrack(null), and that using "enabled = false" for this effect is simply using the API incorrectly.

Remind me, what is there a difference between sending a disabled track and sending a null track?

If on the other hand we do turn off the device and the light, then re-prompting seems like the right thing to do. For privacy reasons we don't want a site to turn on the device without prompting.

henbos:
Hang on, if the desired effect is "turn off the device", why not use track.stop()? The way Chrome handles track.enabled kind of makes sense to me actually.

[...]

PR #662, opened 2020-02-25
Relinquish device when all tracks are muted or disabled.

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/642.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -725,20 +725,23 @@ interface MediaStream : EventTarget {
         provided the UA sets it back to <code>true</code> as soon as any
         unstopped track connected to this device becomes un-muted or enabled
         again.</p>
-        <p>When a live track sourced by a device exposed by <a data-link-for=
-        "MediaDevices">getUserMedia()</a> becomes either muted or disabled,
+        <p>When a <code>live</code>, <a href="#track-muted">unmuted</a>, and
+        <a href="#track-enabled">enabled</a> track sourced by a device exposed
+        by <a data-link-for="MediaDevices">getUserMedia()</a> becomes either
+        <a href="#track-muted">muted</a> or <a href="#track-enabled">disabled</a>,
         and this brings <em>all</em> tracks connected to the device to be either
         muted, disabled, or stopped, then the UA MUST relinquish the device
         within 3 seconds while allowing time for a reasonably-observant user to
         become aware of the transition. The UA MUST attempt to reacquire the
-        device as soon as any live track sourced by the device becomes both
-        unmuted and enabled again, provided the <var>document</var>
+        device as soon as any <code>live</code> track sourced by the device
+        becomes both <a href="#track-muted">unmuted</a> and
+        <a href="#track-enabled">enabled</a> again, provided the <var>document</var>
         <a data-cite="!HTML/#gains-focus">has focus</a> at that time. If the
         <var>document</var> does not
         <a data-cite="!HTML/#gains-focus">have focus</a> at that time,
         the UA MUST instead queue a task to <a href="#track-muted">mute</a> the
-        track, and <a href="#track-muted">unmute</a> it later once the document
-        <a data-cite="!HTML/#gains-focus">gains focus</a>.
+        track, and MUST NOT <a href="#track-muted">unmute</a> the track until
+        the document <a data-cite="!HTML/#gains-focus">regains focus</a>.
         If reacquiring the device fails, the UA MUST
         <a href="#ends-nostop">end the track</a> (The UA MAY end it earlier
         should it detect a device problem, like the device being physically
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -740,7 +740,7 @@ interface MediaStream : EventTarget {
         <var>document</var> does not
         <a data-cite="!HTML/#gains-focus">have focus</a> at that time,
         the UA SHOULD instead queue a task to <a href="#track-muted">mute</a> the
-        track, and not <a href="#track-muted">unmute</a> the track until
+        track, and not queue a task to <a href="#track-muted">unmute</a> it until
         the document <a data-cite="!HTML/#gains-focus">regains focus</a>.
         If reacquiring the device fails, the UA MUST
         <a href="#ends-nostop">end the track</a> (The UA MAY end it earlier
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -729,16 +729,17 @@ interface MediaStream : EventTarget {
         <a href="#track-enabled">enabled</a> track sourced by a device exposed
         by <a data-link-for="MediaDevices">getUserMedia()</a> becomes either
         <a href="#track-muted">muted</a> or <a href="#track-enabled">disabled</a>,
-        and this brings <em>all</em> tracks connected to the device to be either
-        muted, disabled, or stopped, then the UA MUST relinquish the device
+        and this brings <em>all</em> tracks connected to the device (regardless
+        of browsing context) to be either
+        muted, disabled, or stopped, then the UA SHOULD relinquish the device
         within 3 seconds while allowing time for a reasonably-observant user to
-        become aware of the transition. The UA MUST attempt to reacquire the
+        become aware of the transition. The UA SHOULD attempt to reacquire the
         device as soon as any <code>live</code> track sourced by the device
         becomes both <a href="#track-muted">unmuted</a> and
-        <a href="#track-enabled">enabled</a> again, provided the <var>document</var>
+        <a href="#track-enabled">enabled</a> again, provided that track's
+        <a>current settings object</a>'s <a>responsible document</a>
         <a data-cite="!HTML/#gains-focus">has focus</a> at that time. If the
-        <var>document</var> does not
-        <a data-cite="!HTML/#gains-focus">have focus</a> at that time,
+        document does not <a data-cite="!HTML/#gains-focus">have focus</a> at that time,
         the UA SHOULD instead queue a task to <a href="#track-muted">mute</a> the
         track, and not queue a task to <a href="#track-muted">unmute</a> it until
         the document <a data-cite="!HTML/#gains-focus">regains focus</a>.
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -739,8 +739,8 @@ interface MediaStream : EventTarget {
         <a data-cite="!HTML/#gains-focus">has focus</a> at that time. If the
         <var>document</var> does not
         <a data-cite="!HTML/#gains-focus">have focus</a> at that time,
-        the UA MUST instead queue a task to <a href="#track-muted">mute</a> the
-        track, and MUST NOT <a href="#track-muted">unmute</a> the track until
+        the UA SHOULD instead queue a task to <a href="#track-muted">mute</a> the
+        track, and not <a href="#track-muted">unmute</a> the track until
         the document <a data-cite="!HTML/#gains-focus">regains focus</a>.
         If reacquiring the device fails, the UA MUST
         <a href="#ends-nostop">end the track</a> (The UA MAY end it earlier
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -725,6 +725,33 @@ interface MediaStream : EventTarget {
         provided the UA sets it back to <code>true</code> as soon as any
         unstopped track connected to this device becomes un-muted or enabled
         again.</p>
+        <p>When a live track sourced by a device exposed by <a data-link-for=
+        "MediaDevices">getUserMedia()</a> becomes either muted or disabled,
+        and this brings <em>all</em> tracks connected to the device to be either
+        muted, disabled, or stopped, then the UA MUST relinquish the device
+        within 3 seconds while allowing time for a reasonably-observant user to
+        become aware of the transition. The UA MUST attempt to reacquire the
+        device as soon as any live track sourced by the device becomes both
+        unmuted and enabled again, provided the <var>document</var>
+        <a data-cite="!HTML/#gains-focus">has focus</a> at that time. If the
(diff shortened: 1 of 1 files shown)


youennf:
As I said in the issue thread, I am not sure @enabled is the right API for requesting to capture again after some time has passed. For instance, if the user agent decides to not allow to unmute the track, it can either decide to leave it muted or end the track, which requires the application to look for both events and maybe decide to call getUserMedia after some timeouts or after receiving the end event.

A promise-based API would seem more natural. We could gate it as we do for other APIs, for instance by requiring focus or user activation.
Also, it might be used for other purposes. For instance, Safari is muting a page track if user grants another web page access to the camera/microphone. This API could be used by the other page to request access again instead of calling getUserMedia again.

How about the following:
1. Leave @enabled as it is defined
2. Add void mute() that tells the user agent to set track.muted to true (guaranteed to succeed)
3. Add Promise unmute() that asks the user agent to set track.muted to false (no guarantee to succeed but guaranteed to have an answer).

jan-ivar:
I don't think we should introduce new APIs without a need. And I don't understand the need here.

> Add void mute() that tells the user agent to set track.muted to true (guaranteed to succeed)

Why would JS ever mute? Seems confusing.

I like our separation where [muted](https://w3c.github.io/mediacapture-main/getusermedia.html#track-muted) is controlled by the user agent and [enabled](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediastreamtrack-enabled) is controlled by JS.

We even leverage that in webrtc-pc.

> Add Promise unmute() that asks the user agent to set track.muted to false (no guarantee to succeed but guaranteed to have an answer).

If you need a promise, just do
```js
new Promise(r => track.onunmute = r);
```

jan-ivar:
There's no reason we couldn't gate `track.enabled = true` on user activation if we want.

I'm not saying we should, just that the shape of the API doesn't affect whether we can or not.

jan-ivar:
> Safari is muting a page track if user grants another web page access to the camera/microphone. This API could be used by the other page to request access again instead of calling getUserMedia again.

@youennf I tried [this fiddle](https://jsfiddle.net/jib1/Ltg3eqn1/) in two different Safari windows and I understand what you mean now.

But why doesn't Safari automatically unmute the first window when I close the second? Or at least when I focus the first window again?

I don't see why the application should need to ask when it didn't want to be muted in the first place. It seems obvious to me that it wants to be unmuted, so why not just assume it does?

youennf:
> There's no reason we couldn't gate `track.enabled = true` on user activation if we want.

We should first see whether that is web compatible.
We should also decide whether that is desirable for non capture tracks.
I do not see why we should require user activation for a web audio track or a remote audio track for instance.

It is also unclear to me whether we want to restart capturing for getDisplayMedia tracks.
Should there be an indicator when enabled=false? 

> But why doesn't Safari automatically unmute the first window when I close the second? Or at least when I focus the first window again?

The user can perfectly do that with the camera/mic icon to switch capture between pages.

> It seems obvious to me that it wants to be unmuted, so why not just assume it does?

I do not think that switching to a page that was capturing an hour ago is sufficient information to know whether the user wants to restart capturing for this page.
The intent is much clearer if user clicks on a button and the web page is asked to unmute.

Some more thoughts:

Safari camera/mic icon exposes the information of whether the page has a live capture track and if all tracks are muted or not. This proposal introduces a third variable (enabled) that should be taken into consideration in the computation of the page capture state, which does seem unnecessarily complex. It seems more consistent with how the spec is currently defined to piggy back on the muted state.

This PR is also not very clear in how should be treated the case where unmuting is denied by the user agent.
The typical case is a user agent denying recapture for a page that did not capture for several minutes/hours. It could also be that the user agent has a global switch that disables capture for all pages.
The current PR allows two possibilities: keep the track muted or end the track.
A promise-based approach allows to reject with various errors (device no longer there, access denied...).

jan-ivar:
> It is also unclear to me whether we want to restart capturing for getDisplayMedia tracks.

Screenshare has no observable device HW light, so I no reason to relinquish anything there.

> Should there be an indicator when enabled=false?

[Yes](https://w3c.github.io/mediacapture-screen-share/#privacy-indicator-requirements). But whether it needs to be any different from *Live* I think is up to user agents. Firefox implements a red/gray shift [consistent with](https://docs.google.com/presentation/d/1NIHzumglY9cYa4b7rcEbHGVsMam5BiY80VfFDB6cDjQ/edit?ts=5e2b3630#slide=id.g7c3d2d4d1e_7_190) its cam/mic indicators FWIW ([fiddle](https://jsfiddle.net/jib1/6f54yLp8/)).

In short, I think mediacapture-screen-share is fine and out of scope (let's discuss there not) here.

> > It seems obvious to me that it wants to be unmuted, so why not just assume it does?
>
> I do not think that switching to a page that was capturing an hour ago is sufficient information to know whether the user wants to restart capturing for this page.

I didn't say it should do that (although s/hour/seconds/ and I might, given that all other browsers support concurrent access here).

I only said we can assume what the page wants, because it's still holding onto a live track. With that knowledge, Safari should be able to figure this out with UX (e.g. provide said button) without involving the web page.

I'm in favor of Safari experimenting here (I particularly like the idea of an "hour" heuristic), provided it's done within the existing spec model. I.e. I'm not in favor of adding more API surface for it. Having web sites go through new APIs specifically for Safari would be a mistake IMHO.

With `mute`, `enabled` and `ended`, I hope we have sufficient surface to accommodate most UA controls.

jan-ivar:
The consensus at the [virtual interim](https://www.w3.org/2011/04/webrtc/wiki/February_27_2020) last Thursday was to merge this PR.

> This PR is also not very clear in how should be treated the case where unmuting is denied by the user agent.

By "unmuting" I assume you mean setting `track.enabled = true` which is all JS can do.

This has no impact on the track's [muted](https://w3c.github.io/mediacapture-main/getusermedia.html#track-muted) state which is _"out of control for the application"_.

> It could also be that the user agent has a global switch that disables capture for all pages.
The current PR allows two possibilities: keep the track muted or end the track.

This seems true even without this PR (errors don't appear only while muted).

> A promise-based approach allows to reject with various errors (device no longer there, access denied...).

We've discussed exposing errors on the `ended` event in the [past](https://lists.w3.org/Archives/Public/public-media-capture/2012May/0042.html) and rejected it.

youennf:
> Screenshare has no observable device HW light, so I no reason to relinquish anything there.

I might be missing something, but a device HW light does not seem like a good criteria.
Any OS level/user agent icon seems in scope to me.
Microphone seems in scope as well although there is often no HW light, at least for built-in mics.

> This has no impact on the track's [muted](https://w3c.github.io/mediacapture-main/getusermedia.html#track-muted) state which is _"out of control for the application"_.

It has impact on the track's muted state.
An application can actually change the muted state if it is capturing and not having focus by doing enabled=false/enabled=true.

IIRC, the spec currently does not mandate the user agent to mutate the muted state of a capture track. This PR breaks the assumption that the capture muted state is in full control of the user agent.

> Having web sites go through new APIs specifically for Safari would be a mistake IMHO.

This is not only about Safari, this PR tries to minimise the amount of changes to the existing API at the expense of convenience and backward compatibility.

If we add a user activation requirement for this, which I think we should do in this PR and not as a follow-up, we will use a pattern of setting an attribute based on user activation, but this requirement might only apply for a subset of all MediaStreamTrack objects. This seems like a new pattern.

If we add activation requirement, this API does not allow to tell whether the task was rejected because of lack of user activation or something else. Or we will add something like an exception, which might not be very backward compatible. Usually we try to provide this information, see for instance video.play() or device orientation requestPermission.

Setting this attribute might trigger a prompt, which might be a new pattern.
To know whether this async-by-nature task is completed, a web page might have to combine the information given by two event listeners. This also seems like a new pattern.

Web sites will have to adapt whatever the API we choose.
If a web page uses track cloning or requests several media stream tracks, the page will need to make sure to disable all tracks related to a given device to actually get the desired effect. Ditto for enable, potentially in the same runloop if we add the user activation requirement.
This seems error prone and not very web developer friendly.

Given the fact that the target of this PR is specifically about capture sources and MediaStreamTrack can have non-capture sources, I am even wondering whether MediaStreamTrack is the right place for this API.
From a model point of view, it seems this feature is something like 'mute that device' which is closer to MediaDeviceInfo for instance.

> The consensus at the [virtual interim](https://www.w3.org/2011/04/webrtc/wiki/February_27_2020) last Thursday was to merge this PR.

I was planning to attend but unfortunately had to miss it.
I guess I would have expressed the points stated above.
I hope they can be taken into consideration.

jan-ivar:
@youennf Sorry you missed the interim. The goal and scope was specifically to address known web compat problems around temporary camera and microphone disabling only. See [slides](https://docs.google.com/presentation/d/1NIHzumglY9cYa4b7rcEbHGVsMam5BiY80VfFDB6cDjQ/edit?ts=5e2b3630#slide=id.g7c3d2d4d1e_7_182). I'm happy to jump on a call to clarify.

> IIRC, the spec currently does not mandate the user agent to mutate the muted state of a capture track. This PR breaks the assumption that the capture muted state is in full control of the user agent.

Good point. I've updated the PR to reduce the mandate only to muting, not unmuting. I agree the user agent should remain in full control of unmuting.

The reason for mandating muting in this instance is to provide stronger guarantees around privacy, but it's peripheral to the main goal of the PR. Is SHOULD better?

> An application can actually change the muted state if it is capturing and not having focus by doing enabled=false/enabled=true.

True, though a page isn't generally in control of whether it has focus or not, so I don't think we need to change the sentence _"Muted is out of control for the application"_ over this...

There is no consensus for user activation or new APIs.
Track cloning is already well-defined, so their behavior in relation to this is a [feature](https://bugzil.la/1596959#c4).
There's [lots](https://w3c.github.io/mediacapture-main/getusermedia.html#life-cycle-and-media-flow) of [precedence](https://w3c.github.io/mediacapture-main/getusermedia.html#privacy-indicator-requirements) in Mediacapture-main for discussing camera/mic-specific tracks.     
Minimizing the amount of changes to an existing API is generally _helps_ backwards compatibility.

youennf:
> but it's peripheral to the main goal of the PR. Is SHOULD better?

SHOULD for muting sounds good to me.
This allows to both describe the problem in the spec and provide some concrete guidelines for implementors. And it should be easy to move back to a MUST when feasible.
For unmuting, I would personally go with MAY to give time for experimentation to validate or update the proposed approach.

jan-ivar:
I went with _"UA SHOULD instead queue a task to mute the track, and not queue a task to unmute it until the document regains focus"_

youennf:
I was maybe unclear. I am fine with a SHOULD for the whole PR, including relinquishing.
In general, this PR seems to go deep into User Agent and OS UI/UX. We usually try to provide guidelines in that area, not mandate behaviours.

jan-ivar:
> We usually try to provide guidelines in that area, not mandate behaviours.

Yes, and that failed miserably in this case, as I presented at the interim. The consensus was  to enforce web compat around behavior in all browsers on this.

jan-ivar:
I believe the meeting was recorded.

henbos:
Editors can integrate if you change it to a SHOULD. We need to make all the steps well defined (there is a difference between "relinquish device" and "relinquish permission", which could pose problems). We need to follow up on these issues though, in the end we want MUSTs not SHOULDs...

getusermedia.html
```diff
@@ -725,6 +725,36 @@ <h4>Life-cycle</h4>
         provided the UA sets it back to <code>true</code> as soon as any
         unstopped track connected to this device becomes un-muted or enabled
         again.</p>
+        <p>When a <code>live</code>, <a href="#track-muted">unmuted</a>, and
+        <a href="#track-enabled">enabled</a> track sourced by a device exposed
+        by <a data-link-for="MediaDevices">getUserMedia()</a> becomes either
+        <a href="#track-muted">muted</a> or <a href="#track-enabled">disabled</a>,
+        and this brings <em>all</em> tracks connected to the device to be either
```
youennf: 'all tracks' is very unclear.
Does it mean the browsing context? Probably not, since it might well be that two iframes may own two tracks from the same device. So this is probably in the context of a page.

jan-ivar: No it specifically means _all_ tracks connected to the device.

jan-ivar: Otherwise the user agent cannot relinquish the resource.

youennf: Right!
That said, PR is not very clear in the case of two pages capturing and both of them disabling all of their tracks. Then, one of them is restarting capture. What is expected to happen for the second page is unclear since the UA already reacquired the device. Ditto for the case of two iframes in the same page.

Probably the idea is that when the second page also wants to restart capture, the same 'wait-for-focus' requirement is done.

jan-ivar:
Travis appears to be failing over infra here? @dontcallmedom?
```
syntax error at support/link-checker/bin/checklink line 1342, near "$ulinks{"
```

getusermedia.html
jan-ivar: There is only one user agent, and these instructions are for that one user agent, so I don't think there's any race here.

I've added language to clarify the scope. PTAL @youennf

jan-ivar: On the muting intervention: note it doesn't actually say when to unmute, only when _**not**_ to unmute: _" the UA SHOULD instead queue a task to mute the track, and not queue a task to unmute it until the document regains focus."_

This is phrased as a behavioral requirement only, meant to give UAs lots of freedom, which was based on your concerns, I believe.

dontcallmedom:
@jan-ivar I've [disabled the failing script](https://github.com/w3c/webrtc-respec-ci/commit/4093a3556adc5a44d11920b227e4b2d793ef31d9) for now and will look at a better fix soon

PR #664, opened 2020-02-27
Added support for new "category" constraint for audio

sjdallst:
Contributers: +@gregwhitworth

Explainer: https://github.com/MicrosoftEdge/MSEdgeExplainers/tree/master/AudioStreamCategory

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1167,6 +1167,7 @@ interface MediaStreamTrack : EventTarget {
   boolean channelCount = true;
   boolean deviceId = true;
   boolean groupId = true;
+  boolean category = true;
 };</pre>
           <section>
             <h2>Dictionary <a class=
@@ -1257,6 +1258,12 @@ interface MediaStreamTrack : EventTarget {
               <code>true</code></dt>
               <dd>See <code><a href=
               "#def-constraint-groupId">groupId</a></code> for details.</dd>
+              <dt><dfn><code>category</code></dfn> of type <span class=
+              "idlMemberType">boolean</span>, defaulting to
+              <code>true</code></dt>
+              <dd>See <code><a href=
+              "#def-constraint-category">category</a></code> for
+              details.</dd>
             </dl>
           </section>
         </div>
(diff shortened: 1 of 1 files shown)


alvestrand:
This needs to explain how it relates to the content-hint extension; possibly it could be a PR on that specification instead if functionality there is missing.

sjdallst:
> This needs to explain how it relates to the content-hint extension; possibly it could be a PR on that specification instead if functionality there is missing.

I'm wondering if there is a world where the solution I'm looking for is a combination of both of these specs. For example, would it be consistent and clear with both specifications if I:

1. Changed the name of the constraint to something like "platformOptimization" (open to suggestions there).
2. Updated the content-hint spec with new hints such as "speech-Recognition", "speech-Communication", "raw", and then add default behaviors corresponding to these new constraints (ex: "raw" as a content-hint would apply {platformOptimization: "raw"} as a default).

This would follow current patterns, and help show that these new constraints are functionally like other constraints. Functionally alike, as in, they result in new, audible differences if supported by the platform, and can be mixed and matched with other constraints.

alvestrand:
The preferred approach is to open a bug explaining what problem you intend to solve, and link to that in the PR description. Then we can keep the discussion of "how can this problem be solved" and the discussion of "is this solution well specified" separate.

I still haven't figured out precisely what the processing you desire is; for instance, the term "the exact stream that the user's device provides" isn't a well defined term, given that many/most user devices can be opened in multiple modes, where what is being provided is different for each mode. Existing constraints were defined exactly to influence the mode in which the device is opened.

aboba:
@sjdallst Please file an Issue as Harald suggests, taking into account existing issues that have been filed against content-hints, such as: 

https://github.com/w3c/mst-content-hint/issues/28
https://github.com/w3c/mst-content-hint/issues/29
https://github.com/w3c/mst-content-hint/issues/30

aboba:
I believe that this PR will be reformulated as PR to content-hints.

sjdallst:
Yes, closing this PR as it is moving to content-hints. Will follow up with a link to the new PR.

Issue #639, opened 2019-11-04
Enforcing user gesture for getUserMedia


youennf:
Ideally getUserMedia would require a user gesture similarly to getDisplayMedia.
This is not web-compatible as many pages call getUserMedia on page load or quickly after page load.
It would still be nice to define web-compatible heuristics where user gesture could be enforced.

youennf:
One case where we could enforce user gesture is the following:
- User denied access to getUserMedia for a page, for instance at page load time or when entering a chat room
- User wants to grant microphone access and clicks a button on the web page
- getUserMedia is called as part of the button click handler.

We could try to enforce that a user gesture is required to call getUserMedia whenever user previously denied getUserMedia access for a given page.

jan-ivar:
> We could try to enforce that a user gesture is required to call getUserMedia whenever user previously denied getUserMedia access for a given page.

This (is permission-model specific, but) would be more permissive than [today](https://jan-ivar.github.io/dummy/gum.html): Safari and Firefox block temporarily, while Chrome blocks permanently. 

Why would we remove the only disincentive to spamming on pageload?

To me, the case for user activation has to do with navigating to a room on a site already granted persistent permission. An interim solution there would be to ignore persistent grants when user activation is not used. This would let sites migrate on their own time.

jan-ivar:
Complication: I'm sensitive to the number of clicks it takes to get into a call (one vs two), and would like to not penalize privacy sensitive users more (or browsers with privacy sensitive defaults).

How about something like:

_"The user agent MUST ignore persisted permissions unless the method was triggered by user activation"_?

This would mitigate the live cam room dive,  but also be highly backwards compatible (more prompts, but interestingly the same number of clicks).

alvestrand:
If the browser supports querying for whether you have permission or not, the page can tell that it's not getting the device because permission is "denied" - at least when it's for all devices.
In that case, it can point to browser UI (sigh) to reenable camera.

(Without enumerateDevices, he can't find the ID to figure out which specific devices to ask about, so if the individual device is denied, the page has a Hard Problem.)

If any page action allows the page to prompt after being initially denied, this offers the possibility of trapping the user in "request until allowed" loops - which is a Bad Thing. No needs to mean no.

dontcallmedom:
VI discussion: user-agent dependent at this stage, no interest in standardizing behavior on this at this stage

youennf:
Reopening issue.
[...]

PR #666, opened 2020-03-12
Require user activation to automatically grant getUserMedia calls

youennf:


youennf:
PR for https://github.com/w3c/mediacapture-main/issues/639

getusermedia.html
```diff
@@ -3667,6 +3667,12 @@ <h2>Methods</h2>
                           MediaStreamTrack that required the same level of
                           permission to obtain as what is being requested (e.g. not
                           isolated).</p>
+                          <p>If the <a href= "https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global">
+                          relevant global object</a> of <a href="https://heycam.github.io/webidl/#this">this</a> does not have
+                          <a href="https://html.spec.whatwg.org/multipage/interaction.html#transient-activation">
+                          transient activation</a>, the user agent MUST seek active user consent:
+                          if the <a href= "https://w3c.github.io/permissions/#permission-state">
+                          permission state</a> is "granted", the User Agent MUST proceed as if it is "prompt".</p>
```
alvestrand: This introduces extra prompts. App writers hate extra prompts.
Which cases will this cover?
The typical conference UI is "click here to enter conference" (in calendar or overview page), and the page you enter (a NEW page) then immediately requests camera. Will it have transient activation or not?

alvestrand: Note: the case of a page sitting quietly for 10 minutes after the user last touched it and then requesting the camera doesn't bother me that much - although it would, for instance, destroy the possibility of a time-lapse camera that doesn't hold the camera permanently. But "start a VC" is an use case that I want to have as smooth as possible.

youennf: > Will it have transient activation or not?

If you are in a native calendar and click on a link, then surely no.

In general, 'start a VC' should probably have a way to enter without camera/microphone or just microphone or just camera. I believe websites are moving in that direction, the proposed change might not be as disruptive as we might think.
Although this is currently an application design choice, it seems rude for a web page to start capture without any user gesture at all.

I guess there is the case of a page that does the set-up and when you click 'Join the room' , will open a new page that will do the real communication. I am not sure this is great experience (the LED will switch off and on for instance, you cannot prewarm the connection). Maybe transient activation can define cases where activation would be moved to the newly loaded document.

> the case of a page sitting quietly for 10 minutes after the user last touched it and then requesting the camera

That is the typical case that might be good breaking.
To implement time-lapse, page is fine keeping its track live and using enabled=true/enabled=false.

alvestrand:
@youennf is going to collect data.

dontcallmedom:
the underlying issue #639 has been moved to -extensions; @youennf is there any value in keeping that pull request open?

youennf:
> @youennf is there any value in keeping that pull request open?

No, let's close it for now

Issue #652, opened 2019-12-16
In-content device selection a mistake. Too complicated, leaks info


jan-ivar:
In hindsight, in-content device selection was a mistake. It's
 * **Too permissive**—assumes all devices granted up-front to work effectively (wo/reprompt)
 * **Too complicated**—having every site write a decent picker compatibly has been a failure:
    * Exhibit A: Changing camera **_or_** mic in [webrtc samples](https://webrtc.github.io/samples/src/content/devices/input-output/) re-prompts **_both_** in Firefox; flickers
    * Mobile devices typically can't open more than once device at a time ("stop-then-pick")
    * Stop-then-pick is an inferior user experience on desktops
    * Dealing with different browser permission models severely limits design (no previews)
 * **Leaks private info**—fails PING review https://github.com/w3c/mediacapture-main/issues/640
 * **Too limiting**—no path to privacy (can't avoid redundant re-prompts *after* user selection)

The PING outlines the way forward in https://github.com/w3c/mediacapture-main/issues/640#issuecomment-549540203:
> Privacy-by-default flow:
> 
> Initially site has access to no devices or device labels
> 
> 1. site asks for category (or categories) of device
> 2. browser prompts user for one, many or all devices
> 3. site gains access to only the device, and device label, of the hardware the user selects.

That's an in-chrome picker (_"in-chrome"_ = implemented in the browser). In-chrome pickers
 * **Are proven successful**—in [getDisplayMedia](https://jan-ivar.github.io/dummy/gdm.html)
 * **The way forward**—for speakers https://github.com/w3c/mediacapture-output/pull/86
 * **Remove the need to grant all devices** ❤️ 
 * **Let UAs solve mobile** platforms that can’t open multiple devices (mute temporarily)
 * **Let UAs solve camera previews** maybe even in a (non-creepy) Brady Bunch grid!

https://github.com/w3c/mediacapture-main/pull/644 is my proposal for reshaping `getUserMedia` to serve this need, as well as solve https://github.com/w3c/mediacapture-main/issues/648.

guest271314:
If gather the issue correctly, could not a `<select multiple>` or equivalent UI be used to select multiple devices, where such implementation can be uniform across implementations?

```
  <form>
    <select multiple>
      <option value="camera_1">Camera 1</option>
      <option value="camera_2">Camera 2</option>
      <option value="audio_input">Audio input</option>
      <option value="audio_output">Audio output</option>
    </select>
    <input type="submit">
[...]

PR #667, opened 2020-03-18
Add semantics choice to getUserMedia() for in-browser picker.

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/652


aboba:
A few comments: 

* The PR is not clear about transition impacts. Presumably, the transition plan is for `GetUserMediaSemantics` to default to `user-chooses` initially, then eventually to not permit `browser-chooses`?  How does the application determine what choices are available? `feature-detect` only seems to help determine whether `GetUserMediaSemantics` is available at all.
* Reading the PR, I wasn't sure whether "user-chooses" always results in a prompt.  If this is the intent, it should be more explicit.
* "Narrow" might be a better word choice than "widdle".

getusermedia.html
```diff
@@ -3833,9 +3866,79 @@ <h2>Dictionary <a class="idlType">MediaStreamConstraints</a>
...
+            </dd>
           </dl>
         </section>
       </div>
+      <div>
+        <pre class="idl"
+>enum GetUserMediaSemantics {
+  "browser-chooses",
+  "user-chooses",
+  "feature-detect",
```
henbos: Not sure I like a "feature-detect" as part of the API. What if we expose a static defaultSemantics somewhere that says if the default is "browser-chooses" or "user-chooses"? Kind of like a capability.

Its presence would act as a feature detection that the browser is aware of semantics, and its value would tell you whether or not the default has switched yet. Then you can feature detect what phase of the transition plan the browser is as well.

At some point we mandate the default must be "user-chooses", but initially it would be "browser-chooses".

getusermedia.html
```diff
@@ -3833,9 +3866,79 @@ <h2>Dictionary <a class="idlType">MediaStreamConstraints</a>
...
+                </p>
+              </td>
+            </tr>
+            <tr>
+              <td><dfn><code id=
+              "idl-def-GetUserMediaSemantics.user-chooses">user-chooses</code></dfn></td>
+              <td>
+                <p>When application-specified constraints do not widdle
+                multiple choices between devices down to one per kind, the user
+                agent MUST
```
henbos: and if it doesn't? Is prompting in that case a MUST NOT, a SHOULD NOT or MAY?

getusermedia.html
```diff
@@ -3682,41 +3701,54 @@ <h2>Methods</h2>
...
+                        <li>
+                          <p>If the number of unique devices sourcing tracks of
+                          media type <var>kind</var> in <var>candidateSet</var>
+                          is greater than <code>1</code> and
+                          <var>semantics</var> is <code>"user-chooses"</code>,
+                          then <a>prompt the user to choose</a> a device with
+                          <var>descriptor</var>, resulting in provided media.
+                          Otherwise, <a>request permission to use</a> a
+                          device with <var>descriptor</var>, while considering
+                          all devices being attached to a <code>live</code> and
```
henbos: What does "considering all devices [bla bla] to mean having permission status 'granted'" mean?
I assume this has to do with possibly not re-prompting?

I can't tell if the desired behavior of "considering devices" entails MUSTs, SHOULDs, or MAYs with regards to whether or not prompts happen.

getusermedia.html
```diff
@@ -3682,41 +3701,54 @@ <h2>Methods</h2>
...
-                          <p>User Agents are encouraged to default to using the
-                          user's primary or system default device for <var>kind</var>
-                          (when possible). User Agents
+                          algorithm, the value of <var>semanticsPresent</var>,
+                          or any other internally-available information about
+                          the devices, as inputs to its decision.
+                          If <var>semantics</var> is <code>"user-chooses"</code>,
+                          and the application has not narrowed down the choices
+                          to one, then the user agent MUST ask the user to make
+                          the final selection.</p>
```
henbos: And if semantics is "user-chooses" but the user agent has not narrowed down the choices?

getusermedia.html
```diff
@@ -3833,9 +3866,79 @@ <h2>Dictionary <a class="idlType">MediaStreamConstraints</a>
...
+              </td>
+            </tr>
+          </tbody>
+        </table>
+        <div class="note">
+          <p>Support for <code>semantics</code> may be detected as follows:</p>
+          <pre>
+async function supportsSemantics() {
+  try {
+    await navigator.mediaDevices.getUserMedia({semantics: "feature-detect"});
```
henbos: If (navigator.mediaDevices.defaultSemantics != undefined) {
  console.log('Semantics is not supported.');
} else {
  console.log('Semantics is supported, the default value is: ' + 
      navigator.mediaDevices.defaultSemantics);
}

jan-ivar:
> * Presumably, the transition plan is for GetUserMediaSemantics to default to user-chooses initially, then eventually to not permit browser-chooses

As I recall, we did not reach consensus on a transition plan last virtual interim, just [option B](https://docs.google.com/presentation/d/1NIHzumglY9cYa4b7rcEbHGVsMam5BiY80VfFDB6cDjQ/edit?ts=5e2b3630#slide=id.g7ead6f3266_5_70) for now.

So there are no transition impacts from this PR, since it effectively defaults to `"browser-chooses"` for backwards compatibility (yet allows Firefox to prompt like it does today).

This is the minimal consensus needed to replace labels and in-content selection.

> How does the application determine what choices are available?

Like it does today, count devices per kind using *enumerateDevices*. Although if constraints are used, this might narrow choices, and it could be hard to detect. Right now I think the API would succeed right away without a prompt in that case (assuming you already have access to the current device). Would you like it to fail? That would be a bigger semantic difference.

The feature detection is there to let us start using this before labels go away.

> Reading the PR, I wasn't sure whether "user-chooses" always results in a prompt.

It does not. Only when there's a choice. Otherwise, only a permission prompt is mandated.

getusermedia.html
jan-ivar: "Otherwise," below kicks in. There wasn't a MUST there before, but I can add one.

henbos: It would be nice if we got as close as possible to not letting implementation-specific logic decide whether or not to prompt. Like with getDisplayMedia()... it always prompts. Can getUserMedia() always prompt no matter what?

When you lost audio during the meeting @youennf made a point of what if the user wants to share only audio and not video or vice versa? That may require further discussion but I like the idea of "user-chooses" to be "audio and/or video" so that you never have to reject and then re-prompt.

henbos: Idea in the future "user-chooses" world.
- getUserMedia() always prompts. It gives audio and/or video, depending on user choice. "kind" is optional.
- If you want to change device mid-call, track.changeUserMedia() (or track.applyConstraints("user-chooses") if that makes transition easier) re-prompts the user for audio and/or video, depending on user choice.

getusermedia.html
jan-ivar: > I assume this has to do with possibly not re-prompting?

Yes. The [live](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediastreamtrackstate-live) permission-exemption lives here and not in the permission spec, because it's unique to our model where a page can call *getUserMedia* multiple times asking for the same track(s), and we did not want another permission prompt then. As I recall, this was before stream.[clone](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediastream-clone)() existed, which is arguably the better pattern for doing that IMHO.

It's why we need `semantics` now (more specifically: it's the main reason Firefox's existing behavior cannot reliably be used to replace in content device selection today).

> I can't tell if the desired behavior of "considering devices" entails MUSTs, SHOULDs, or MAYs with regards to whether or not prompts happen.

This is existing language we have today effectively. It just got moved around a bit.

Isn't all prose normative, even without MUST? Can we open a separate issue?

getusermedia.html
jan-ivar: This is a modifier, so I thought it was sufficient to describe the modification. Open to wording ideas.

getusermedia.html
henbos: My concern is not about "MUST consider devices" versus "do consider devices". Either way of phrasing is normative, I'm not nit-picking on choice of words.

I'm saying I can't tell because the act of "considering devices" sounds vague to me. Maybe this is well-defined if I do my homework and read more of the spec?

But from reviewing this PR I couldn't tell.

So spec language aside, I'm asking you, what is the intent here:
- When is it mandatory to prompt?
- When is it up to the implementation to decide whether or not to prompt?
- If I use "user-chooses", is there any case where it is not mandatory to prompt the user? If so, which cases?

Let's agree on the intent before worrying about how to phrase it.

getusermedia.html
jan-ivar: > Can getUserMedia() always prompt no matter what?

No-one wants that. Users don't want to be prompted for their camera every time.

Many have only one camera, thus no need for a picker at all (ditto microphone).

Even with `"user-chooses"` (a semantic we aspire to default to as sole modus operandi), it remains up to the app to request what it wants: a _specific device_, or _any device fitting its constraints_. A user agent picker would kick in (regardless of permission) only when that request meets a user with multiple valid devices meeting those constraints.

I think that's the right division of concerns. App needs meet user needs.

The following then guarantees a picker, e.g. letting us replace in-content selection:
```js
if (devices.filter(({kind}) => kind == "videoinput").length > 1) {
  await navigator.mediaDevices.getUserMedia({
    video: true, semantics: "user-chooses"
  });
}
``` 

> what if the user wants to share only audio and not video or vice versa?

Firefox used to do this, but we had to stop because web sites were expecting two tracks. I'd argue it's an inherent problem with the API, not solvable in the spec.

In fact, the [spec algorithm](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-getusermedia) is already written as concurrent steps per _kind_ that _"User agents are encouraged to bundle ... into a single user-facing permission prompt."_

In fact, Hangouts already asks for camera and microphone separately, and I have a bug filed on Firefox somewhere to combine these requests into one prompt. That problem goes away then.

getusermedia.html
jan-ivar: Again, the "consider" language predates this PR, so I'm going to close this as resolved. @henbos PTAL if the latest version addresses your other concerns.

jan-ivar: I hope my comments from 5 minutes ago address how it works.

jan-ivar:
> A user agent picker would kick in (regardless of permission) only when that request encounters a user with multiple valid devices meeting those constraints.

...otherwise users would instead see a _permission prompt_ if permission is needed, or no prompt at all if the app has persistent permission.

jan-ivar:
**Summary of PR:** Based on consensus in February call (Option A/B in [slides](https://docs.google.com/presentation/d/1NIHzumglY9cYa4b7rcEbHGVsMam5BiY80VfFDB6cDjQ/edit#slide=id.g7ead6f3266_5_70)). New semantics:
```js
navigator.mediaDevices.getUserMedia({video: constraints, semantics: "user-chooses"});
```
...to work as follows:
 * Users with 1 device satisfying the constraints: permission prompt only if needed (like today). 
 * Users with >1 device satisfying the constraints: **_browser_** opens device picker always.

Sites avoid picker on revisits by remembering w/`deviceId` constraint to narrow constraints to 1.

Use this to replace today's in-content device pickers. May also be used on initial call.

a2sheppy:
@jan-ivar Please be sure any bug around this gets DDN, or file a bug against the MDN content (https://github.com/mdn/sprints/issues/new/choose) so we can be sure this change is documented.

getify:
I just wanted to chime in that I have used a number of web-based video chatting apps over the last few weeks that don't offer me any choice of my camera (my setup has multiple cameras). Of course, those sites *could* have used APIs to enumerate my cameras and let me pick, but they didn't. And since that's left to the code, I the user am given no choice in the matter.

I really wish a change could be made that would, without any changes to existing code, give me the option to pick or authorize a different camera. In other words, what I wish is that this "user-chooses" semantics was the default.

Even if the browser UI still only let me pick/authorize a single camera (and not giving me the option to select/authorize multiple cameras and mics), at least I as a user would have the option in the browser to choose to force-expose a specific camera to the site rather than them getting to pick my default camera.

guest271314:
@getify Is the browser tested Chrome or Chromium?

getify:
Chrome (windows).

It happens to me every other day or so, I have to connect to webchat service (like telehealth, whatever) and their UI doesn't let me choose what mic or camera I want... I assume JS wise they just grab the first camera and mic, but from my perspective it feels random... sometimes it's the right one of each, sometimes it's the right camera wrong mic, sometimes right mic wrong camera, sometimes both are wrong.

Even though I go out of my way to set the default mic in windows, that's not always the mic that such a webchat app tries to pick up. That boggles my mind. But worse: my windows laptop doesn't let me set a "default camera" -- it just assumes my built-in camera is what I want to use -- though it almost never is!

Chrome doesn't offer me any UI option to select which camera or mic are exposed to a web page. All I can do is just turn on or off camera/mic, but not select.

If this "user-chooses" semantic isn't the made the default, then this feature won't do anything to help me, because all those poorly designed webchat apps will never update their code to turn it on -- either they don't know about it or don't care.

guest271314:
@getify Have not used Windows is some time. FWIW a workaround for Linux https://github.com/w3c/mediacapture-main/issues/693#issuecomment-643283729 to select _audio_ devices other than default microphone. If Native File System is not enabled by default you can launch Chrome with `--enable-experimental-web-platform-features` then use the same approach at a Windows shell script with a port of `inotify-tools` e.g., https://github.com/thekid/inotify-win to launch a native application to change the _video_ device after `getUserMedia()` has permissions, if Windows has such a display or device selection native application.

guest271314:
@getify Per https://answers.microsoft.com/en-us/windows/forum/windows_10-hardware/how-to-change-default-webcam-in-windows-10/1e8377c8-4fd5-4e9c-8ab6-67913c5c8e0c you should be able to launch Control Panel to change the camera device and per https://docs.microsoft.com/en-us/windows/uwp/audio-video-camera/camera-profiles it should be possible to change the device progammatically. Either program can be launched in the browser session using the linked approach. Have not found any other means to change the default audio or camera at Chromium _other than_ using a fake media stream and supplying the data thereto by way of a file.

guest271314:
@getify Another alternative is to file an issue with the service to resolve

>  and their UI doesn't let me choose what mic or camera I want... I 

which should be possible utilizing the current iteration (using `label`s of `enumerateDevices()`) and selecting device by `deviceId`. To test hot-plugged a USB camera labeled `"Digital_Camera (<UUID>)"`

navigator.mediaDevices.getUserMedia({video: true})
.then(async stream => {
  const {deviceId} = devices.find(({label}) => label === 'Digital_Camera (093a:2700)')
  let [track] = stream.getVideoTracks();
  console.log(await track.getConstraints());
  track.stop();
  stream = await navigator.mediaDevices.getUserMedia({video: {deviceId: {exact: deviceId}}});
  [track] = stream.getVideoTracks();
  console.log(await track.getConstraints());
  const v = document.createElement('video');
  v.controls = v.autoplay = true;
  document.body.appendChild(v);
  v.srcObject = stream;           
});

guest271314:
> Chrome doesn't offer me any UI option to select which camera or mic are exposed to a web page. All I can do is just turn on or off camera/mic, but not select.
> 
> If this "user-chooses" semantic isn't made the default, then this feature won't do anything to help me, because all those poorly designed webchat apps will never update their code to turn it on -- either they don't know about it or don't care.



Agree. 

Not an issue at Firefox or Nightly.

Perhaps Chromium authors will listen to you. 

Have been asking for this feature for some time in various ways, too many to list here, e.g., https://bugs.chromium.org/p/chromium/issues/detail?id=1032815; https://bugs.chromium.org/p/chromium/issues/detail?id=865799;  for reasoning at  Chrome, Chromium see https://bugs.chromium.org/p/chromium/issues/detail?id=1013881#c9. 

Eventually decided to stop asking at specifications or implementers and create code that achieves the expected result.

jan-ivar:
Moved to https://github.com/w3c/mediacapture-extensions/pull/9.

Issue #598, opened 2019-05-23
Should deviceId be partitioned by top level origin and document origin


youennf:
WebKit is currently partitioning deviceId by top level origin and document origin to prevent tracking. On going work is done to define such kind of partitioning for the HTTP cache and there are thoughts to extend it to other types of storage like service workers and IDB.

alvestrand:
Firefox used to use double keying (top level & document) for permissions. The spec was written to permit both patterns to be conformant. I believe double keying has been abandoned by all - instead, we changed the default delegation to cross-origin iframes to "block" - one has to explicitly delegate permissions to an iframe if you want it to be able to use the invoker's permissisons.

I'm not sure I understand the implications of double keying on device ID - would that mean that a conferencing app that opened in an iframe on multiple sites would be unable to store device IDs between sessions on different top level origins?

youennf:
Agreed for permissions. WebKit is requesting decision on behalf of the top level origin, which is the one the user is most aware of.

For device ID, that would mean that a conferencing app opened in an iframe on multiple sites would see different device IDs for the same device. This makes most sense if you partition other storages like IDB.

Ideally, a partitioning concept would be defined for HTTP cache and other storages and we would just need to adopt it in this spec. I filed this issue to not forget about it and since we are touching deviceId definition in https://github.com/w3c/mediacapture-main/pull/587.

alvestrand:
At the moment, no change in the spec seems required.

alvestrand:
Repurposing the "Icebox" label for things that we might want to look at sooner or later due to things changing in the larger Web environment, but no change to spec needed at this time.

alvestrand:
Since this was re-raised by PING at TPAC 2019, and the resolution of the group seemed to be that this change was warranted, I'm taking "icebox" off again.

The relevant (proposed) privacy principle is here:

https://github.com/michaelkleber/privacy-model#identity-is-partitioned-by-first-party-site

PR #674, opened 2020-04-02
Have deviceId follow partitioning rules for storage (such as localStorage)

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/598. @annevk PTAL.


youennf:
Can we refer to https://storage.spec.whatwg.org/ directly then?

jan-ivar:
We should, but it looks like doing so is blocked on https://github.com/whatwg/storage/issues/18, so I think we should merge this for now and open an issue to come back to it.

Issue #683, opened 2020-04-07
Device info Access Control Model needs an update


dontcallmedom:
[9.2.1 Access control model](https://w3c.github.io/mediacapture-main/#access-control-model) says that before permission, `deviceId` and `groupId` are available, which is no longer the case. The section should also be updated to indicate that the number of devices available is no longer given before permission (only if there is one of a given type)

youennf:
MediaDeviceInfo is an interface so all deviceId, groupId, label and kind would be defined but their value would be the empty string.
Agreed this section needs to be updated to more clearly describe the new algorithm.

PR #686, opened 2020-04-23
Update the description of the DeviceInfo Access model given tigthening made to enumerateDevices

youennf:
Fixes https://github.com/w3c/mediacapture-main/issues/683

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2952,13 +2952,12 @@ interface MediaDevices : EventTarget {
       <section>
         <h2>Access control model</h2>
         <p>The algorithm described above means that the access to media device
-        information depends on whether or not permission has been granted to
-        the page's origin.</p>
-        <p>If no such access has been granted, the
-        {{MediaDeviceInfo}} object will contain the
-        {{MediaDeviceInfo/deviceId}}, {{MediaDeviceInfo/kind}}, and {{MediaDeviceInfo/groupId}}.</p>
-        <p>If access has been granted for a media device, the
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -2952,13 +2952,12 @@ <h2>Methods</h2>
...
-        information depends on whether or not permission has been granted to
-        the page's origin.</p>
-        <p>If no such access has been granted, the
-        {{MediaDeviceInfo}} object will contain the
-        {{MediaDeviceInfo/deviceId}}, {{MediaDeviceInfo/kind}}, and {{MediaDeviceInfo/groupId}}.</p>
-        <p>If access has been granted for a media device, the
-        <dfn>MediaDeviceInfo</dfn> object will contain the {{MediaDeviceInfo/deviceId}}, {{MediaDeviceInfo/kind}},
+        information depends on whether or not the browsing context did capture.</p>
+        <p>If the browsing context did not capture (i.e. {{MediaDevices/getUserMedia()}} was not called or never resolved successfully), the
+        {{MediaDeviceInfo}} object will contain a valid value for {{MediaDeviceInfo/kind}} but empty strings
```
dontcallmedom, line 2957: isn't it also the case that enumerateDevices() will return at most one entry for each kind (i.e. not expose the real number of items in the list)?
per
> If filteredList contains a MediaDeviceInfo object whose kind is the same as deviceInfo.kind, skip deviceInfo.

youennf: Will add it in this PR.
I wonder though whether we should not just get rid off the whole section given it is redundant with the above algorithm.

dontcallmedom: Given how hard it is to make sense of the algorithm, I think having a human readable interpretation is really useful (for spec reviewers, documentation writers)

Issue #682, opened 2020-04-07
Stop recommending UUID for deviceId/groupId


dontcallmedom:
Because UUID are unique, and until we have double-keyed storage widely available, having deviceId (and groupId) being UUID creates a tracking opportunity for `getUserMedia()` callers embedded as third-party iframes.

I think having deviceIds be simple monotonically incremented integers should be sufficient for our use case. I'm not sure how we need to be here, beyond no longer recommending to use UUIDs as ids.

/cc @pes10k since we were discussing this while reviewing progress on privacy-related work on the spec

youennf:
Related to or a dupe of https://github.com/w3c/mediacapture-main/issues/607.

> deviceIds be simple monotonically incremented integers should be sufficient for our use case

This has been discussed in the past.
If you are incrementing integers and try to persist them so that they can be used by web pages across navigations, these integers will end up becoming a tracker for those users that use non built-in camera/microphone devices. These integers might be unique to the user and not even tied to an origin.

If you do not have double-keyed storage, users are anyway being tracked with or without device Ids (given IDs are regenerated anytime website data like IDB is cleared). The spec now mentions double keyed storage, maybe the wording should be made stronger?

Note also that the spec now mandates device Ids to only be exposed after the page is capturing.
This makes it difficult to be used as a tracker in practice. And once page is capturing, device IDs are probably not the most personal information that is leaking to the page.

dontcallmedom:
thanks @youennf for clarifying the current thinking on that particular sub-issue.

@pes10k - wdyt?

pes10k:
> these integers will end up becoming a tracker for those users

I don't believe this is correct.  If you are reusing the same identifiers across users, they're by definition not identifying that user :) They're only "identifying" when you already have an identifying key to join against (in which case, all bets are off). 

The goal of making these less identifying is:

1) to remove more foot guns for vendors who are managing storage for users (for privacy reasons) in ways that aren't easily compatible with the "all or nothing" definitions keyed off in the spec.  The different implementations of Storage Access API do this to a degree (especially in storage upgrade cases), Safari's ITP does this in places, Brave does this in places, etc.  There are places where storage is cleared / changed, and not having unique identifiers here makes it much easier to reason about the privacy boundaries in such cases (and to prevent the device ids from being the unique key to join tracking session data).

2) In general, we should have a hard line about adding unique identifiers into the platform.  This is the only case im aware of where this is, and it would be *very* good to address this too.

[...]

PR #687, opened 2020-04-24
Remove UUID recommendation for deviceId and groupId.

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/682.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3750,7 +3750,7 @@ interface InputDeviceInfo : MediaDeviceInfo {
         <span class="practicelab" id="handling-devices">Generating deviceIds</span>
         <p class="practicedesc">An efficient practice for generating a
         {{MediaDeviceInfo/deviceId}} is to generate a cryptographic hash from
-        a generated privately key + (origin or origin + top-level origin, based
+        a generated private key + (origin or origin + top-level origin, based
         on the user agents' partitioning rules) + salt + device's underlying
         (hardware) id in the driver, presented as an alphanumeric string. Using
         32 bits or lower is recommended, but not much lower, to avoid risk of
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3752,10 +3752,11 @@ interface InputDeviceInfo : MediaDeviceInfo {
         {{MediaDeviceInfo/deviceId}} is to generate a cryptographic hash from
         a private key + (origin or origin + top-level origin, based on the user
         agents' partitioning rules) + salt + device's underlying (hardware) id
-        in the driver, presented as an alphanumeric string. Using 32 bits or
-        lower is recommended, but not much lower, to avoid risk of collision.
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -3022,10 +3016,8 @@ <h2>Attributes</h2>
               documents, MUST follow the user agents' partitioning rules for
               storage (such as {{WindowLocalStorage/localStorage}}), if any,
               to not interfere with mitigations for cross-site correlation.
-              The identifier SHOULD be origin-unique,
-              in which case some sort of UUID is recommended.
-              If the identifier can uniquely identify the user, as is usually
-              the case with UUID, it MUST be un-guessable by browsing contexts
+              The identifier SHOULD be origin-unique. If the identifier can uniquely
```
youennf, line 3019: It is a bit sad that we use "SHOULD be origin-unique" and "MUST be un-guessable" without providing any implementation guideline. Maybe we should briefly describe what all browsers seem to do.

jan-ivar: @youennf before https://github.com/w3c/mediacapture-main/commit/d3eb1302f0379f4e9a9f4a25e57b873a33b14fcb it used to say _"MUST ... be different for other origins"_ and _"MUST be un-guessable"_. What would you like it to say?

jan-ivar: Looks like SHOULD came from trying to allow id's like `"default"`. So really two forms of id:
 1.  Can uniquely identify user: MUST be un-guessable and origin-unique
 2. Cannot uniquely identify the user: MUST be un-guessable, but not origin-unique.

Seeing if I can reword

jan-ivar: Oh wait, I take that back, since Peter's idea would not necessarily be unique per origin.

jan-ivar: How about we simply remove _"SHOULD be origin-unique"_ since uniquely identifying an origin is actually NOT a desired property? We already have "un-guessable" which is the desired property I think.

getusermedia.html
```diff
@@ -3754,6 +3746,22 @@ <h2>Implementation Suggestions</h2>
...
+        <p class="practicedesc">An efficient practice for generating a
+        {{MediaDeviceInfo/deviceId}} is to generate a cryptographic hash from
+        a private key + (origin or origin + top-level origin, based on the user
+        agents' partitioning rules) + salt + device's underlying (hardware) id
+        in the driver, and present the resulting hash as an alphanumeric string.
+        Using 32 bits or fewer for the hash is recommended, but not much lower,
+        to avoid risk of collision.
+        </p>
+        <p class="practicedesc">A better, lower-entropy, alternative, at the
+        cost storage, is to assign the numbers 0 through 255 randomly to each
```
alvestrand, line 3760: I'd suggest changing "A better, lower-entropy, alternative" to "A lower-entropy alternative".
English: "At the cost storage" should be "at the cost of storage".

Issue #688, opened 2020-04-24
Only fire devicechange event when devices physically added/removed


jan-ivar:
The primary use case for `devicechange` may potentially be undermined by https://github.com/w3c/mediacapture-main/pull/632.

**Primary use case:** Inserting a USB device or enabling a bluetooth headset (e.g. putting on AirPods) during a call, is a **strong signal** users want to switch to it, that many apps want, like a key press.

Unfortunately, with https://github.com/w3c/mediacapture-main/pull/632 Safari is firing `devicechange` immediately upon `getUserMedia` success, for all users with more than one camera and microphone, because the list technically changed.

Problem: as a web developer, I can no longer differentiate between:
 1. The user just inserted or enabled a headset, which I want to switch to immediately.
 2. The user did not insert or enable anything, and I _shouldn't_ switch to a secondary device.

Today, [devicechange](https://w3c.github.io/mediacapture-main/getusermedia.html#event-mediadevices-devicechange) says: *"The set of media devices, available to the User Agent, has changed."*

Note is says: _"available to the **User Agent**"_, not _"the application"_.

The latter would support firing the event whenever the JS-exposed list changes. The former not.

The spec requires there be a change in device availability _to the user agent itself_.

I read that to mean the `devicechange` event MUST NOT fire solely from user agent decisions.

Do we need to clarify the language around this to ensure web compat?

jan-ivar:
Updated with the right link.

youennf:
Thanks for the raising issue.

I am not aware of any compat issue with Safari current behavior.

This is done as you described in Safari to limit compat risks with other browsers that were not doing the filtering. It seems the right time now that the spec mandates the filtering to get consistency here.

Some web applications may buffer enumerateDevices result and only update it based on the device change event. Firing the event after the first successful getUserMedia call ensures that web applications do not stick with a filtered list of devices.

As of detecting whether there is a new device or not, this is easy to determine.
In the new device detection algorithm, if the previous list of devices is empty or devices have empty device ids, there is no new device.

As of whether the spec can mandate to not fire an event, I am not sure it actually can.
Before the first successful getUserMedia call, there is no need for the user agent to retrieve and store the full list of devices when enumerateDevices is called.

guest271314:
`devicechange` is fired when setting a fake media stream at Chromium, without inserting a physical device to the machine. That is useful. The user is notified that the stream is Fake.

[...]

PR #689, opened 2020-04-28
Clear [[storedDeviceList]] on setting [[canExposeDeviceInfo]] to true

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/688.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2663,7 +2663,7 @@ interface MediaStreamTrackEvent : Event {
             </li>
             <li>
               <p>Create one internal slot: <dfn>[[\storedDeviceList]]</dfn>, initialized
-              to null.</p>
+              to <code>null</code>.</p>
             </li>
             <li>
               <p>Create one internal slot: <dfn>[[\canExposeDeviceInfo]]</dfn>, initialized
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -2892,15 +2893,10 @@ <h2>Methods</h2>
...
-                          browsing context, set <var>list-permission</var> to
-                          <code>"granted"</code>, otherwise set <var>list-permission</var>
-                          to <code>"granted"</code> if <a>[[\canExposeDeviceInfo]]</a> is
-                          <code>true</code> and <code>"denied"</code> otherwise.</p>
-                        </li>
-                        <li>
-                          <p>If <var>list-permission</var> is <code>"granted"</code>,
+                          <p>If <a>[[\canExposeDeviceInfo]]</a> is
+                          <code>true</code>, or if any of the local devices are
+                          attached to a live MediaStreamTrack in the current
```
youennf, line 2898: This is pre-existing text but I always found this text a bit odd.
If we have a live capture track, it means getUserMedia was granted so canExposedeviceInfo is set to true.

The only case would be that of User Agent calling set the device information exposure to false some time after getUserMedia succeeded. But it does not make any real sense to do that if there are live tracks.

Maybe we could just remove this 'if any of the ...' check?

jan-ivar: @youennf I considered that, but i assumed it was there because of [this note](https://w3c.github.io/mediacapture-main/getusermedia.html#set-device-information-exposure-0):

> To set the device information exposure, with a value of type boolean, run the following steps:
> 1. Set [[canExposeDeviceInfo]] to value.
>
> NOTE
> A User Agent MAY at any point set the device information exposure back to `false`, for instance if the User Agent decides to revoke device access on a given browsing context."*

As currently written, this wouldn't override a page's access if it already has a live track. what you propose would.

jan-ivar: Said differently, the current spec forbids removing the device info without also revoking permission, which I think makes sense.

Issue #670, opened 2020-03-23
What happens when laptop lid is closed / phone display is turned off?


jan-ivar:
Seperate from https://github.com/w3c/mediacapture-main/issues/668 because, as mentioned in https://github.com/w3c/webrtc-pc/issues/2489#issuecomment-602879082:
> 1. Closing a laptop lid is _**not**_ the same as freezing. It may _trigger_ it, but was observed not to.
> 2. It's a physical act that (sans external display) seizes display, audio output, mic & cam capture.

This physical act seems similar (but not identical) to pressing the physical "display off" button on most smartphones and tablets, or closing a flip phone.

To complicate matters, there appear to be several device exceptions:
1. A laptop with external monitor attached may not seize media i/o.
2. A flip phone with back-display may have its back camera work as front camera when closed.
3. Some phone apps (like YouTube now) still play audio while screen is off (with or wo/headset)

But one commonality seems to be that most devices have some way to shut media playback up temporarily, and resume later. Should we piggyback media capture on that?

Secondly, we already have the `mute` and `unmute` user agent events for notifying JS. But from my observations, browsers aren't firing them when I close and reopen the laptop lid, even though capture is interrupted. Do we need to say anything more, or simply file bugs on implementations?

Thirdly, there may be common privacy issues around this behavior. E.g. Users may be surprised
 1. if microphone keeps recording when they switch display off phone/tablet to put in pocket.
 2. if laptop microphone keeps recording when they close the laptop lid.
 3. USB camera keeps recording when they close the laptop lid.
 4. If camera and mic (and screen-sharing?) resume during laptop lock screen on reopen
 5. If camera and mic don't shut off when audible playback shuts off.
 6. If camera and mic resumes before audible playback resumes.

Do we want to try to say anything around that?

rshpount:
The physical "display off" button on smartphones does not seem to work correctly either. If you are on the call via WebRTC in Chrome on Android, switching display off stops microphone and playback a short time after the screen is turned of. At the same time, if you are on the regular phone call, switching display off does not disconnect the call or mutes the microphone.  This is one of the common customer complaints from people using our WebRTC application from smartphones.

jan-ivar:
Good point that user expectations for "display off" on phones/tablets may differ from laptop lid.

guest271314:
> Do we need to say anything more, or simply file bugs on implementations?

What is the bug? 

[...]

PR #700, opened 2020-06-25
Mention closing a laptop lid w/camera as a mute reason.

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/670.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -800,9 +800,9 @@ interface MediaStream : EventTarget {
         to the associated events {{mute}} and {{unmute}}. There can be
         several reasons for a {{MediaStreamTrack}} to be muted:
         the user pushing a physical mute button on the microphone, the user
-        toggling a control in the operating system, the user clicking a mute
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -800,9 +800,9 @@ <h4>Media Flow</h4>
         to the associated events {{mute}} and {{unmute}}. There can be
         several reasons for a {{MediaStreamTrack}} to be muted:
         the user pushing a physical mute button on the microphone, the user
-        toggling a control in the operating system, the user clicking a mute
-        button in the browser chrome, the User Agent (on behalf of the user)
-        mutes, etc.</p>
+        closing a laptop lid with an embedded camera, the user toggling a
```
youennf, line 803: I would guess that 'closing the lid' might well mute a microphone even if there is no camera.
Can we remove ''with an embedded camera'?

jan-ivar: Yes, this list is not exhaustive, just trying to list uncontroversial examples

PR #706, opened 2020-07-15
Clarify label may have no identifying info.

jan-ivar:
Clarify https://github.com/w3c/mediacapture-main/issues/640

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3116,7 +3116,8 @@ interface MediaDeviceInfo {
               <p>A label describing this device (for example "External USB
               Webcam"). This label is for display purposes only. There is no
               requirement the label contain any identifying information about
-              the device.</p>
+              the device. If the device has no associated label, then this
+              attribute MUST return the empty string.</p>
             </dd>
             <dt id="def-mediadeviceinfo-groupId"><dfn>groupId</dfn>
             of type {{DOMString}}, readonly</dt>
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -3114,7 +3114,9 @@ <h2>Attributes</h2>
             <dt><dfn>label</dfn> of type {{DOMString}}, readonly</dt>
             <dd>
               <p>A label describing this device (for example "External USB
-              Webcam"). If the device has no associated label, then this
+              Webcam"). This label is for display purposes only. There is no
+              requirement the label contain any identifying information about
```
jan-ivar, line 3118:
```suggestion
              requirement that the label contain any identifying information about
```

getusermedia.html
```diff
@@ -3114,7 +3114,9 @@ <h2>Attributes</h2>
...
+              the device. If the device has no associated label, then this
```
alvestrand, line 3119: Suggested alternative: "This label is intended to allow the end user to tell the difference between devices. Applications SHOULD NOT assume that the label contains any specific information."

getusermedia.html
```diff
@@ -3114,7 +3114,9 @@ <h2>Attributes</h2>
             <dt><dfn>label</dfn> of type {{DOMString}}, readonly</dt>
             <dd>
               <p>A label describing this device (for example "External USB
-              Webcam"). If the device has no associated label, then this
+              Webcam"). This label is for display purposes only. There is no
+              requirement that the label contain any identifying information about
+              the device. If the device has no associated label, then this
```
jan-ivar, line 3119:
```suggestion
              Webcam"). This label is intended to allow the end user to tell the
              difference between devices. Applications can’t assume that the
              label contains any specific information, such as the device type or model.
```

PR #707, opened 2020-07-23
Identify the list of allowed required constraints for getUserMedia

youennf:
In case an unallowed required constraint is given to getUserMedia, skip it.
This would help https://github.com/w3c/mediacapture-image/issues/229.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3480,7 +3480,7 @@ interface InputDeviceInfo : MediaDeviceInfo {
                         <a>"audio"</a> are simply ignored rather than causing
                         <code>OverconstrainedError</code>.</li>
                         <li><p>If <var>CS</var> contains a constrainable property
-                        that is <a data-lt="required">required</a> and whose name is not in the
+                        that is <a>required</a> and whose name is not in the
                         list of <a>allowed required constraints for device selection</a>,
                         jump to the step labeled <em>Constraint Failure</em> below.</p></li>
                         <li>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3479,9 +3479,10 @@ interface InputDeviceInfo : MediaDeviceInfo {
                         inside of <a>"video"</a> and video-only constraints inside of
                         <a>"audio"</a> are simply ignored rather than causing
                         <code>OverconstrainedError</code>.</li>
-                        <li>Remove any constrainable property inside of
-                        <var>CS</var> that is <a data-lt="required">required</a>
-                        and whose name is not in <a>getUserMedia allowed required constraint list</a>.</li>
+                        <li><p>If <var>CS</var> contains a constrainable property
+                        that is <a data-lt="required">required</a> and whose name is not in the
+                        list of <a>allowed required constraints for device selection</a>,
+                        jump to the step labeled <em>Constraint Failure</em> below.</p></li>
                         <li>
                           <p>Run the <a>SelectSettings</a> algorithm on each
                           track in <var>candidateSet</var> with <var>CS</var>
@@ -3668,7 +3669,7 @@ interface InputDeviceInfo : MediaDeviceInfo {
         no longer suitable. In this case, a NotReadableError will result.</p>
       </div>
       <div>
-        <p>The <dfn>getUserMedia allowed required constraint list</dfn>
+        <p>The <dfn>allowed required constraints for device selection</dfn>
         contains the following constraint names:
           <a>width</a>,
           <a>height</a>,
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3479,10 +3479,11 @@ interface InputDeviceInfo : MediaDeviceInfo {
                         inside of <a>"video"</a> and video-only constraints inside of
                         <a>"audio"</a> are simply ignored rather than causing
                         <code>OverconstrainedError</code>.</li>
-                        <li><p>If <var>CS</var> contains a constrainable property
-                        that is <a>required</a> and whose name is not in the
+                        <li><p>If <var>CS</var> contains a member that is a
+                        <a>required constraint</a> and whose name is not in the
                         list of <a>allowed required constraints for device selection</a>,
-                        jump to the step labeled <em>Constraint Failure</em> below.</p></li>
+                        then <a>reject</a> <var>p</var> with a {{TypeError}}, and abort
+                        these steps.</p></li>
                         <li>
                           <p>Run the <a>SelectSettings</a> algorithm on each
                           track in <var>candidateSet</var> with <var>CS</var>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1802,19 +1802,19 @@ interface MediaStreamTrack : EventTarget {
           </thead>
           <tbody>
             <tr id="def-constraint-sampleRate">
-              <td>sampleRate</td>
+              <td><dfn>sampleRate</dfn></td>
               <td>{{ConstrainULong}}</td>
               <td>The sample rate in samples per second for the audio
               data.</td>
             </tr>
             <tr id="def-constraint-sampleSize">
-              <td>sampleSize</td>
+              <td><dfn>sampleSize</dfn></td>
               <td>{{ConstrainULong}}</td>
               <td>The linear sample size in bits. This constraint can only be
               satisfied for audio devices that produce linear samples.</td>
             </tr>
             <tr id="def-constraint-echoCancellation">
-              <td>echoCancellation</td>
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
@jan-ivar @alvestrand  What are your thoughts?

henbos (approved):

getusermedia.html
```diff
@@ -3462,6 +3462,9 @@ <h2>Methods</h2>
                         inside of <a>"video"</a> and video-only constraints inside of
                         <a>"audio"</a> are simply ignored rather than causing
                         <code>OverconstrainedError</code>.</li>
+                        <li>Remove any constrainable property inside of
+                        <var>CS</var> that is <a data-lt="required">required</a>
+                        and whose name is not in <a>getUserMedia allowed required constraint list</a>.</li>
```
henbos, line 3467: Should we throw OverconstrainedError?

getusermedia.html
```diff
@@ -3635,6 +3638,27 @@ <h2>Methods</h2>
         between those checks, so it is conceivable that the selected device is
         no longer suitable. In this case, a NotReadableError will result.</p>
       </div>
+      <div>
+        <p>The <dfn>getUserMedia allowed required constraint list</dfn>
```
alvestrand, line 3642: Suggest renaming this to "allowed required constraints for device selection", to make it more obvious that applyConstraints() is still allowed to use required constraints for other names.

henbos:
Editors can integrate with comments addressed

getusermedia.html
youennf: Updated accordingly

getusermedia.html
youennf: Updated accordingly

getusermedia.html
```diff
@@ -3479,6 +3479,10 @@ <h2>Methods</h2>
                         inside of <a>"video"</a> and video-only constraints inside of
                         <a>"audio"</a> are simply ignored rather than causing
                         <code>OverconstrainedError</code>.</li>
+                        <li><p>If <var>CS</var> contains a constrainable property
+                        that is <a data-lt="required">required</a> and whose name is not in the
```
dontcallmedom, line 3483: data-lt not needed
```suggestion
                        that is <a>required</a> and whose name is not in the
```

getusermedia.html
```diff
@@ -3479,6 +3479,10 @@ <h2>Methods</h2>
                         inside of <a>"video"</a> and video-only constraints inside of
                         <a>"audio"</a> are simply ignored rather than causing
                         <code>OverconstrainedError</code>.</li>
+                        <li><p>If <var>CS</var> contains a constrainable property
+                        that is <a>required</a> and whose name is not in the
+                        list of <a>allowed required constraints for device selection</a>,
+                        jump to the step labeled <em>Constraint Failure</em> below.</p></li>
```
jan-ivar, line 3485: We have a [precedent](https://github.com/w3c/mediacapture-image/pull/261#issuecomment-698514049) in screen-capture we can copy for navigating these dictionaries.

Also, for consistency, we think in https://github.com/w3c/mediacapture-image/pull/261#issuecomment-698516937 it should throw `TypeError`.

```suggestion
                        <li><p>If <var>CS</var> contains a member whose name is not in the
                        list of <a>allowed required constraints for device selection</a>,
                        and whose value in turn is a dictionary containing a member named
                        <code>min</code>, <code>max</code>, or <code>exact</code>,
                        then <a>reject</a> <var>p</var> with a {{TypeError}}, and abort
                        these steps.</p></li>
```

jan-ivar (changes requested):

getusermedia.html
```diff
@@ -3664,6 +3668,27 @@ <h2>Methods</h2>
...
+          <a>sampleRate</a>,
+          <a>sampleSize</a>,
+          <a>echoCancellation</a>,
+          <a>autoGainControl</a>,
+          <a>noiseSuppression</a>,
+          <a>latency</a>,
+          <a>channelCount</a>,
+          <a>deviceId</a>,
+          <a>groupId</a>.
+        </p>
```
jan-ivar, line 3689: This change flips the default behavior on all other constraints, e.g. all [image capture constraints](https://w3c.github.io/mediacapture-image/#mediatrackconstraintset-section):
```js
  ConstrainDOMString           whiteBalanceMode;
  ConstrainDOMString           exposureMode;
  ConstrainDOMString           focusMode;
  ConstrainPoint2D             pointsOfInterest;

  ConstrainDouble              exposureCompensation;
  ConstrainDouble              exposureTime;
  ConstrainDouble              colorTemperature;
  ConstrainDouble              iso;

  ConstrainDouble              brightness;
  ConstrainDouble              contrast;
  ConstrainDouble              saturation;
  ConstrainDouble              sharpness;

  ConstrainDouble              focusDistance;
  (boolean or ConstrainDouble) pan;
  (boolean or ConstrainDouble) tilt;
  (boolean or ConstrainDouble) zoom;

  ConstrainBoolean             torch;
```
@youennf Was that the intent? Or does imageCapture need a PR as well?

Did we consider having a deny list instead with just pan, tilt and zoom?

youennf: That is the intent. mediacapture-image will have to do dedicated opt-in.
A deny list is not great since we want to move to a world where required constraints disappear from getUserMedia.
For that reason, I would hope we could agree in not adding any additional required constraints to this list.

getusermedia.html
youennf: > We have a [precedent](https://github.com/w3c/mediacapture-image/pull/261#issuecomment-698514049) in screen-capture we can copy for navigating these dictionaries.

What is the benefit? I prefer referring to a concept (required, defined in https://w3c.github.io/mediacapture-main/#dfn-required-constraints) instead of trying to inline the definition of the concept. I would tend to fix media capture-screenshare so that it refers to the concept in mediacapture-main.

Or do you want to disallow these constraints in advanced constraints too?

I am ok with TypeError instead of OverConstrained.

jan-ivar: @youennf All inputs can be validated using the terminology of the incoming data type. No need to involve concepts.

The input we're reading here is a dictionary, which does not contain a "constrainable property" (which is even the wrong concept, it should at least be "constraint") but a [member](https://heycam.github.io/webidl/#dfn-dictionary-member), which has a name (aka key) and a value.

The problem I have using our concepts is that they're complicated and harder to navigate than the WebIDL primitives.

For example, "constrainable property" != "constraint". "[required](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-required)" != "[required constraint](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-required-constraints)".

Specifically, I worry that the sentence _"If CS contains a constrainable property that is **required**"_ is vague about whether to reject on `advanced` constraints which are also [required](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-required) and also something the CS may "contain".

I also worry implementers may miss that the dictionary to inspect is nested one level down.

youennf: > I worry that the sentence _"If CS contains a constrainable property that is **required**"_ is vague about whether to reject on `advanced` constraints which are also [required](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-required) and also something the CS may "contain".

That is why the term "required" is a link that points to the following definition:
"required constraints are all non-advanced constraints that are required."

We could rephrase a bit the sentence though to use the term "required constraint".
This term is already used in getUserMedia just below. I do not think implementors will miss the difference here.

I think this PR is clear about allowing advanced constraints that are required.
We might want to further restrict this but this could be done as a follow-up.

jan-ivar: > That is why the term "required" is a link that points to the following definition:
"required constraints are all non-advanced constraints that are [required](https://w3c.github.io/mediacapture-main/#dfn-required)."

@youennf but that's not where it links. You're using `<a>required</a>` which links to the [required](https://w3c.github.io/mediacapture-main/#dfn-required) definition which doesn't preclude `advanced` constraints. This is the kind of confusion I'm talking about.

 Also the terminology is wrong. How about:
```suggestion
                        <li><p>If <var>CS</var> contains a member that is a
                        <a>required constraint</a> and whose name is not in the
                        list of <a>allowed required constraints for device selection</a>,
                        then <a>reject</a> <var>p</var> with a {{TypeError}}, and abort
                        these steps.</p></li>
```

youennf: I see, thxs for the update.

youennf:
Are we good margin the PR?

jan-ivar (approved):

Issue #703, opened 2020-07-02
Spec should provide algorithms to expose individual devices via enumerateDevices


youennf:
As part of media capture-output and getUserMedia in chrome picker, there is an effort to only expose a device after device has been picked by user. Algorithms should be added to allow this.

guest271314:
Some recent notes 

To resolve a series of issues the formal algorithm should be 

1. `enumerateDevices()` => lists all devices => before current `getUserMedia()` algorithm
2. For backward compatibility `getUserMedia()` implicitly calls `enumerateDevices()` => lists all devices 
3. User can select any and all devices for that permission session, which avoids needing to execute `getUserMedia()` more than once
4. All of the captured devices each have single `MediaStreamTrack` representation of capture, necessarily meaning there can be more than 1 microphone and 1 camera in `MediaStream` from `getUserMedia()`; it is up to user to process the tracks thereafter; there should not be an asumption that only 1 microphone and 1 camera are being captured for the duration of permissions, with speakers, cameras and other accessible media capable of being added as a device during permission being granted
5. However, to reign in confusion as to which devices are being captured the list of captured devices needs to be dynamic; and to do that with existing `enumerateDevices()` would mean the function must perform at least two different task; replace `getUserMedia()` as primary permission and device selection mechanism and thereafter list captured devices. 
6. Avoid overloading algorithms designed for a different purpose
&. Describe a `getSystemDevices()` function used to first get permission to list all devices, not just microphone and camera; user selects devices for that permission session, thereafter `enumerateDevices()` lists only those initially selected devices for that session; if plug-and-play is expected to work `ondevicechange` needs to work by user checking to watch for plug-and-play that the user explicitly grants permission to do so for meaning `enumerateDevices()` will be consistent and reliable due to being dynamic, otherwise only return selected devices
7. The algorithm needs to be basic enough to be implemented universsally with the same code, to avoid retuning back to this issue due to Media Capture and Streams not mandating a uniform UI, to that end a simple form that can be compised of plain text to either check or not check spekers, monitor of device, cameras, microphones, MIDI, et al., e.g., (https://github.com/w3c/mediacapture-main/issues/640#issuecomment-653108135) to avoid workarounds and inconsistent behaviour moving forward it must be acknowledged that `getUserMedia()` is capable of and is being used for other or more than 1 micorphone and 1 camera, that is, an extensible API design-pattern from the beginning instead of attempting to amend the functionality of prior art
[...]

PR #718, opened 2020-08-27
Extend enumerateDevices algorithm to handle non camera and microphone devices

youennf:
This introduces a hook so that other specifications can define the exact exposure algorithm.


youennf:
Fixes #703

guest271314:
Does this effectively mean that `getUserMedia()` is no longer restricted to capturing only microphone input? So that per the specification `enumerateDevices()` should now expose Monitor devices at Chrome, Chromium?

PR #721, opened 2020-09-08
Pass camera and microphone lists to the exposure decision algorithm

youennf:
Loop through discovered devices first to build a microphone list.
Loop through discovered devices second to build a camera list.
Append microphone list then camera list to resultList.
Then use exposure decision algorithm for other devices with microphone list and camera list as input.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2942,7 +2942,7 @@ interface MediaDevices : EventTarget {
                       </ol>
                     </li>
                     <li>
-                      <p>Else, run the following steps:<p>
+                      <p>Otherwise, run the following steps:<p>
                       <ol>
                         <li>
                           <p>Append to <var>resultList</var> all devices of <var>microphoneList</var> in order.</p>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2885,41 +2885,92 @@ interface MediaDevices : EventTarget {
                       </ol>
                     </li>
                     <li>
-                      <p>Probe the User Agent for available media devices, and
-                      run the following sub steps for each discovered device,
-                      <var>device</var>:</p>
+                      <p>Probe the User Agent for available media devices, and let be <var>deviceList</var>
+                      be the list of all discovered devices.</p>
+                    </li>
+                    <li>
+                      <p>Let <var>microphoneList</var> and <var>cameraList</var> be empty lists.</p>
+                    </li>
+                    <li>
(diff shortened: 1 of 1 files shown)


alvestrand:
This PR imposes a requirement that all cameras come first, then all microphones.
It will make implementations that return cameras and microphones interleaved non-conformant, and will move the default microphone from "one of the two first entries" to "the position after all the cameras".
Why?

(I am very sensitive to changing specs to require things that are not currently required where existing implementations have been written without knowledge of these new requirements.)

youennf:
> This PR imposes a requirement that all cameras come first, then all microphones.

This is the reverse actually that this PR intends: microphone first, then camera, then other devices.
This matches what Chrome and Safari do. Firefox also sorts device per kind but put camera first.

> It will make implementations that return cameras and microphones interleaved non-conformant,

Correct, though browsers I tested are not doing that.
There is a compat risk for user agents to interleave devices.

> will move the default microphone from "one of the two first entries" to "the position after all the cameras".

PR intent is to put default microphone as the first entry (but default camera further away).
No browser I tested implements the current wording of the spec.

> Why?

In general, I think specifying order will make developer's life easier.

We can probably mandate per-device-kind sorting as this seems already supported by implementations.
We need to do something about the default device since no browser implements the spec.

I also hope implementations can agree to a particular kind order (microphone first, camera second).
Firefox point of view is especially useful here since this PR requires a change on their end.

youennf:
@alvestrand, for context, I updated https://github.com/w3c/mediacapture-output/pull/109 to use microphoneList based on this PR.

henbos:
Looks good to me

youennf:
@jan-ivar, can you look at the PR and see whether Firefox is fine aligning with it.

getusermedia.html
```diff
@@ -2885,41 +2885,92 @@ <h2>Methods</h2>
...
+                        <li>
+                          <p>Add to <var>resultList</var> the first item of <var>microphoneList</var> if <var>microphoneList</var> is not empty.</p>
+                        </li>
+                        <li>
+                          <p>Add to <var>resultList</var> the first item of <var>cameraList</var> if <var>cameraList</var> is not empty.</p>
+                        </li>
+                      </ol>
+                    </li>
+                    <li>
+                      <p>Else, run the following steps:<p>
```
jan-ivar, line 2945:
```suggestion
                      <p>Otherwise, run the following steps:<p>
```

PR #722, opened 2020-09-09
Define the fitness distance for all required constrains.

eehakkin:
This CL defines the fitness distance also for required constrains whose
value contains a member named 'ideal' in addition to members named
'exact', 'max' and/or 'min'.

Lets take a **width** constraint `{ideal: 1920, min: 320}` as an example and lets assume that the browser supports the constraint and that there are settings dictionaries which satisfy the constraint (so fitness distance algorithm steps 1 and 2 can be ignores).
Then,
* the step 3 does not apply because the constraint is required (has 'min'),
* the step 4 does not apply because the ideal value is specified, and
* the steps 5 and 6 do not apply because the constraint is not non-required but required,

thus the fitness distance is undefined.

Clearly, the step 5 should apply and the fitness distance should be positive infinity for settings dictionaries with width less than 320, 0 for settings dictionaries with width of 1920 and  |actual - 1920| / max(|actual|, 1920) for other settings dictionaries.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -4271,15 +4271,13 @@ interface ConstrainablePattern {
                 either contains no member named 'ideal', or, if bare values are
                 to be treated as 'ideal', isn't a bare value), the fitness
                 distance is 0.</li>
-                <li>For all positive numeric
-                <a data-lt="required">non-required</a> constraints (such as
+                <li>For all positive numeric constraints (such as
                 height, width, frameRate, aspectRatio, sampleRate and
                 sampleSize), the fitness distance is the result of the formula
                   <pre>
 (actual == ideal) ? 0 : |actual - ideal| / max(|actual|, |ideal|)</pre>
                 </li>
-                <li>For all string and enum
-                <a data-lt="required">non-required</a> constraints (e.g.
+                <li>For all string, enum and boolean constraints (e.g.
                 deviceId, groupId, facingMode, resizeMode, echoCancellation),
                 the fitness distance is the result of the formula
                   <pre>(actual == ideal) ? 0 : 1</pre>


youennf:
Booleans are also not covered.
It might be good to add a catch-them-all case where the distance would be 0.

henbos:
@guidou Can you add your two cents?

guidou:
I think this change is correct (wasn't the spec written this way in the past?), but booleans need to be added to step 6.

getusermedia.html
```diff
@@ -4271,15 +4271,13 @@ <h2>Methods</h2>
                 either contains no member named 'ideal', or, if bare values are
                 to be treated as 'ideal', isn't a bare value), the fitness
                 distance is 0.</li>
-                <li>For all positive numeric
-                <a data-lt="required">non-required</a> constraints (such as
+                <li>For all positive numeric constraints (such as
```
alvestrand, line 4274: This needs to take account of the case where ideal is not present.
I suggest "For all positive numeric constraints where there is an ideal value present (such as..."

Otherwise, the formula calculation aborts because "ideal" is an undefined value.

eehakkin: > This needs to take account of the case where ideal is not present.

Actually, ideal is guaranteed to be present during these steps 5 and 6 because the previous steps states that if no ideal value is specified, the fitness distance is 0.

guidou: Step 4 already says that fitness is zero if there is no ideal value.

eehakkin:
> I think this change is correct (wasn't the spec written this way in the past?), but booleans need to be added to step 6.

Done.

youennf:
I am pretty sure this is inline with what all implementations are doing but it is worth checking.
Can we write a WPT test checking exactly that something like getUserMedia({ video : { min : 160, ideal : 320 }) actually returns a video track of width 320 (and not 160 or the default 640 that some implementations may use)?

eehakkin:
> I am pretty sure this is inline with what all implementations are doing but it is worth checking.
> Can we write a WPT test checking exactly that something like getUserMedia({ video : { min : 160, ideal : 320 }) actually returns a video track of width 320 (and not 160 or the default 640 that some implementations may use)?

Done in https://github.com/web-platform-tests/wpt/pull/25493

henbos (approved):

Issue #724, opened 2020-09-10
Provide guidelines to persist capture transition indicators for long enough


youennf:
The spec says:
Any false-to-true transition indicated MUST remain observable for a sufficient time that a reasonably-observant user could become aware of it.

Current implementations seem to limit the observability of the transition if getUserMedia is called and the track got stopped shorter after that
The following fiddle https://jsfiddle.net/6axho8Lj/ shows the current behavior. The faster the camera start up time, the shorter the capture indicator. 

Should we provide more guidelines about what is 'sufficient time'. Mention the case of call getUserMedia then stop the tracks?
3 seconds is already used in the spec: 'the UA SHOULD relinquish the device within 3 seconds'.
Is 3 seconds a good suggestion?

PR #734, opened 2020-10-06
Privacy indicator false-to-true transition SHOULD be 3 seconds.

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/724.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/mediacapture-main/pull/734.html" title="Last updated on Nov 12, 2020, 3:23 PM UTC (8476eca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-main/734/ea45c6a...jan-ivar:8476eca.html" title="Last updated on Nov 12, 2020, 3:23 PM UTC (8476eca)">Diff</a>

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -5136,7 +5136,7 @@ window.onload = async () =&gt; {
       indicate when the value changes.</li>
       <li>Any false-to-true transition indicated MUST remain observable for a
       sufficient time that a reasonably-observant user could become aware of
-      it.</li>
+      it (this SHOULD be 3 seconds).</li>
       <li>Any of the above transition indications MAY be combined as long as
       the combined indication cannot transition to false if any of its
       component indications are still true.</li>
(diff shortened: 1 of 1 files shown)


alvestrand:
Having the "false-to-true transition remain observable" is kind of a strange way to put it. Should we instead "the inidcator MUST remain in the true position for at least 3 seconds"?

jan-ivar:
> I take it referring to "the transition" rather than "the state" is to permit both implementation by indicator light (the typical webcam case) and by transient popups.

Yes, e.g. toast messages on phones,  but also for normative clarity, since it is where the transition happens that we want implementers to put code. E.g. a polling implementation of the state would be non-compliant, as it may miss the state changing quickly.

Language predates this PR.

henbos (approved):

Issue #735, opened 2020-10-06
Make fitness distance a SHOULD instead of MAY on device delection.


jan-ivar:
I think we should strengthen this to ensure web compat around constraints use in `getUserMedia`.

Sites are going to do things like:
```js
await navigator.mediaDevices.getUserMedia({video: {deviceId: deviceFromLastVisitId, height: 1080}});
```
...and expect comparable results across browsers (same device if available, or comparable 1080p cam if not).

I suspect implementations already comply, but right now this hinges on a **MAY** in the spec which seems too weak.

On the **MAY**: Firefox is the lone implementation to allow users to override the (ideal) device choice in its permission prompt, but I think that's a valid exception (if the user explicitly chose a device, the site should probably honor it!)

But that shouldn't stand in the way of improving web compat of the automated device selection process where the user does _not_ override.

So I'm proposing a **SHOULD** instead of **MAY** here, with a mention of _"user preference"_ as an example of a valid exception.

guest271314:
> Sites are going to do things like:
> 
> ```js
> await navigator.mediaDevices.getUserMedia({video: {deviceId: deviceFromLastVisitId, height: 1080}});
> ```
> 
> ...and expect comparable results across browsers (same device if available, or comparable 1080p cam if not).

Well, that expectation would be misplaced. Currently there is zero consistency with regard to browsers and video device selection: Firefox and Chromium proffer different devices as default https://github.com/w3c/mediacapture-main/issues/727. 

Chromium does not support capture of monitor devices and uses the label `"audiooutput"` when in fact that is the exact same device as the microphone, and neither this specification nor Audio Output Devices API actually define capture of headphones or speakers. 

"SHOULD" is still not strong enough for these billigerent implementers. Am skeptical that "MUST" is even strong enough language, yet with "MUST" there is no ambiguity with regard to intent of the authors.

Issue #720, opened 2020-09-07
Clarify "audiooutput" does not mean capture of audio output to headphones or speakers


guest271314:
Per Issue 1114422: enumerateDevices() listing device kind "audioouput" is incorrect and misleading https://bugs.chromium.org/p/chromium/issues/detail?id=1114422

> "audiooutput" refers to audio playback via a media element. It does not refer to microphone input or audio capture of anything. 

This specification states https://w3c.github.io/mediacapture-main/#idl-def-MediaDeviceKind.audiooutput

> **audiooutput** | Represents an audio output device; for example a pair of headphones.

Audio Output Devices API https://w3c.github.io/mediacapture-output/ does not actually use or define the term `"audiooutput"`. 

There is a section of Audio Output Devices API that addresses `getUserMedia()` relations to that specification

> [4.2 Obtaining Consent](https://w3c.github.io/mediacapture-output/#privacy-obtaining-consent)
> The user agent may explicitly obtain user consent to play audio out of non-default output devices using [`selectAudioOutput`](https://w3c.github.io/mediacapture-output/#dom-mediadevices-selectaudiooutput).
> 
> Implementations MUST also support implicit consent via the [`getUserMedia()`](https://www.w3.org/TR/mediacapture-streams/#dom-mediadevices-getusermedia) permission prompt; when an audio input device is permitted and opened via [`getUserMedia()`](https://www.w3.org/TR/mediacapture-streams/#dom-mediadevices-getusermedia), this also permits access to any associated audio output devices (i.e., those with the same groupId). This conveniently handles the common case of wanting to route both input and output audio through a headset or speakerphone device.

the remainder of comment from the above-linked Chromium bug 

> selectAudioOutput() is to be used in combination with setSinkId() and it has nothing to do with audio capture or microphone inputs; only playback (on media elements). It's a way to get access to a deviceId of kind "audiooutput" without using enumerateDevices. This deviceId is useful only with setSinkId since it's of kind "audiooutput".
> 
> Wrt the getUserMedia() UI in Chromium, its purpose is to request permission and not to select device. The device is selected based on the constraints passed to getUserMedia().  This might change in the future, following some spec changes, but for now it is totally spec compliant. Even if the prompt allowed for device selection, it wouldn't support capturing from monitor devices since those devices are not supported by Chromium. It would only list the devices reported as audioinput by enumerateDevices.

**Problems**

[...]

PR #736, opened 2020-10-06
Make fitness distance a SHOULD instead of MAY on device selection, …

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/735.

…with allowance for user preference.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3623,11 +3623,11 @@ interface InputDeviceInfo : MediaDeviceInfo {
                           the User Agent and may be determined by asking the user.
                           Once selected, the source of the
                           {{MediaStreamTrack}} MUST NOT change.</p>
-                          <p>The User Agent MAY use the value of the computed
-                          "fitness distance" from the <a>SelectSettings</a>
-                          algorithm, or any other internally-available information
-                          about the devices, as an input to the selection
-                          algorithm.</p>
+                          <p>The User Agent SHOULD use the value of the computed
+                          <a>fitness distance</a> from the <a>SelectSettings</a>
+                          algorithm as an input to the selection algorithm.
+                          However, it MAY also use other internally-available
+                          information about the devices, such as user preference.</p>
                           <p>User Agents are encouraged to default to using the
                           user's primary or system default device for <var>kind</var>
                           (when possible). User Agents


guest271314:
> Once selected, the source of the {{MediaStreamTrack}} MUST NOT change.</p>	                          {{MediaStreamTrack}} MUST NOT change.

is not really possible to enforce from the specification, and should not be specified anyway, as that takes away any control the user has over determining which devices at their own system they capture. To overcome Chromium refusal to list and capture monitor devices at Linux it is possible to utilize `pavucontrol` to dynamically switch devices during capture from either Chromium input or Nightly audio callback driver input. That is a feature that _should_ be specified here, rather than attempt to restricted as must not happen.

guest271314:
Technically the stream is moved between devices by 

> `pactl move-source-output` is the command to move capture streams ("source output" means capture stream and "sink input" means playback stream).

https://gitlab.freedesktop.org/pulseaudio/pavucontrol/-/issues/91#note_590795.

What it appears you are trying to do will first require creating some form of initial consistency between browsers, so that they select the same default devices by default, which does not occur now. When the issue of compatibility and interoperability arise, that is where the rubber meets the road, as them old folks say. 

For example, Firefox does correctly list monitor devices, as those devices are inputs. Chromium does not. To work around that one option is to stream monitor devices from Firefox to Chromium using WebRTC, among other workarounds, when we should just be able to perform the same or similar tasks at each browser, without undue restrictions.

Now, have filed multiple issues at this repository to include language that officially supports capture of monitor devices at Linux, or, if preferred, system audio output. The resounding response has been, essentially, "No". The is fine, as am well suiting to achieving requirements not officially supported by specification or implementation.

The ambiguity becomes apparent when stating, o.k., you do not support capture of speakers or headphones, why is `"audiooutput"` in this specification at all? Which leads to confusion for those users in the field that take the definition literally and happen to be actually trying to capture audio output to speakers and headphones. Then the problem is attempted to be deferred to Audio Output Devices API, where the term of art `"audiooutput"` is still not applicable, as that specification does not define capture of speakers or headphones either. Thus, if you really do not want or intend to capture headphones or speakers, remove that suggestive language from _your_ specification, _and_ tell those users that you are concerned about in the filed that `"audiooutput"` is a misnomer that does not intend to convey that capture of speakers or headphones is possible. However, as yet, none of the authors of that language have notified users in the field at large https://stackoverflow.com/a/45003549 that their efforts to capture headphones and speakers are for naught, https://github.com/w3c/mediacapture-main/issues/720, or at least include a conspicuous note stating the same, as it is reasonable to take the current definition of `"audiooutput"` literally, and as Chromium is the only implementation that defines that label, users will try at tht browser, to no avail.



In summary, there are many inconsistencies between implementations, and several outright ambiguities and language that when tested in the field does not produce expected results. Once you begin this procecure of trying to align consistency, if the motivation of this PR is really "Sites" then the users of those sites should be heard, clearly.

jan-ivar:
TPAC consensus was to merge this PR and deal with any PTZ exceptions in image capture.

From [minutes](https://www.w3.org/2020/10/20-webrtc-minutes.html): _"PR looks fine, but there are cases with PTZ where this may not work ... the goal is to clarify our intent that this is what ought to be done"_

henbos (approved):

PR #742, opened 2020-10-30
Define the fitness distance for unexposed constraints

eehakkin:
This is to clarify meaning of presence of certain constraints (especially PTZ constraints) when devices do not support them, as discussed in TPAC meeting.

This also partially addresses https://github.com/w3c/mediacapture-image/issues/256.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -4351,6 +4351,14 @@ interface ConstrainablePattern {
                 either contains no member named 'ideal', or, if bare values are
                 to be treated as 'ideal', isn't a bare value), the fitness
                 distance is 0.</li>
+                <li>
+                  <p>If <var>constraintName</var> is not in the list of
+                  <a>allowed required constraints for device selection</a> and
+                  the <a>settings dictionary</a>'s value for the constraint
+                  does not exist (because the source does not support the
+                  constraint or because the support is not exposed), the fitness
+                  distance is 1.</p>
+                </li>
                 <li>For all positive numeric constraints (such as
                 height, width, frameRate, aspectRatio, sampleRate and
                 sampleSize), the fitness distance is the result of the formula


guidou (approved):
I think we need additional language to cover the special true/{} case for PTS, but I guess that should be covered in the image capture spec.

getusermedia.html
```diff
@@ -4351,6 +4351,14 @@ <h2>Methods</h2>
                 either contains no member named 'ideal', or, if bare values are
                 to be treated as 'ideal', isn't a bare value), the fitness
                 distance is 0.</li>
+                <li>
+                  <p>If <var>constraintName</var> is not in the list of
+                  <a>allowed required constraints for device selection</a> and
+                  the <a>settings dictionary</a>'s value for the constraint
+                  does not exist (because the source does not support the
+                  constraint or because the support is not exposed), the fitness
+                  distance is 1.</p>
```
jan-ivar, line 4360: I think this is the right place, but I'm not sure those are the ingredients. I think we need something like this:

_"If the constraint is [not required](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-required), applies for this type of device but not this particular device, and the constrainable property is not on the [list of inherent constrainable track properties](https://w3c.github.io/mediacapture-main/getusermedia.html#list-of-inherent-constrainable-track-properties), then the fitness distance is 1 (that is, the presence of a constraint does not influence the fitness distance)."_

jan-ivar (changes requested):

getusermedia.html
eehakkin: That might be in the right direction, but I  think that it is not enough.

I think that the condition **_the constraint is [not required](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-required)_** is extraneous. That leaves the fitness distance undefined when a constrain is required, the settings dictionary somehow satisfies the constraint and an ideal value is specified but does not apply for this particular device.
Something like `{foo: {exact: false, ideal: "bar"}}` and a settings dictionary for a foo incapable device. In that case the fitness distance should probably be 1 so this step should define the fitness distance. Also, there is no _actual_ foo value to compare against so the remaining steps cannot define the fitness distance.

The condition **_[the constraint] applies for this type of device_** is probably not strictly needed (because the  [fitness distance](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-fitness-distance) step 3 already defines the fitness distance to be 0 if the constraint does not apply for this type of device) but it does no harm.

I think that the condition **_[the constraint does] not [apply for] this particular device_** is problematic. If a constraint set is `{zoom: {ideal: 2}}` and the device is zoom capable (thus the constraint applies for this particular device) but the UA does not expose the setting (due to permissions for instance and thus the settings dictionary does not have a setting for the zoom), shouldn't the fitness distance still be 1? Otherwise the application could select devices based on unexposed constrainable properties.

jan-ivar: > I think that the condition **_the constraint is [not required](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-required)_** is extraneous. That leaves the fitness distance undefined when a constrain is required,

By definition, satisfied required constraints don't have fitness distance. They're delimiters (`min` or `max`) or hard requirements (`exact`), whose sole purpose is candidate elimination.

> the settings dictionary somehow satisfies the constraint and an ideal value is specified but does not apply for this particular device.
> Something like `{foo: {exact: false, ideal: "bar"}}` and a settings dictionary for a foo incapable device. In that case the fitness distance should probably be 1

I think it needs to be `Infinity`, by the same logic that they're unfit, from https://github.com/w3c/mediacapture-image/issues/256

E.g. had we still supported required constraints for `pan`, we'd agree non-ptz cameras shouldn't satisfy
```js
await navigator.mediaDevices.getUserMedia({video: {pan: {exact: 0}}}); // OverconstrainedError
```
Now `pan` is not on the list of [allowed required constraints for device selection](https://w3c.github.io/mediacapture-main/#dfn-allowed-required-constraints-for-device-selection) so it throws `TypeError` before we get this far. But for other future constraints, I think `Infinity` would be correct.

> The condition **_[the constraint] applies for this type of device_** is probably not strictly needed (because the [fitness distance](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-fitness-distance) step 3 already defines the fitness distance to be 0 if the constraint does not apply for this type of device) but it does no harm.

Yeah unfortunately, the reliance on order here is not terribly explicit, so being overly specific might be necessary.

> I think that the condition **_[the constraint does] not [apply for] this particular device_** is problematic. If a constraint set is `{zoom: {ideal: 2}}` and the device is zoom capable (thus the constraint applies for this particular device) but the UA does not expose the setting (due to permissions for instance and thus the settings dictionary does not have a setting for the zoom), shouldn't the fitness distance still be 1? Otherwise the application could select devices based on unexposed constrainable properties.

This seems like an existing problem imageCapture needs to handle. How does it avoid device selection by ptz today?

jan-ivar: I also propose the following further up:
```
               <p>We use the term <dfn data-export>settings dictionary</dfn> for the set of
-              values that might be applied as settings to the object.</p>
+              values that might be applied as settings to the object. Do not
+              include inherent properties as members unless they are in the
+              <a>list of inherent constrainable track properties</a> (to avoid
+              a broken clock being right twice a day).</p>
```

jan-ivar:
I've created https://github.com/w3c/mediacapture-main/pull/766 as an alternative to this PR, since github wouldn't allow me to suggest changes outside lines already touched. PTAL

jan-ivar:
Closing in favor of https://github.com/w3c/mediacapture-main/pull/766

Issue #842, opened 2021-12-17
devicechange should fire when enumerateDevices works


jan-ivar:
[enumerateDevices()](https://w3c.github.io/mediacapture-main/getusermedia.html#dom-mediadevices-enumeratedevices) says the _"User Agent MAY wait to proceed to the next step until document is fully active and has focus"_, but [devicechange](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-device-change-notification-steps) doesn't.

The point of the event is to alert JS to call enumerateDevices, so firing it when the method will just block seems pointless.

Moving the _MAY_ into the [device enumeration can proceed](https://w3c.github.io/mediacapture-main/getusermedia.html#device-enumeration-can-proceed) algorithm should fix this.

tonyherre:
As currently written, the devicechange event steps MUST only happen when device enumeration can proceed, so the current behaviour is that events are only fired for active & focussed pages, but moving the "MAY wait" into "device enumeration can proceed" would mean events are now fired on focus.
Sounds like a better way to me, but is this what you intended?

alvestrand:
The interesting version of devicechange with backgrounded tabs is if someone unplugs his headset while the tab is backgrounded, and expects sound to automagically switch to speakers without having to foreground the tab first.

jan-ivar:
> As currently written, the devicechange event steps MUST only happen when [device enumeration can proceed](https://w3c.github.io/mediacapture-main/getusermedia.html#device-enumeration-can-proceed), so the current behaviour is that events are only fired for active & focussed pages

@tonyherre Not quite. That algorithm returns true if [device information can be exposed](https://w3c.github.io/mediacapture-main/getusermedia.html#device-information-can-be-exposed), which is effectively [[[canExposeCameraInfo]]](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-canexposecamerainfo) || [[[canExposeMicrophoneInfo]]](https://w3c.github.io/mediacapture-main/getusermedia.html#dfn-canexposemicrophoneinfo), which get set to true on gUM success.

I.e. the spec mandates firing the event on background tabs that have called gUM successfully at least once since opened.

As mentioned in the OP, it's only enumerateDevices that has an additional _"MAY ... wait ... until ... focus"_. We'd like to fix this.

> ... moving the "MAY wait" into "device enumeration can proceed" would mean events are now fired on focus.

[...]

PR #858, opened 2022-01-26
Move enumerateDevices' 'MAY wait for focus' to affect devicechange as well

jan-ivar:
Fixes https://github.com/w3c/mediacapture-main/issues/842.


karlt:
LGTM thanks.

PR #864, opened 2022-03-03
Specify sub-type requirement for cloning tracks

### LINKED ISSUE #857: It is unclear from the spec whether cloning a MediaStreamTrack preserves the subtype or not of the cloned track
It is unclear from the spec whether cloning a MediaStreamTrack preserves the subtype or not of the cloned track.
This can be tested for canvas tracks and cloning does preserve the subtype.
It might be good to add a note stating this intent in this spec.

eladalon1983:
Fixed #857.

***
<a href="https://pr-preview.s3.amazonaws.com/eladalon1983/mediacapture-main/pull/864.html" title="Last updated on Mar 3, 2022, 12:02 PM UTC (4f7a02c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-main/864/eb26b96...eladalon1983:4f7a02c.html" title="Last updated on Mar 3, 2022, 12:02 PM UTC (4f7a02c)">Diff</a>


getusermedia.html
```diff
@@ -697,9 +697,12 @@ <h2>{{MediaStreamTrack}}</h2>
...
         </li>
         <li>
-          <p>Let <var>trackClone</var> be the result of
-          [=create a MediaStreamTrack | creating a MediaStreamTrack=] with
-          <var>source</var> and the value <code>false</code>.
+          <p>
+            Let <var>trackClone</var> be the result of
+            [=create a MediaStreamTrack | creating a MediaStreamTrack=] with
+            <var>source</var> and the value <code>false</code>.
+            <var>trackClone</var> MUST be of the same sub-type of
```
alvestrand, line 704: Suggest phrasing as "If <var>track</var> is a subclass of MediaStreamTrack, and the subclass does not specify a different rule for "clone", <var>trackClone</var> MUST be an object of the same subclass.

youennf:
I think we should change the create a MediaStreamTrack algorithm instead.
In particular, we could have the source has a MediaStreamTrackType slot.
If the slot is there, we would create a track of type MediaStreamTrackType, defaulting to MediaStreamTrack.
We might need to add some extension points for being able to initialize additional slots.
I would hope we can do that inside create a MediaStramTrack algorithm.
This would solve the transfer case as well. I can have a stab at what I have in mind if I am not clear here.

eladalon1983:
> I can have a stab at what I have in mind if I am not clear here.

Be my guest. :-)

Issue #887, opened 2022-03-22
Spec-hooks requested


eladalon1983:

youennf:
Proposal is to add a hook for specs to check on a per document basis whether a document has any live and unmuted capture track. This includes camera, microphone and screen share.

youennf:
@kenchris, do you want one hook for all capture tracks or three hooks (one for microphone, one for camera and one for screensharing)?

kenchris:
I am fine with one, thanks

PR #890, opened 2022-06-20
Add an algorithm to determine whether a document is capturing or not.

youennf:
A document is said to be capturing if one of its track source is a capture source and is live.
A capture source is created either by getUserMedia or getDisplayMedia.

Fixes https://github.com/w3c/mediacapture-main/issues/887

***
<a href="https://pr-preview.s3.amazonaws.com/youennf/mediacapture-main/pull/890.html" title="Last updated on Aug 18, 2022, 2:50 PM UTC (e86e63f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-main/890/7b093a3...youennf:e86e63f.html" title="Last updated on Aug 18, 2022, 2:50 PM UTC (e86e63f)">Diff</a>

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3151,6 +3151,32 @@ interface MediaDevices : EventTarget {
         <p>By default, it returns <code>false</code>.</p>
         <p>Other specifications can define the algorithm for specific device types.
       </section>
+      <section>
+        <h2>Context capturing state</h2>
+        <p>To perform a <dfn data-export id="context-is-capturing">context is capturing</dfn> check
+        for <var>context</var>, run the following steps:</p>
+        <ol>
+          <li>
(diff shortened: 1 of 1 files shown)


youennf:
The current algorithm is implemented so that the capturing state will return true for documents for which there is a privacy indicator. This, in particular, has an impact in case of transferred track.

alvestrand:
This seems OK to me - but I'd like to understand where the linkage will be - this will return true both for camera and screen capture, I think.

youennf:
> this will return true both for camera and screen capture, I think.

Right, or microphone.

alvestrand:
Editors meeting: need to have answers to questions, but otherwise seems uncontroversial.

youennf:
This will be used by https://github.com/WICG/compute-pressure/issues/82.

kenchris:
Awesome, thanks all

Issue #898, opened 2022-08-11
Should enumeration of microphones and cameras depend on device-kind-specific exposure checks?


karlt:
https://github.com/w3c/mediacapture-main/pull/773 split "device information can be exposed" into "camera information can be exposed" and "microphone information can be exposed" and applied the new checks before initializing device-specific information when [creating a device info object](https://w3c.github.io/mediacapture-main/#creating-a-device-info-object).

`enumerateDevices()` still [uses](https://w3c.github.io/mediacapture-main/#ref-for-device-information-can-be-exposed-1) the weaker combined "device information can be exposed" check to determine whether or not to truncate both `microphoneList` and `cameraList` to their first items.

The effect is that, when only camera (not microphone) information can be exposed, for example, the set of device info objects produced by `enumerateDevices()` includes a list of (potentially several) identical mostly uninitialized audioinput device info objects, one for each microphone.

Such a list of unitialized device info objects contains information about the number of devices of that kind, but exposure of this information based on permission for another device kind is unexpected (to me at least).

Proposal: Use "camera information can be exposed" and "microphone information can be exposed" checks separately to determine whether to truncate each of `microphoneList` and `cameraList`.

[Consensus](https://github.com/w3c/mediacapture-main/issues/645#issuecomment-724199486) was to ["limit enumerateDevices to the type of devices being shared" "we will align with what Chromium implements which is more privacy sensitive"](https://www.w3.org/2020/10/20-webrtc-minutes.html#t04).

[Experiments](https://jan-ivar.github.io/dummy/enumerate.html) with Chrome 103 indicate that Chrome is using separate camera and microphone permissions for decisions of truncation of `microphoneList` and `cameraList`.

jan-ivar:
@youennf can you speak to this?

youennf:
> Such a list of unitialized device info objects contains information about the number of devices of that kind, but exposure of this information based on permission for another device kind is unexpected (to me at least).

Right, this was unintended, PR should fix it as you proposed.
Thanks!

Issue #646, opened 2019-11-28
Should enumerateDevices by default return an empty list?


youennf:
That would prevent all enumerateDevices fingerprinting for pages without camera/microphone permission.

alvestrand:
My personal thought:
returning an empty list is an active false statement ("I have no devices") for all code that depends on the current behavior, and is likely to trigger "you can't use this feature until you plug in..." UI.
If we want to break stuff that depends on listing devices without having one open, we should return an error / rejected promise.

youennf:
enumerateDevices already returns an empty list for third-party iframes that do not have permission to access cameras/microphones. This is not false, it just means that the page is not granted access to more information.

Rejecting might have bigger backward compatibility risk since navigator.mediaDevices.enumerateDevices() never rejects currently, this might be a good experiment but I don't think we should mandate that.

guest271314:
As noted by @pes10k at https://github.com/w3c/mediacapture-main/issues/690#issuecomment-623555482

> Sorry, i just don't understand what you're asking, other than to say that the values exposed by enumerateDevices are used as inputs to generating fingerprint-derived identifiers, and for some users will be highly identifying. Hope that helps

One solution would be to not expose devices to websites at all. Devices can be exposed to users. The user can then select devices that the website cannot "see" in the form they are now, and grant website permission to generic "audio" or "video" device.

A rough concept can take several forms. Consider a `Blob URL`, where, in general, the lifetime of the `Blob URL` is the lifetime of the `document` which created it. When the user executes `getUserMedia()` a prompt is displayed save for the case of enabling fake media streams using flags or preferences. 

When `enumerateDevices()` is executed the result can be displayed, or accessed only by the user, not the website. The user must always have access to their devices as  a list, unless they explicitly set otherwise, which should also be an option. Similar to current `getUserMedia()` and `getDisplayMedia()` prompt. 

[...]

PR #900, opened 2022-08-25
Update enumerateDevices algorithm to make use of device-kind specific exposure checks when building cameraList and microphoneList.

youennf:
This ensures we do not get dummy similar audioinput or camera devices if only camera or only microphone is used.

Fixes https://github.com/w3c/mediacapture-main/issues/898

***
<a href="https://pr-preview.s3.amazonaws.com/youennf/mediacapture-main/pull/900.html" title="Last updated on Oct 6, 2022, 2:30 PM UTC (a8699aa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-main/900/e84ad3e...youennf:a8699aa.html" title="Last updated on Oct 6, 2022, 2:30 PM UTC (a8699aa)">Diff</a>

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -2933,16 +2933,12 @@ interface MediaDevices : EventTarget {
                   </ol>
                 </li>
                 <li>
-                  <p>If [=device information can be exposed=] is <code>false</code>,
-                  run the following sub steps:<p>
-                  <ol>
-                    <li>
-                      <p>If <var>microphoneList</var> is not empty, truncate <var>microphoneList</var> to its first item.</p>
-                    </li>
-                    <li>
-                      <p>If <var>cameraList</var> is not empty, truncate <var>cameraList</var> to its first item.</p>
-                    </li>
-                  </ol>
+                  <p>If [=microphone information can be exposed=] is <code>false</code>,
+                    truncate <var>microphoneList</var> to its first item.</p>
+                </li>
+                <li>
+                  <p>If [=camera information can be exposed=] is <code>false</code>,
+                    truncate <var>cameraList</var> to its first item.</p>
                 </li>
                 <li>
                   <p>Run the following sub steps for each discovered device in <var>deviceList</var>, <var>device</var>:</p>


karlt (approved):
Thanks!

jan-ivar (approved):
LGTM!

eladalon1983:
This seems reasonable to me. (I ran into #903 while reading this, btw.) LGTM from my side, but please wait for @alvestrand to review as well, as I've not had much mileage on `enumerateDevices()` yet, and I'd prefer one more person from Chrome take a look.

> This ensures we do not get dummy similar audioinput or camera devices if only camera or only microphone is used.

Btw, where's the part of the spec that deals with dummy objects? Looking at [creating a list of device info objects](https://w3c.github.io/mediacapture-main/#creating-a-list-of-device-info-objects), I see that:
* Lists are created for distinct device types.
* The system default device is kept first.
* In the absence of the necessary permission, the lists are truncated to a single element - the first one.

I don't see mention of dummy devices, though. Has this dropped on the edit room floor at some point, or am I misreading things?

karlt:
> Btw, where's the part of the spec that deals with dummy objects?

"videoinput" and/or "videooutput" `MediaDeviceInfo` objects are all dummy objects when their respective "information can be exposed" flag is false, according to [creating a device info object](https://w3c.github.io/mediacapture-main/#creating-a-device-info-object).

eladalon1983:
> > Btw, where's the part of the spec that deals with dummy objects?
> 
> "videoinput" and/or "videooutput" `MediaDeviceInfo` objects are all dummy objects when their respective "information can be exposed" flag is false, according to [creating a device info object](https://w3c.github.io/mediacapture-main/#creating-a-device-info-object).

Ah, the early-exit in steps 3 and 4 (as of the time of this writing). Thanks for clarifying.

eladalon1983:
> > > Btw, where's the part of the spec that deals with dummy objects?
> > 
> > 
> > "videoinput" and/or "videooutput" `MediaDeviceInfo` objects are all dummy objects when their respective "information can be exposed" flag is false, according to [creating a device info object](https://w3c.github.io/mediacapture-main/#creating-a-device-info-object).
> 
> Ah, the early-exit in steps 3 and 4 (as of the time of this writing). Thanks for clarifying.

Actually, do I read correctly that the presence/absence of a default device can still be deduced by an application, even if a permission is lacking, because a dummy object will not be produced on a machine that has no devices at all?

(This is a separate issue, of course. I'll create a new issue for it if you guys think it's real, or if we cannot settle this in one more iteration. In either case, the current PR - LGTM with the previous caveat of wanting another Chromium reviewer this time.)

youennf:
> Actually, do I read correctly that the presence/absence of a default device can still be deduced by an application, even if a permission is lacking, because a dummy object will not be produced on a machine that has no devices at all?

Right, not really a default device though, but whether there is a camera or microphone available.
It would be good if we could find a way to remove those 2 bits of fingerprinting.
Backward compatibility concern was raised when we decided to improve enumerateDevices privacy story a while ago.

eladalon1983:
> Backward compatibility concern was raised when we decided to improve enumerateDevices privacy story a while ago.

Is there a thread I could revive, or shall I start a new one?

youennf:
> Is there a thread I could revive, or shall I start a new one?

There is some history at https://github.com/w3c/mediacapture-main/issues/646.
I think it would be worthwhile to check again what websites are doing in that area and reopen a new issue.

If it is still backward incompatible to return an empty list, we could think of specifying behaviours like:
- enumerateDevices always returns a camera and microphone at first, no matter whether there is actually one.
- After a call to getUserMedia leads to NotFoundError, enumerateDevices exposes whether there is no camera/microphone.

getusermedia.html
```diff
@@ -2921,6 +2925,10 @@ <h2>Methods</h2>
                       [=allowed to use=] the feature identified by <a>"camera"</a>,
                       abort these sub steps and continue with the next device (if any).</p>
                     </li>
+                    <li>If [=camera information can be exposed=] is <code>false</code> and
+                      <var>cameraList</var> is not empty, abort these sub steps and
+                      continue with the next device (if any).</p>
+                    </li>
```
alvestrand, line 2931: I have problems seeing what step is creating the empty camera object for the case where camera information can be exposed is false. Will have to read through the entire algorithm to check that.

jan-ivar: The following sentence creates the camera object:
* Let deviceInfo be the result of [creating a device info object](https://pr-preview.s3.amazonaws.com/youennf/mediacapture-main/pull/900.html#creating-a-device-info-object) to represent device.

Just as for microphone, the modifications are in a for-loop over the device list.

The added condition ISN'T
* Run the following sub steps for each discovered device in deviceList, device
  * If camera information can be exposed is false ... abort

IT'S:
* Run the following sub steps for each discovered device in deviceList, device
  * If camera information can be exposed is false and cameraList is not empty, abort
  * Let deviceInfo be the result of [creating a device info object](https://pr-preview.s3.amazonaws.com/youennf/mediacapture-main/pull/900.html#creating-a-device-info-object) to represent device.
  * [prepend or append] deviceInfo to cameraList.
  * loop

If there is a camera, then we will loop once before we exit.

jan-ivar:
The conditional production of label is unchanged by this PR and happens in [creating a device info object](https://pr-preview.s3.amazonaws.com/youennf/mediacapture-main/pull/900.html#creating-a-device-info-object).

alvestrand (approved):
Yes, this is simple to understand.
Approved.

jan-ivar (approved):

PR #901, opened 2022-09-12
Add `SecureContext` attribute to `InputDeviceInfo`.

dogben:
Per [the WebIDL spec](https://webidl.spec.whatwg.org/#SecureContext), "An interface without the [SecureContext] extended attribute must not inherit from another interface that does specify [SecureContext]." Since `MediaDeviceInfo` has the `SecureContext` attribute, so must `InputDeviceInfo`.

***
<a href="https://pr-preview.s3.amazonaws.com/dogben/mediacapture-main/pull/901.html" title="Last updated on Sep 22, 2022, 2:09 PM UTC (68cdfeb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-main/901/e84ad3e...dogben:68cdfeb.html" title="Last updated on Sep 22, 2022, 2:09 PM UTC (68cdfeb)">Diff</a>

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3323,7 +3323,7 @@ interface MediaDeviceInfo {
       capabilities of the input device it represents.</p>
       <div>
         <pre class="idl"
->[Exposed=Window]
+>[Exposed=Window, SecureContext]
 interface InputDeviceInfo : MediaDeviceInfo {
   MediaTrackCapabilities getCapabilities();
 };</pre>


dogben:
Sending to @youennf as followup to https://github.com/w3c/mediacapture-main/pull/550.

I suspect that the partial interfaces for `MediaDevices` should also have `SecureContext`, but I'm not sure. These include https://www.w3.org/TR/mediacapture-streams/#webidl-823932458 and https://www.w3.org/TR/mediacapture-streams/#webidl-823932458.

youennf (approved):

youennf:
> I suspect that the partial interfaces for `MediaDevices` should also have `SecureContext`, but I'm not sure

I do not think we are doing it right now (meaning that SecureContext and Exposed are defined in the main interface definition).
Maybe we should though, https://webidl.spec.whatwg.org/#dfn-partial-interface tells us:
`The following extended attributes are applicable to partial interfaces: CrossOriginIsolated, Exposed, LegacyOverrideBuiltIns, and SecureContext.`
Do you know what other specs are doing there?

dontcallmedom:
`Exposed` and `SecureContext` only need to appear on partial interfaces to override what is defined in the original interface (e.G. if the partial defines a part of the interface that is only available in a worker):
* the [exposure set](https://webidl.spec.whatwg.org/#dfn-exposure-set) fallbacks on the original interface exposure set (step 4.2)
* `SecureContext` is [forbidden to appear](https://webidl.spec.whatwg.org/#SecureContext) on both an interface and a partial interface (mostly because it's redundant)

Since the main interface of `MediaDevices` is already configured with what we want, no additional extended attributes are needed here.

eladalon1983 (approved):

jan-ivar (approved):

PR #942, opened 2023-03-24
Add note about deviceId and groupId inconsistency with other capabilities.

### LINKED ISSUE #941: Spec-inherent capabilities of one differ from other capabilities of one, an unnecessary variant
I'm raising this as promised in https://github.com/w3c/mediacapture-main/issues/915#issuecomment-1302634199:
> ... consider how constraints with only one valid value would look. This comes up both for non-inherent constraints like [channelCount](https://w3c.github.io/mediacapture-main/#def-constraint-channelCount) on a mono mic (`[1]`), as well as inherent ones like deviceId.
> 
> Having the latter be `"mydeviceIdString1234"` rather than `["mydeviceIdString1234"]` was [an editor's mistake](https://github.com/w3c/mediacapture-main/commit/059a9211d1622094dd8857db9beb477976f971dd) IMHO, but I never complained about it (I'm regretting that right now).
> 
> I'd support changing it (back) to a sequence of one. ... But this would need to be raised in mediacapture-main.

If we look at [MediaTrackCapabilities](https://w3c.github.io/mediacapture-main/#dom-mediatrackcapabilities) today, we see that every member is a range, except for [deviceId](https://w3c.github.io/mediacapture-main/#dom-mediatrackcapabilities-deviceid) and [groupId](https://w3c.github.io/mediacapture-main/#dom-mediatrackcapabilities-groupid):

image

If we dig into why, it's from the commit https://github.com/w3c/mediacapture-main/commit/059a9211d1622094dd8857db9beb477976f971dd, which (outside the scope of its title) removes ranges from `facingMode`, `echoCancellation`, `deviceId`, and `groupId` (`echoCancellation` and `facingMode` were later restored in https://github.com/w3c/mediacapture-main/commit/d5820e22cb76b2b2d3a455c4ad9e723d37378360 & https://github.com/w3c/mediacapture-main/pull/286).

I suppose an argument for this anomaly would be that unlike the others, `deviceId` and `groupId` can _never_ have more than one value, and that therefore it doesn't _need_ to be represented as an array. But this breaks the [Constrainable Pattern](https://www.w3.org/TR/mediacapture-streams/#constrainable-interface).

To the extent that pattern makes any sense at all, we should at least follow it. [applyConstraints with its own exact deviceId succeeds in all browsers](https://jsfiddle.net/jib1/03yg28b9/).

jan-ivar:
Fixes #941.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/mediacapture-main/pull/942.html" title="Last updated on Sep 20, 2023, 9:11 PM UTC (f4e73c5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-main/942/582a0c2...jan-ivar:f4e73c5.html" title="Last updated on Sep 20, 2023, 9:11 PM UTC (f4e73c5)">Diff</a>

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1390,9 +1390,16 @@ interface MediaStreamTrack : EventTarget {
   sequence&lt;boolean&gt; noiseSuppression;
   DoubleRange latency;
   ULongRange channelCount;
-  sequence&lt;DOMString&gt; deviceId;
-  sequence&lt;DOMString&gt; groupId;
+  DOMString deviceId;
+  DOMString groupId;
 };</pre>
+        <div class="note">
+          <p>For historical reasons, {{MediaTrackCapabilities/deviceId}} and
+          {{MediaTrackCapabilities/groupId}} are {{DOMString}} instead of the
+          `sequence&lt;DOMString&gt;` expected by {{Capabilities}} in the
+          <code><a>ConstrainablePattern</a></code>.</p>
+        </div>
+
           <section>
             <h2>Dictionary {{MediaTrackCapabilities}}
             Members</h2>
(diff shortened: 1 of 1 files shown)


eladalon1983:
Could you please help me figure out a hypothetical Web app that'd benefit of this change?

jan-ivar:
It would simplify any app that throws up a user interface to modify constraints. But the overall benefit here is to simplify the logic. The fact that few actual apps rely on it actually helps with that.

alvestrand:
We have used sequence<DOMString> to indicate a bounded set of possible values, I'm not sure if this syntax was chosen to indicate an unknowable set of values.

From the constrainable pattern text:

"Capabilities is a dictionary containing one or more key-value pairs, where each key MUST be a constrainable property, and each value MUST be a subset of the set of values allowed for that property. "

Query: should there be a recommendation that the value given is the default device for that type, and that more values should only be added if they would be returned in navigator.mediaDevices? Or is that making life more complicated?

jan-ivar:
The history here is that settings, capabilities, and constraints originally all used the same single WebIDL dictionary, [MediaTrackConstraintSet](https://w3c.github.io/mediacapture-main/#dom-mediatrackconstraintset). This made sense from a spec economy pov, since you can express both settings and capabilities as constraints. I.e. settings and capabilities are subsets of constraints (and also valid arguments to constraints, even today).

But then we realized this was a parsing nightmare for JS, since no guarantees were given in prose about how a value would be returned, e.g. in getCapabilities().

JS would _in theory_ have to be ready for a user agent to return e.g. `{channelCount: 1` or `{channelCount: [1]}` or `{channelCount: {exact:1}` or `{channelCount: {exact: [1]}`. And that this was terrible for web compat.

It was when making this transition to tripple dictionaries that this editorial mistake was introduced IMHO.

jan-ivar:
Folks prefer an oops note instead.

PR #996, opened 2024-04-19
Add DeviceChangeEvent interface.

jan-ivar:
Fixes #972.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/mediacapture-main/pull/996.html" title="Last updated on Apr 25, 2024, 2:34 PM UTC (7bec415)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-main/996/6a4eccb...jan-ivar:7bec415.html" title="Last updated on Apr 25, 2024, 2:34 PM UTC (7bec415)">Diff</a>

PR #1000, opened 2024-04-27
Fix DeviceChangeEventInit WebIDL definition

chrisguttandin:
This PR is intended to fix a tiny typo introduced here: https://github.com/w3c/mediacapture-main/pull/996/files#diff-1217ca1c44ff30a33dd50c49d03b5cadc9633c789df8ff9370ed4a42859e1211R3468-R3470.

I guess it would also resolve #999.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3471,7 +3471,7 @@ interface DeviceChangeEvent : Event {
       <div>
         <pre class="idl"
 >dictionary DeviceChangeEventInit : EventInit {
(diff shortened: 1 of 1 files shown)


alvestrand (approved):

alvestrand:
@jan-ivar I think this is a fix - the "members" section says that it is a sequence of deviceInfo.
@dontcallmedom I think this doesn't need to be submitted by a group member (however joining the group would still be OK!)

guidou (approved):

PR #1007, opened 2024-06-14
Add userInsertedDevices attribute to DeviceChangeEvent

### LINKED ISSUE #966: Should devicechange fire when the device info changes?
I noticed a browser inconsistency in the way the `'devicechange'` event gets fired. But I think the spec isn't very clear which browser is behaving correctly.

These days all browsers reduct the information available by querying `enumerateDevices()` as long as the page has for example no permission to access the microphone. However this changes after a successful call to `getUserMedia()`. This behavior seems to be similar across all browser.

But only Safari fires a `'devicechange'` event when it exposes the `label` of a device as a result of a permission change. This inconsistency can be easily worked around by calling `enumerateDevices()` again after each call to `getUserMedia()` but it doesn't work if the user changes the page permission in the browser settings.

The spec currently says ...

> When new media input and/or output devices are made available to the [User Agent](https://w3c.github.io/mediacapture-main/#dfn-user-agent), or any available input and/or output device becomes unavailable, or the system default for input and/or output devices of a [MediaDeviceKind](https://w3c.github.io/mediacapture-main/#dom-mediadevicekind) changed, ...

At least on my MacBook Chrome is the only browser which actually adds a device after the user has granted permissions to see the full list of devices. It shows the default device twice. However it doesn't fire a `'devicechange'` event.

Firefox only changes the `label` which I would argue means the device is still the same and therefore `'devicechange'` shouldn't fire at least in the way it is defined today.

Safari changes all properties including `deviceId` and `groupId` which could be interpreted as a completely new device. It consequently fires a `'devicechange'` event.

As a developer building applications using this API I would love all browsers to fire a `'devicechange'` event whenever anything changes in the device list including `label` changes.

jan-ivar:
Fixes #966. For discussion.

***
<a href="https://pr-preview.s3.amazonaws.com/jan-ivar/mediacapture-main/pull/1007.html" title="Last updated on Jun 20, 2024, 1:33 PM UTC (898177b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mediacapture-main/1007/cec8e2a...jan-ivar:898177b.html" title="Last updated on Jun 20, 2024, 1:33 PM UTC (898177b)">Diff</a>

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3467,6 +3467,10 @@ interface DeviceChangeEvent : Event {
                 to disambiguate this signal from differences in {{devices}}
                 that might happen from changes in device information exposure.</p>
               </div>
+              <div class="note">
+              If the attribute is missing, it means that this UA has not been upgraded to implement
+              this version of the specification.
+              </div>
             </dd>
           </dl>
         </section>
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -3412,6 +3412,7 @@ interface InputDeviceInfo : MediaDeviceInfo {
 interface DeviceChangeEvent : Event {
   constructor(DOMString type, optional DeviceChangeEventInit eventInitDict = {});
   [SameObject] readonly attribute FrozenArray&lt;MediaDeviceInfo&gt; devices;
+  [SameObject] readonly attribute FrozenArray&lt;MediaDeviceInfo&gt; userInsertedDevices;
 };</pre>
         <section>
           <h2>Constructors</h2>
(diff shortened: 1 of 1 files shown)


getusermedia.html
```diff
@@ -3439,6 +3440,34 @@ <h2>Attributes</h2>
...
+              <p>The {{MediaDeviceInfo}} objects, if any, MUST also exist
+              in {{devices}}.</p>
+              <div class="note">
+                <p>A user inserting a device during (or immediately ahead of)
+                a call can be a strong signal that they wish to use the
+                device immediately.</p>
+                <p>Applications are encouraged to rely on this attribute
+                to disambiguate this signal from differences in {{devices}}
+                that might happen from changes in device information exposure.</p>
+              </div>
```
alvestrand, line 3469:
```suggestion
              </div>
              <div class="note">
              If the attribute is missing, it means that this UA has not been upgraded to implement
              this version of the specification.
              </div>
```

PR #1042, opened 2025-05-08
Pair resizeMode: "none" settings-dictionaries w/identical crop-and-scale ones.

### LINKED ISSUE #1041: Clarify resizeMode: none ⊆ crop-and-scale
While [implementing resizeMode](https://phabricator.services.mozilla.com/D241316#8589575), we ran into some confusion:
> a camera with one capability of 640x480@30 and the constraint `{resizeMode: {exact: 'crop-and-scale'}}`, what settings do we pick?

A strict interpretation of the [fitness distance](https://w3c.github.io/mediacapture-main/#dfn-fitness-distance) algorithm might suggest this should _exclude_ native modes, and return the lower-distance 639x480@30.

But that's an undesirable result, unless one's goal is to constrain around non-native modes, which was never our intent.

The more intuitive expectation (that also matches implementations) is that `"none"` acts as a filter. I.e. native modes ⊆ `crop-and-scale` modes.

Fix:
- Pair every [settings dictionary](https://w3c.github.io/mediacapture-main/#dfn-settings-dictionary) containing `resizeMode: "none"`, with an identical settings dictionary containing `resizeMode: crop-and-scale`.

jan-ivar:
Fixes #1041.

### DIFF
--- a/getusermedia.html
+++ b/getusermedia.html
@@ -1183,6 +1183,19 @@ interface MediaStreamTrack : EventTarget {
                             constrainable properties that at times <a>must not
                             be exposed</a>.</p>
                           </li>
+                          <li>
+                            <p>For every [=settings dictionary=] with
+                            <a href="#def-constraint-resizeMode">resizeMode</a>
+                            set to
+                            <a href="#idl-def-VideoResizeModeEnum.none">"none"</a>,
+                            the [=User Agent=] MUST include another otherwise
+                            identical [=settings dictionary=] with
+                            <a href="#def-constraint-resizeMode">resizeMode</a>
+                            set to
+                            <a href="#idl-def-VideoResizeModeEnum.cropandscale">"crop-and-scale"</a>.
+                            Constraining around non-native modes is not possible.
+                            </p>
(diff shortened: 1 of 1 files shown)


youennf (approved):

getusermedia.html
```diff
@@ -1183,6 +1183,19 @@ <h2>Methods</h2>
...
+                            <p>For every [=settings dictionary=] with
+                            <a href="#def-constraint-resizeMode">resizeMode</a>
+                            set to
+                            <a href="#idl-def-VideoResizeModeEnum.none">"none"</a>,
+                            the [=User Agent=] MUST include another otherwise
+                            identical [=settings dictionary=] with
+                            <a href="#def-constraint-resizeMode">resizeMode</a>
+                            set to
+                            <a href="#idl-def-VideoResizeModeEnum.cropandscale">"crop-and-scale"</a>.
+                            Constraining around non-native modes is not possible.
```
jan-ivar, line 1196: Add a note saying this effectively means crop-and-scale is a superset of none.

